PR #4, opened 2021-05-08
Add TypeScript, Prettier and ESLint

MrTimscampi:


src/components/FreeSoftware.tsx
```diff
@@ -10,11 +10,21 @@ export default function FreeSoftware() {
           <div className="col col--6">
             <Svg className={styles.logo} />
             <h1>Your media, your server, your way</h1>
-            <div className="margin-bottom--sm"><b>Jellyfin is fully self-hosted and fully open source.</b></div>
-            <div>We don't do tracking or paid plans. There is no forced connection to a remote server. You are in control from start to finish.</div>
+            <div className="margin-bottom--sm">
+              <b>Jellyfin is fully self-hosted and fully open source.</b>
+            </div>
+            <div>
+              We don&apos;t do tracking or paid plans. There is no forced
```
dkanada, line 17: Is this somehow more web compliant than a regular apostrophe?

src/css.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "*.module.css" {
+  const classes: { [key: string]: string };
+  export default classes;
+}
```
dkanada, line 4: pick a lane :p

src/components/FreeSoftware.tsx
MrTimscampi: The React rules were complaining about using a normal single quote here, and recommended this as a replacement.

src/pages/index.tsx
```diff
@@ -19,10 +17,27 @@ function HomepageHeader() {
           <div className="hero-content">
             <div className="hero-content-inner">
               <h1 className="hero-title">The Free Software Media System</h1>
-              <p className="hero-text">Jellyfin is the volunteer-built media solution that puts <em>you</em> in control of your media. Stream to any device from your own server, with no strings attached. Your media, your server, your way.</p>
-              <a href="https://demo.jellyfin.org/stable" className="button button--secondary button--outline">See it in Action</a>
-              <a href="/downloads" className="button button button--primary margin-horiz--md">Download Now</a>
-              <a href="" className="button button--secondary button--outline scroll">Learn More</a>
+              <p className="hero-text">
+                Jellyfin is the volunteer-built media solution that puts{' '}
```
dkanada, line 21: What's with the weird space in brackets?

src/css.d.ts
MrTimscampi: It's not quite working anyway :yum:

That's how all the examples online do it, but it's throwing an error.

src/pages/index.tsx
MrTimscampi: > This is used to put an explicit space in a text block, since leading and trailing spaces are ignored at compile/transformation time when there is another tag.

https://stackoverflow.com/a/38547635/7617396

ferferga (approved):

PR #5, opened 2021-05-08
Decrease hero height and remove background image

nielsvanvelzen:
image


package-lock.json
```diff
@@ -1,8 +1,12971 @@
 {
   "name": "jellyfin-org",
   "version": "0.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
MrTimscampi, line 4: We probably want to keep Node 14/NPM 6 until Node 16 enters LTS (It's not there yet, I think it's for next month).

MrTimscampi (commented):
If we're going to replace the image later, shouldn't we simply keep the current one until then and only reduce the size?

nielsvanvelzen:
> If we're going to replace the image later, shouldn't we simply keep the current one until then and only reduce the size?

The image looked really ugly with this size (pretty much only the television is visible) so I just removed it.

package-lock.json
ferferga: On October 26

https://nodejs.org/en/about/releases/

MrTimscampi (approved):

PR #6, opened 2021-05-08
Port documentation from DocFX

MrTimscampi:
This starts porting the documentation.

image

* It reorganizes pages into more logical categories
* It removes and moves some pages that will/should be moved out of the user documentation (Codecs, FAQ, Getting Help, Community Standard)
* It starts converting DocFX syntax to Docusaurus, but doesn't do all of it (Help is wanted here)

The pages which were moved have been added to the footer:
image


docs/migrate-from-emby.md
```diff
@@ -0,0 +1,72 @@
+---
+id: admin-migrate-from-emby
+title: Migrating from Emby
```
Artiume, line 3: This name is originally from when migrating from emby was possible and common. perhaps we should rename it to just Migrations. We could also put it under the Server section (and remove off the sidebar) since these methods aren't technically supported and aren't guaranteed to work.

MrTimscampi: Maybe "Migrating to Jellyfin", with a warning that it's possibly deprecated?

Artiume: i was thinking that there's also cross system migrations like baremetal > docker migrations that we should provide documentation on. "Migrations" should be vague enough

MrTimscampi: Fair, that makes sense.

image

I moved it to "Tips & tricks" (For which I'd like to find a different name... If you have ideas x)

docs/server/index.md
```diff
@@ -0,0 +1,43 @@
+---
+id: server-live-tv-index
```
Artiume, line 2: the name index.md doesnt seem to be consistent with the file itself.

docs/setup-live-tv.md
```diff
@@ -0,0 +1,85 @@
+---
+id: setup-live-tv
+title: Setting up live TV
```
Artiume, line 3: this file seems to repeat some of the content from docs/server/index.md, maybe we should combine them into one or figure out which one should be first

MrTimscampi: Yeah, there's a bunch of re-writing that needs to be done, this was mainly intended to move files into Docusaurus and give them a decent structure, I didn't touch the content at all :smiley:

Artiume:
On the sidebar, there's Server configuration and Server Administration, perhaps there's better names we could use to differentiate

docs/setup-live-tv.md
Artiume: yeah, and once the site is live, i'll be able to give better feedback on the structuring

Artiume:
For the 'Using with a Reverse Proxy' perhaps we could call it 'External Access' or maybe 'Internet Access'

docs/configuration/networking.md
```diff
@@ -0,0 +1,133 @@
+---
+id: networking
```
Artiume, line 2: this was originally the index page to the Networking Section. Ideally, this file would be index to the 'Internet Access' section, but we'd have to rewrite the intro to it to talk about it

MrTimscampi: There is no index pages here, per se, here, afaik.

The best we can do is make an "Introduction" page and set it as position one, but that's about it.

MrTimscampi: There's also no Networking section anymore.

MrTimscampi:
> On the sidebar, there's Server configuration and Server Administration, perhaps there's better names we could use to differentiate

One is about using the server configuration files, hardware acceleration, networking and how to troubleshoot.

The other is basically describing all the admin pages from the web client...

> For the 'Using with a Reverse Proxy' perhaps we could call it 'External Access' or maybe 'Internet Access'

I'd like to keep it somewhat direct, so it's easier to find. Plus you don't only use reverse proxies for access over the Internet and "External access" is pretty vague.

The idea here is to group related sections together, in a way that would make sense to a user coming in for specific info.

Artiume:
I figured it out, the 'Using a reverse proxy' was my Networking section.

> One is about using the server configuration files, hardware acceleration, networking and how to troubleshoot.
> The other is basically describing all the admin pages from the web client...

Perhaps we should rename the server setup one to Server Setup. and move the Getting Started,  Installing the Server and Setting up Live TV sections into it since that's all one topic.

MrTimscampi:
> move the Getting Started, Installing the Server and Setting up Live TV sections into it since that's all one topic.

These are all on the top level for quick access. They're stuff that most people would be looking for, so having them directly visible is an advantage.

> Perhaps we should rename the server setup one to Server Setup

It's not really about server setup, though? That would be "Installing the server". This is more about configuration files and specific server administration...

It's honestly fairly difficult to make our current documentation make sense without rewriting/merging/splitting a bunch of it :/

Artiume:
yeah, a lot of my comments are just carryover from the old docs format, this is my attempt to get you to fix some of it for me x).

As for the install vs setup, that's typically done at the same time, no? I don't really config it except during installs/upgrades. The Getting Started section could be absorbed into the Install footer, it doesn't say much except to Install and visit the web port

MrTimscampi:
> this is my attempt to get you to fix some of it for me

I'll probably do a large editing pass on the docs after this one is merged :smile: Mostly to further improve the structure, fix links, fix a bunch of formatting and so on.

This one is meant to leave files mostly untouched, aside from putting in a basic structure, updating the front matter, etc :)

> that's typically done at the same time, no?

Not necessarily. You can come back later to add hardware acceleration, for example, or want to change some option. These are usually split in documentation.

> The Getting Started section could be absorbed into the Install footer, it doesn't say much except to Install and visit the web port

That's the point :yum: Think of it as like a "what's needed to get setup, as simply as possible". It's goal is to introduce the general steps and redirect you into the more specific sections. Though it needs a bunch of editing, currently.

PR #20, opened 2021-07-23
Redesign the landing page

MrTimscampi:
I went over the home page and the general theme in order to start to move towards a less monochromatic and brighter theme.

This introduces a lot more color into our landing page, to break the flow a bit, and plays a lot more with typography.

image

**Updated to current design as of 2021/08/02**


MrTimscampi:
Of note, this also proposes a change of our primary accent color to `#9d37c2` (From `#00A4DC`).

This color has nicer pairings and is brighter/more bold, and also fits well with the WCAG checks (The current blue doesn't pass any WCAG contrast checks with white text).

joshuaboniface:
Things I like:

The new accent colour looks brighter, and definitely pops more.

I like the way that first bar works.

In general I love the content.

Things I dislike:

I'm really not a fan of the overall page colour layout Jellyfin (purple-blue gradient)/Dark Grey/White/Red/White/Yellow/Dark Grey colour progression. It seems really jarring to me, specifically having the red and yellow sections in there.

I think the Jellyfin colours alone are enough, without the Red and Yellow parts. I'd be curious to see mockups of:

Jellyfin/Dark Grey/White/Dark Grey/White/Jellyfin/Dark Grey

and

Jellyfin/Dark Grey/Jellyfin/White/Jellyfin/White/Dark Grey

Or perhaps having Jeff Blue/Jeff Purple/Jeff Blue/Jeff Purple alternate in the middle.

Maybe too the little apps list could be in the grey colour as well?

joshuaboniface:
>a less monochromatic

I think this can be achieved while sticking with our 4 primary colours (blue, purple, grey, and white) without looking really jarring TBH. I'm a fan of monochromatic, or at least fixed-palate, designs.

nvllsvm:
Not a fan of the seemingly random colors, red and yellow. Keeping with the existing colors creates a sense of branding.

Also - the blue background behind the logo is jarring and makes the logo difficult to see.

Screenshot_20210724-011410_Firefox_1

prayag17:
How did you create the device image?

MrTimscampi:
@joshuaboniface @nvllsvm Taking your feedback into account, here is a second pass on the page:

image

I removed the yellow in order to have less colors, but I'd like to keep the reddish gradient in the middle, as I think it makes for a nice break. It's not random, like was supposed above, but is actually a split complementary of the primary color (the purple used on the buttons, which is itself a modified purple from our gradient, selected to pass WCAG tests with white text).

In order to "offset" the loss of a color, I introduced a rotating image header. I contains 5 images of public domain movies I took myself and processed a bit using Gimp (instructions will be in the CONTRIBUTING.md once I push this).

The use of images also fixes Andrew's point about the logo's gradient clashing with the background one.

I'd like to avoid using big swathes of light grey, and a darker one between two white blocks would clash a bit too much visually.

As for the suggestion of using a shade of grey for the devices/clients block, I wanted to keep it light in order to break the visual flow of the page a bit more, and keep a generally bright look.

The idea here is to "sell" Jellyfin to first-time visitors, so I'd like to have something that pops out and keeps their attention. I feel like the current home page, while nice, doesn't achieve that due to being too monotone and hard to parse, hence the somewhat jarring changes here.

One issue with monotone designs is that it relies heavily on typography to make things nicer, since you only have one real hue to play with. So, essentially, we'd have a very white design, with some small accents of blueish-purple, which looks less than a product page for software and more like a luxury or clothing brand.

Introducing more color also gives us more options for designs of other pages down the line. While I don't think we'll use a bright gradient like the contributors' one anywhere else, that section does introduce the "idea" of pinkish-red/orange into the theme, which gives us a nice contrast color for the announcement banner, or accent colors for the documentation or client page, etc.

I hope my explanations give you a bit more insight into my intent with this redesign. I'm open to more feedback, should you have any (Though I think this fixes all your points while keeping my original intent).

Also of note, the devices image still needs to be redone with newer/better device templates and better screenshots. @prayag17 It's made using a crappy online tool. I'm currently looking at good alternatives to do these in a cleaner way.

prayag17:
Newer mind about the image of the device I made it myself in Figma. 
https://www.figma.com/files/project/32994119/Team-project?fuid=988853667586204645

joshuaboniface:
OK, now I like that!

prayag17:
I am also working on the redesign but it would take me some time, this is what I have changed so far
JF Site redesign WIP

MrTimscampi:
@prayag17 We appreciate your enthusiasm, however for central project areas such as the homepage, we usually work on this among the team. There is extensive internal communication through team channels while I work on this, in order to get some amount of pre-approval on most aspects of the new designs, which is then formalized.

While your feedback is welcome and gets evaluated, please understand that the end result will likely look closer to something the team has come to agree upon.

We understand that it's not really 100% in the spirit of the rest of the project, but since this represents the project as a whole and serves as branding and marketing material, we feel it's important to have a slightly tighter grip on the process, at least in this early part.

MrTimscampi:
This PR will be continue in #27 , as we are moving from Docusaurus to Gatsby, due to limitations in regards to some of the things we'd like to do.

PR #29, opened 2021-08-05
Pull GitHub stats from GraphQL API

MrTimscampi:


nielsvanvelzen:
Wouldn't it be better to get the GitHub GraphQL schema from their own package?
https://www.npmjs.com/package/@octokit/graphql-schema

PR #46, opened 2022-07-29
Move hero to component

thornbill:
* Moves the hero element to a component for reuse on other pages
* Removes some unused/unneeded code


src/components/HomepageFeatures.tsx
```diff
@@ -69,7 +68,7 @@ function Feature({
   key: number;
 }) {
   return (
-    <div className={clsx('col col--3')}>
```
thornbill, line 72: `clsx` is unnecessary here since there are no conditional classes

src/pages/index.module.css
```diff
@@ -1,25 +0,0 @@
-/* stylelint-disable docusaurus/copyright-header */
```
thornbill, line 1: This file was unused.

nielsvanvelzen (approved):
I think adding SCSS at some point would be a nice improvement for managing all styling.

thornbill:
Agreed. Docusaurus has a sass plugin that looks pretty easy to add.

PR #49, opened 2022-08-01
Add downloads page

thornbill:
Screenshot 2022-07-31 at 20-59-04 Downloads Jellyfin


anthonylavado:
I like the little "Official"/"Community" tags, those are a nice touch.

I wonder if later on we could even restructure bit further with collapsible sections by OS, but I digress.

anthonylavado (approved):

PR #73, opened 2022-08-04
Add clients page

thornbill:
Data was sourced from https://jellyfin.org/clients/, https://jellyfin.org/docs/general/clients/index.html, and individual project repositories. 

Any clients that did not meet the WIP requirements described in https://github.com/jellyfin/jellyfin-meta/discussions/27 were excluded. There had been some discussion about whether in-development clients should be included or not, so I have excluded them for now. I think they could probably be added to the "All Clients" listing with a badge indicating their status though. The most notable example of this would probably be Swiftfin.

Some clients were difficult to categorize into device types and platforms... so additional options might be needed there.

There is some groundwork for including additional filtering options, but I think I'm going to leave that for a separate PR considering the size of this one. :eyes: 

**Recommended Clients**
Screenshot 2022-08-04 at 00-54-55 Clients Jellyfin

**All Clients**
Screenshot 2022-08-04 at 00-55-05 Clients Jellyfin

### DIFF
--- a/src/pages/clients.tsx
+++ b/src/pages/clients.tsx
@@ -6,10 +6,10 @@ import ClientDetails from '../components/ClientDetails';
 import { Clients, DeviceType, Platform } from '../data/clients';
 
 type ClientFilter = {
-  recommended: boolean,
-  deviceType?: DeviceType,
-  platform?: Platform
-}
+  recommended: boolean;
+  deviceType?: DeviceType;
+  platform?: Platform;
+};
 
 export default function Plugins() {
   const [filter, setFilter] = useState<ClientFilter>({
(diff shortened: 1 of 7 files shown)


src/pages/clients.tsx
```diff
@@ -1,11 +1,81 @@
...
-    <Layout>
-      <h1>Clients</h1>
-      <p>Placeholder page</p>
+    <Layout title='Clients'>
+      <h1 className='text--center'>Clients</h1>
+
+      <main className='margin-vert--lg'>
+        <section className='container'>
+          <ul className='pills'>
+            <li
```
nielsvanvelzen, line 26: Wouldn't it be possible to use native <button> elements here?

thornbill: Native elements like `<button>`? or create a `<Pill>` component in react?

I had attempted to use a button for these instead of using the role + key event handler but the css framework seems to not support them unfortunately. :confused:

nielsvanvelzen: Native element as in the button element. that would help with the enter key

But if the CSS doesn't work with it I guess this will do for now

nielsvanvelzen (approved):

PR #80, opened 2022-08-05
Downgrade to React 17

thornbill:
Docusaurus has not [migrated to React 18](https://github.com/facebook/docusaurus/issues/7264) yet. While it doesn't really seem to cause any issues, it does log an error, and it is explicitly not supported for some plugins (specifically the redirect plugin I was looking at).

The lockfile was bumped to the v2 format which should be fine here since we are using node 16.

~~Also I have no idea why prettier is randomly removing trailing newlines in json files, but it's kind of annoying...~~ It seems VS Code formatting on save is enabled, but is not using prettier and it's formatting removes trailing newlines and makes other changes that do not align with the prettier config.


.github/renovate.json
```diff
@@ -1,7 +1,13 @@
...
+  "extends": ["github>jellyfin/.github//renovate-presets/nodejs", ":semanticCommitsDisabled"],
+  "packageRules": [
+    {
+      "description": "Disable major updates for React",
+      "matchPackagePrefixes": ["react"],
+      "matchPackageNames": ["@types/react", "@types/react-helmet", "@types/react-router-dom"],
+      "matchUpdateTypes": "major",
+      "enabled": false
+    }
+  ]
```
nielsvanvelzen, line 12: Do we **never** want to get React updates from Renovate or just skip React 18? Closing a Renovate update PR just skips a single specific version.

thornbill: I believe any major updates for React will require support in docusaurus first.

nielsvanvelzen (approved):

PR #85, opened 2022-08-16
Redesign the downloads page to match the clients page design

thornbill:
Screenshot 2022-08-16 at 13-06-14 Downloads Jellyfin

### DIFF
--- a/src/components/ClientDetails.tsx
+++ b/src/components/ClientDetails.tsx
@@ -1,9 +1,8 @@
 import React from 'react';
 
+import DetailsCard from './common/DetailsCard';
 import PlatformIcon from './PlatformIcon';
 import { Client, ClientType, LicenseType } from '../data/clients';
-
-import './ClientDetails.css';
 
 const ClientTypeBadge = ({ clientType }: { clientType: ClientType }) => {
   if (clientType === ClientType.Official) {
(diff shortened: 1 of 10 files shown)


nielsvanvelzen:
Looks great overall, got some suggestions for the page (although they could be implemented in future PR's):

- Show current platform by default (so a user using Windows sees the Windows tab when opening the page)
- Add toggle for stable/unstable and only show instructions/downloads for that specific version (and ofc use stable by default)
- Add the current version to the page, maybe even to the header so it's immediately clear what the latest (stable) version is when visiting the site
- Make "docker" it's own tab, I believe it's popular enough to do so, I've seen other software do this too

thornbill:
* Show current platform by default (so a user using Windows sees the Windows tab when opening the page)
  * I was planning to add this in a separate PR in the future
* Add toggle for stable/unstable and only show instructions/downloads for that specific version (and ofc use stable by default)
  * It's not obvious in a single screenshot, but the instructions are hidden by default. When a user clicks a stable/unstable button it will either display the installation instructions or open the relevant link to the repo site.
* Add the current version to the page, maybe even to the header so it's immediately clear what the latest (stable) version is when visiting the site
  * This is a great idea... I'll see what I can come up with for this.
* Make "docker" it's own tab, I believe it's popular enough to do so, I've seen other software do this too
  * Good call... Currently the "Other" tab only includes docker and the dotnet portable build. I think I will rename it to "Docker" and include the portable build on the other tabs.

nielsvanvelzen:
> - Add toggle for stable/unstable and only show instructions/downloads for that specific version (and ofc use stable by default)
>   - It's not obvious in a single screenshot, but the instructions are hidden by default. When a user clicks a stable/unstable button it will either display the installation instructions or open the relevant link to the repo site.

I think adding a single toggle (switch control on the right of the tabs maybe?) might still be better. 

You don't need to repeat all the unstable buttons everywhere, and make it will be slightly harder to  accidentally install unstable when you want stable (we have quite some users that aren't that technical and just press the very first button they see). 

Take for example the .NET download site: https://dotnet.microsoft.com/en-us/download. You need to open the "all versions" list before you can attempt to download something different from a stable release.

joshuaboniface:
>Show current platform by default (so a user using Windows sees the Windows tab when opening the page)

For the server page, I do not like this. It gives the false idea that an e.g. Windows client user "should" use the Windows server, which is not really a valid assumption. It also assumes the server would be installed on the machine doing the browsing, which is a little weird.

>Add the current version to the page, maybe even to the header so it's immediately clear what the latest (stable) version is when visiting the site 

If we can get this automatically from the releases page that's ideal. If it's manual updates hard no.

>Make "docker" it's own tab, I believe it's popular enough to do so, I've seen other software do this too 

Yes Docker is by far one of the most popular install types. It should be its own category high up (maybe just below Debuntu repo).

thornbill:
I agree that listing the current release should be automated... although I'm not sure what the best way to handle that is currently. We can definitely fetch the release info via the GH API, but it would probably be best to do that as part of a build/ci process.

Updated screenshot that incorporates feedback:
Screenshot 2022-08-17 at 03-00-50 Downloads Jellyfin

src/css/custom.css
```diff
@@ -5,20 +5,20 @@
  * work well for content-centric websites.
  */
 
- html {
+html {
```
thornbill, line 8: Prettier made some formatting changes to this file... the only relevant change is the the font color for badges.

thornbill:
Updated screenshot with pills for the stable selector instead of radio buttons:

Screenshot 2022-08-17 at 09-44-23 Downloads Jellyfin

Issue #89, opened 2022-08-17
Update docs content


thornbill:
The docs site content needs ported over again since it is ~1 year out of date. There are also some pages that were in docs that were moved to the site root that should probably be moved back for now. 

The docs here were updated as of https://github.com/jellyfin/jellyfin-docs/commit/612a09258fe1793a091ce5d5cc365c0651a0c4f9.

The following changes need ported here to be current:

- [x] https://github.com/jellyfin/jellyfin-docs/pull/698
  - Blocked by https://github.com/jellyfin/jellyfin.org/pull/150
  - https://github.com/jellyfin/jellyfin.org/pull/217
- [x] https://github.com/jellyfin/jellyfin-docs/pull/705
  - https://github.com/jellyfin/jellyfin.org/pull/182
- [x] https://github.com/jellyfin/jellyfin-docs/pull/700
  - https://github.com/jellyfin/jellyfin.org/pull/182
- [x] https://github.com/jellyfin/jellyfin-docs/pull/708
  - https://github.com/jellyfin/jellyfin.org/pull/166
- [x] https://github.com/jellyfin/jellyfin-docs/pull/710
  - https://github.com/jellyfin/jellyfin.org/pull/217
- [x] https://github.com/jellyfin/jellyfin-docs/pull/712
  - https://github.com/jellyfin/jellyfin.org/pull/217
- [x] https://github.com/jellyfin/jellyfin-docs/pull/714
  - https://github.com/jellyfin/jellyfin.org/pull/217

joshuaboniface:
Did the About page in #214

thornbill:
This is being tracked under #218 now

PR #100, opened 2022-08-18
Update docs content

thornbill:
This pulls in the docs site contents as of https://github.com/jellyfin/jellyfin-docs/commit/612a09258fe1793a091ce5d5cc365c0651a0c4f9

I think it would be good to separate out administration, community, and developer documentation at some point in the future, but it's probably best to port it over 1:1 at first.

Screenshot 2022-08-23 at 11-27-55 Introduction Jellyfin

Screenshot 2022-08-23 at 11-28-35 Installing Jellyfin Jellyfin

Related to #89, but additional updates will be needed as PRs are merged to jellyfin-docs prior to this site going live...


joshuaboniface (approved):
Massive change but given that I've been following your work I trust that this is all good. :+1:

nielsvanvelzen (approved):

PR #106, opened 2022-08-23
Add URL test action

thornbill:
`jellyfin-web` urls were found by grepping for "jellyfin.org"

`jellyfin-blog` and `jellyfin-docs` urls were found using their respective sitemaps and `xq '[.urlset.url[].loc]'`. All `plugin-api` urls were manually removed from the list for jellyfin-docs.


thornbill:
These pages _should_ work, but for some reason they return a 404 when using `docusaurus serve` even though the correct content is rendered. It seems to affect all urls containing periods... :thinking: 
* /posts/roku-v1.5.0/
* /posts/android-v2.4.0/
* /posts/android-v2.3.0/
* /posts/android-v2.1.0/

These pages were never actually used on the current site and are more or less broken... so perhaps they should just be removed from the check.
* /categories/
* /tags/

nielsvanvelzen (approved):

nielsvanvelzen (approved):

PR #122, opened 2022-08-29
Add OS specific download pages

thornbill:
Like #121 but for downloads

### DIFF
--- a/src/pages/downloads/macos.tsx
+++ b/src/pages/downloads/macos.tsx
@@ -1,0 +1,8 @@
+import React from 'react';
+
+import DownloadsPage from '.';
+import { OsType } from '../../data/downloads';
+
+export default function LinuxDownloads() {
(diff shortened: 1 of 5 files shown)


src/pages/downloads/macos.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+
+import DownloadsPage from '.';
+import { OsType } from '../../data/downloads';
+
+export default function LinuxDownloads() {
```
crobibero, line 6: is everything a `LinuxDownload`?

thornbill: Linux is the default value https://github.com/jellyfin/jellyfin.org/pull/122/files#diff-91bb82f8d57d0f0b7c232e8df20f8997c93c56bb50c46197297937c06c325e6cR11

PR #134, opened 2022-09-01
Home screen cleanup

thornbill:
Fixes a few minor miscellaneous issues:

* Prevent orphaned word in clients heading
* Fix features section class
* Fix type error when importing png files

### DIFF
--- a/src/components/home/ClientSection.tsx
+++ b/src/components/home/ClientSection.tsx
@@ -16,7 +16,7 @@ export default function ClientSection() {
       <div className='container text--center'>
         <div className='row row--center'>
           <div className='col col--8'>
-            <h2>Your media, wherever you are</h2>
+            <h2>Your media, wherever you&nbsp;are</h2>
             <p>
               With a large array of official and third-party clients, Jellyfin is available on most popular platforms.
               Your media is ready to follow you, wherever you go.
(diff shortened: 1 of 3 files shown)


src/components/home/ClientSection.tsx
```diff
@@ -16,7 +16,7 @@ export default function ClientSection() {
       <div className='container text--center'>
         <div className='row row--center'>
           <div className='col col--8'>
-            <h2>Your media, wherever you are</h2>
+            <h2>Your media, wherever you&nbsp;are</h2>
```
nielsvanvelzen, line 19: Probably better to use CSS for this;

https://developer.mozilla.org/en-US/docs/Web/CSS/white-space

thornbill: I'm not trying to stop all wrapping here, just avoid "are" being alone on a second line since that looks weird. Is there a way to handle that with CSS other than wrapping the two words in a `<span>` and using `no-wrap`?

I have been using non-breaking spaces for this for a very long time, so I would not be surprised at all if there is a better way to handle it now. :sweat_smile:

nielsvanvelzen: Ah got it. I'm not sure if that's easily fixable with CSS, using `&nbsp;` might be the easiest solution then :p

nielsvanvelzen (approved):

PR #142, opened 2022-09-03
Modernize APT configuration for Debian & Ubuntu

iammattcoleman:
Recently, Debian-based distributions have switched to using the deb822
configuration format for APT sources by default.

This feature was first released in APT 1.1.

So, deb822-formatted sources have been supported since...
* Debian 9
* Ubuntu 16.04

Additionally, it is no longer recommended to globally add repository GPG
keys to APT's configuration. Instead, keys should be stored in
/etc/apt/keyrings and referenced only by the sources signed by them.

See: https://wiki.debian.org/DebianRepository/UseThirdParty
See: https://salsa.debian.org/apt-team/apt/-/commit/81460e32961bb0b9922bf8a1a27d87705d8c3e51

### DIFF
--- a/src/data/downloads.tsx
+++ b/src/data/downloads.tsx
@@ -57,20 +57,28 @@ export const Downloads: Array<Download> = [
             <pre>
               <code>
                 {`sudo apt install curl gnupg
-curl -fsSL https://repo.jellyfin.org/ubuntu/jellyfin_team.gpg.key | sudo gpg --dearmor -o /etc/apt/trusted.gpg.d/jellyfin.gpg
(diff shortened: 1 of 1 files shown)


iammattcoleman:
My first pull request only updated the commands: https://github.com/jellyfin/jellyfin.org/pull/141

This one contains the same changes as the first as well updated versions of the notes that follow the commands.

joshuaboniface (approved):
Seems sensible to me; TIL about the `deb822` format. There's a few more enhancements I'll want to make to this after but it's a good start.

Issue #146, opened 2022-09-07
Logo on branding page not visible in dark theme


nielsvanvelzen:
The logo section on the branding page doesn't work with the dark theme enabled. The logo uses dark text. Fix would be to add a white background for the images or change the logo depending on the theme.

http://next.jellyfin.org/docs/general/contributing/branding#logo

PR #147, opened 2022-09-07
logo uses white text on dark mode

leewan-09:
✅ Change the logo depending on the theme ( issue #146 )

> Toggle the visibility of an image using the path fragment


src/css/custom.css
```diff
@@ -112,6 +112,12 @@ a.pills__item:not(.pills__item--active) {
   fill: var(--ifm-font-color-base);
 }
 
+/* To toggle the visibility of an image using the path fragment */
+[data-theme='light'] img[src$='#gh-dark-mode-only'],
+[data-theme='dark'] img[src$='#gh-light-mode-only'] {
```
thornbill, line 117: I like this solution, but what does "gh" stand for in the hash? Maybe `hidden--light-mode` would be a better name to align with the desktop/mobile utility classes. 🙂

thornbill (changes requested):

src/css/custom.css
leewan-09: Good suggestion 👍. Actually, `gh` means `Github`. https://docusaurus.io/docs/next/markdown-features/assets#github-style-themed-images

leewan-09:
I have the CSS class name. It is much cleaner.

thornbill (approved):

PR #150, opened 2022-09-09
fixed broken reference links

leewan-09:
I notice few broken reference links. In order to fix I changed `<sup></sup>` to Unicode.


nielsvanvelzen:
Docusaurus should support Markdown footnotes (see the sample from GitHub [^1]) so maybe we can use that?

[^1]: https://github.blog/changelog/2021-09-30-footnotes-now-supported-in-markdown-fields/

leewan-09:
> 
We could use that, but there are no reference links given for the footer notes.

image

thornbill:
Good point. Since these aren't strictly just reference links, I think keeping `<sup>` and changing the links to `<a>` tags would be better imho.

PR #175, opened 2022-09-17
Neater display of text on the Contribute page

alanazar:
Not sure if I used the \n correctly, appreciate any comments

Since it is not a paragraph, it's better we break the line where it makes sense so it looks neater.
Before \ After example:

image


src/pages/contribute.tsx
```diff
@@ -14,16 +14,16 @@ export default function Contribute() {
     <Layout title='How to Contribute'>
       <Hero title='How to Contribute'>
         <p className='hero__text'>
-          Jellyfin is a community project run by volunteers. We&apos;re always looking for additional help.
+          Jellyfin is a community project run by volunteers.\n We are always looking for additional help.
```
thornbill, line 17: Sorry I thought I had commented on this already... maybe it would make more sense to use `<br />` here for the line break?

```suggestion
          Jellyfin is a community project run by volunteers.
          <br />
          We are always looking for additional help.
```

alanazar:
@thornbill thanks for the advice, I had tried googling it but couldn't get the br to work. I also know now how to test it on my end so when I get some time I'll work on the rest

PR #184, opened 2022-09-20
Add client filtering options

### LINKED ISSUE #133: Add additional filtering options to clients page
Add options to filter clients by device type or platform

thornbill:
Adds device and platform filtering options to the Clients page.

**Default (Filters Collapsed)**
Screenshot 2022-09-20 at 15-42-19 Clients Jellyfin

**Filters Applied**
Screenshot 2022-09-20 at 17-23-30 Clients Jellyfin

**Filters Applied w. No Results**
Screenshot 2022-09-20 at 17-23-46 Clients Jellyfin

Fixes #133

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -99,6 +99,12 @@ a.pills__item:not(.pills__item--active) {
   gap: 0.25rem;
 }
 
+.card--outline {
+  background-color: transparent;
+  border: 2px solid #172138;
+  box-shadow: none;
+}
+
 .row--center {
   justify-content: center;
 }
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -92,6 +92,13 @@ a.pills__item:not(.pills__item--active) {
   padding: 0 var(--ifm-pre-padding);
 }
 
+.button--icon {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  gap: 0.25rem;
+}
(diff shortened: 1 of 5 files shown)


LePips (approved):

nielsvanvelzen:
It feels a bit messy to me. Normally when you can expand a section it outlines that section with a background/border/whatever so you can visually see what opened/closed. Maybe adding the filters in a card could help with this?

Another approach would be to always make the filters visible but change them to dropdowns.

thornbill:
Yeah those are definitely fair points. We are kind of hitting a limitation of what docusaurus ships out of the box for components here. I'm sure eventually we will need to pull in something like bulma to make up the difference, but I'm trying to avoid it for now. :sweat_smile: 

Any form components are not going to look great because they aren't styled out of the box. I think that rules out the dropdown option.

The default card won't work well since the active pill background color is the same as the card background. Maybe a variation of the card with just an outline would work?

nielsvanvelzen:
An outline might work too yeah. I'll approve a "good enough for now" solution :p

thornbill:
Screenshot 2022-09-23 at 15-45-10 Clients Jellyfin

I added a subtle outline around the filters (uses the same color as the card background) and centered the filter options.

nielsvanvelzen (approved):

PR #209, opened 2022-10-05
Add sass support

### LINKED ISSUE #91: Add sass support
Docusaurus can be configured to support sass which would be a nice development feature.

Refs: https://docusaurus.io/docs/styling-layout/#sassscss

sylvainmetayer:
This should close #91


.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 16.14.0
```
nielsvanvelzen, line 1: What is this file for?

sylvainmetayer: It is a file used by [asdf](https://github.com/asdf-vm/asdf) to define versions of tools, node in this case. It is like nvm but not specific for node.

I forgot to exclude it when I made the commit, I will update my commit and remove it later today.

sylvainmetayer: I updated my commit to remove it.

nielsvanvelzen (approved):

thornbill (approved):

PR #228, opened 2022-10-19
Update site url configuration

thornbill:
Update the site url in the docusaurus config and CNAME file

**NOTE:** This should probably not be merged until immediately before we make the necessary DNS changes

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -2,7 +2,7 @@
 module.exports = {
   title: 'Jellyfin',
   tagline: 'The Free Software Media System',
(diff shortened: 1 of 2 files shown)


nielsvanvelzen (approved):
next.jellyfin.org is also mentioned in the README, should that be updated in this PR or separately?

thornbill:
Yeah I was going to update that in a separate PR since that change can go in at anytime, but this one needs to go in right before we cut over.

joshuaboniface (approved):

PR #237, opened 2022-10-20
Add jellyfin-vue to client list

ferferga:
Adds Jellyfin Vue to the client page as an "official beta"

Similar in fashion to #232

### DIFF
--- a/src/data/clients.ts
+++ b/src/data/clients.ts
@@ -118,6 +118,34 @@ export const Clients: Array<Client> = [
     ]
   },
   {
+    id: 'jellyfin-vue',
+    name: 'Jellyfin Vue',
+    description: 'A modern web client for Jellyfin based on Vue',
+    clientType: ClientType.OfficialBeta,
+    deviceTypes: [DeviceType.Mobile, DeviceType.Desktop],
+    licenseType: LicenseType.OpenSource,
+    platforms: [Platform.Browser],
+    primaryLinks: [
+      {
+        id: 'browser',
+        name: 'Open in Browser',
+        url: 'https://jf-vue.pages.dev/'
+      },
+      {
+        id: 'docker-ghcr',
+        name: 'Docker',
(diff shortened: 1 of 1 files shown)


thornbill:
I knew I was forgetting something when I added Swiftfin 😅

nielsvanvelzen:
Is the Vue client really in a public beta? It's own README calls it "experimental" and there are no GH releases either so it's not easy to get started with the client. In comparison, the Swiftfin beta releases are tagged and easy to use (sign up for Testflight).

ferferga:
@nielsvanvelzen 

1. Can be used rightaway at https://jf-vue.pages.dev/ (this link is in the PR as well)
2. There are no tags because we never found good reasons to, as everything around it was evolving so quickly. However, we have Docker images that are tagged for every commit sha automatically. At the end of the day it's like web, where in GH we don't provide any built asset, just sources, expecting everyone to build their own or download them from our servers. 
3. There are also Tauri builds for installing it as a native app.

nielsvanvelzen:
Oh no need to defend it, just a "yes" would have been enough :p. For Swiftfin it's quite clear that it's in beta and I didn't find it obvious when looking at the jf-vue repo so that's why I asked.

ferferga:
@nielsvanvelzen Oh no, completely my fault if I came across like that! Sorry if my wording sounded rough because all my intention was to shed some light to your points with a more detailed explanation because you actually are right and we've been lazy at releases and our docs could be improved by a lot :).

thornbill:
I am questioning the placement of this in the list a bit... I tried to generally sort the clients by platform  and it seems to be in the middle of TV clients currently. 😄

ferferga:
@thornbill The ordering of the elements was something I was trying to figure out myself, but didn't found a correlation 😅, so just kept the two official betas together (which are also together the official clients). Thanks for pointing out the actual ordering because not figuring it out was driving me mad :P.

Sounds good to you to place it before Preserve?

thornbill:
Yeah that sounds good to me!

thornbill (approved):

PR #248, opened 2022-10-25
Add line breaks and note to copy cat as a block

anthonylavado:
I don't know if our system respects line breaks as added, or if I need to add a `<br/>` or similar.

Adding some line breaks around the `cat` command to make it clear that it is a large block. Reworded the note to mention the cat command specifically (and it is no longer the third in the list anyway).


nielsvanvelzen:
Since this is just HTML I'm quite sure those `<br />` tags need to be added. There's also a few liner issues that need fixing.

nielsvanvelzen:
@anthonylavado you'll need to rebase or merge master to fix the merge conflict.

PR #274, opened 2022-11-21
Update docker-compose.yml example

cttipton:
Add optional `extra_hosts:` argument to docker-compose example, which is necessary to pass docker healthcheck when running in host network mode.  Otherwise results in false-positive "unhealthy" warning in `docker container ls` output, as localhost is unresolved without setting host.docker.internal:host-gateway.  At least, this is the case for me on linux, running:

~/self-hosted/jellyfin
⟩ pacman -Q | grep docker
docker 1:20.10.21-1
docker-compose 2.12.2-1

~/self-hosted/jellyfin
⟩ pacman -Q | grep container
containerd 1.6.10-1
containers-common 1:0.50.1-2

### DIFF
--- a/docs/general/administration/installing.md
+++ b/docs/general/administration/installing.md
@@ -115,6 +115,9 @@ services:
     # Optional - alternative address used for autodiscovery
     environment:
       - JELLYFIN_PublishedServerUrl=http://example.com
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

nielsvanvelzen (approved):

nielsvanvelzen:
Thanks!

PR #291, opened 2022-12-09
Combine client and server download pages

thornbill:
Combines the client and server download pages to help with reports of people accidentally downloading the server when they mean to download a client (hopefully)...

Screenshot 2022-12-09 at 01-26-29 Clients Jellyfin

Screenshot 2022-12-09 at 01-26-51 Downloads Jellyfin


thornbill:
Screenshot 2022-12-09 at 02-14-47 Clients Jellyfin

Updated header margins

nielsvanvelzen:
Maybe the stable/unstable toggle should be in a filters thingy too so it's consistent?

thornbill:
I don't know... that might be weird since there are only the 2 options :man_shrugging:

nielsvanvelzen:
With #294 it might make sense

joshuaboniface:
> With #294 it might make sense, add a second toggle for "easy install" and "advanced install" instructions

I explicitly didn't want to do this, hiding it behind another button is just cumbersom esoecially when it only applies to a single platform.

nielsvanvelzen (approved):

PR #294, opened 2022-12-10
Add curl-bash type installer type for Debuntu

joshuaboniface:
This should hopefully make things easier for new users who might be offput or confused by how complex the install instructions are.

The script is visible at https://repo.jellyfin.org/install-debuntu.sh

I'm game to move it somewhere else more "visible", or just leave it there in the root of the actual repository, depending on preferences. The caveat of this is that the selection of Ubuntu/Debian releases and architectures is static in the file, so version-controlling it would make quick changes harder, but I'm flexible.

### DIFF
--- a/src/data/downloads.tsx
+++ b/src/data/downloads.tsx
@@ -58,7 +58,7 @@ export const Downloads: Array<Download> = [
               <code>{`curl https://repo.jellyfin.org/install-debuntu.sh | sudo bash`}</code>
             </pre>
             <p>
-              If you don't have <code>curl</code> installed, you can try <code>wget</code> instead:
+              If you do not have <code>curl</code> installed, you can try <code>wget</code> instead:
             </p>
             <pre>
               <code>{`wget -O- https://repo.jellyfin.org/install-debuntu.sh | sudo bash`}</code>
--- a/src/data/downloads.tsx
+++ b/src/data/downloads.tsx
@@ -55,21 +55,15 @@ export const Downloads: Array<Download> = [
         details: (
           <>
             <pre>
-              <code>
-                {`curl https://repo.jellyfin.org/install-debuntu.sh | sudo bash`}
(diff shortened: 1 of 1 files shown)


src/data/downloads.tsx
```diff
@@ -54,15 +54,33 @@ export const Downloads: Array<Download> = [
...
-curl -fsSL https://repo.jellyfin.org/$( awk -F'=' '/^ID=/{ print $NF }' /etc/os-release )/jellyfin_team.gpg.key | sudo gpg --dearmor -o /etc/apt/keyrings/jellyfin.gpg
+DISTRO="$( awk -F'=' '/^ID=/{ print $NF }' /etc/os-release )"
+CODENAME="$( awk -F'=' '/^VERSION_CODENAME=/{ print $NF }' /etc/os-release )"
+curl -fsSL https://repo.jellyfin.org/\${DISTRO}/jellyfin_team.gpg.key | sudo gpg --dearmor -o /etc/apt/keyrings/jellyfin.gpg
 cat <<EOF | sudo tee /etc/apt/sources.list.d/jellyfin.sources
 Types: deb
-URIs: https://repo.jellyfin.org/$( awk -F'=' '/^ID=/{ print $NF }' /etc/os-release )
-Suites: $( awk -F'=' '/^VERSION_CODENAME=/{ print $NF }' /etc/os-release )
+URIs: https://repo.jellyfin.org/\${DISTRO}
+Suites: \${CODENAME}
```
joshuaboniface, line 83: I don't know for sure if this is the correct escape here or if it even matters, so worth checking.

nielsvanvelzen (approved):

src/data/downloads.tsx
joshuaboniface: For future benefit: yes, this was correct :-)

PR #334, opened 2023-01-12
Syncloud self-hosting

cyberb:


anthonylavado:
While we appreciate the contribution, we're currently not accepting any third party cloud services for our documentation. If this changes in future, we'll let you know.

cyberb:
this is not a cloud service this is pure self-hosting by users themselves, cloud is only dns, if it makes any difference,

PR #357, opened 2023-01-22
Initial changes for developer site

nielsvanvelzen:
**This PR targets the new "developers" branch so we don't update the production site until these changes are more stable.**

---

This is the initial PR for a new developers site. It's basically a new section in our existing documentation instead of using a separate repository. This helps with maintenance because we don't need to manage two repositories and stuff like the theme stays in sync.

Changes in this PR include:

- Create new "developers" docs section
  - Adds /docs/developers
  - Updates the header menu (removed contributing, added "developers" after docs)
  - Adds a separate sidebar for developers docs
  - Moves the api.jellyfin.org link to developers sidebar
- Move existing docs pages related to development to new developers section
- Add empty developers index

image

Screenshot shows the empty developers page and the menu changes.


nielsvanvelzen:
Since this isn't merged yet I didn't have a place to PR the first content to, so I've put it in a gist for now:

https://gist.github.com/nielsvanvelzen/ea047d9028f676185832e51ffaf12a6f

ThibaultNocchi:
I agree with the opinion of adding this to the main site. Easier to maintain, we can avoid having X websites for different things, etc. We also don't have that much dev doc, so I don't think it'd be warranted to create a new repo and/or website for this feature. If so, why wouldn't the doc be split out?

thornbill (approved):
I do still think a dedicated developer site would be better in the long run to be easily discoverable and searchable. It wouldn't need to be anything fancy really (we would probably be fine with just using docusaurus out of the box mostly). I understand the maintenance concerns though. 

Regardless I think this is a fine place to start. 🚀

Issue #372, opened 2023-02-03
Don't suggest obsolete/dangerous X-XSS-PROTECTION anymore


ghost:
The X-XSS-PROTECTION header is deprecated and dangerous. It can cause perfectly secure websites to be vulnerable to XSS attacks. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection

It is also not standards compliant and doesn't have uniform behaviour across browsers. The only safe value as of today is to explicitly disable it by setting it to 0, not 1

See hardenize recommendation

https://www.hardenize.com/report/jellyfin.org/1675457811#www_xxssp

Also see: https://news.ycombinator.com/item?id=20472947
https://thexssrat.medium.com/x-xss-protection-headers-protection-or-vulnerability-bc7213951320

PR #373, opened 2023-02-03
modernize headers, disable XSS protection

randomhydrosol:
See https://github.com/jellyfin/jellyfin.org/issues/372
Signed-off-by: randomhydrosol <randomhydrosol@glassrom.org>


joshuaboniface (approved):
I think this makes sense.

PR #377, opened 2023-02-04
Add JellyBook to list of clients

Kara-Zor-El:
- The purpose of this PR is to add my client for Jellyfin, JellyBook to the client list
 - I am a little unsure of if there is a order in which the list is in since I didn't see any pattern to how the list was organized so if I need to change it, let me know


nielsvanvelzen (approved):
Changes look fine. Always good to see new clients in the Jellyfin ecosystem! Don't forget to update the links here when it releases to GPlay or goes out of TestFlight.

PR #396, opened 2023-02-11
Replace @docusaurus/preset-classic with manual plugin configurations

nielsvanvelzen:
This PR targets the developers branch.

----

With this PR the classic preset is removed in favor of manual plugin/theme configurations in preparation for a second blog.

This isn't strictly necessary for multi-blog etc. but it makes the config easier to read when we do, it also shaves of _some_ of the other plugins we don't use from the dependency graph which should help a bit with the node_modules folder size.

Other changes include:

- Move the redirects for the redirect plugin to their own file (easier to maintain that way)
- Add `@type` comments for IDE assistance to plugins/themes
- Add second developer blog (commented out) to confirm multi blog works (and it does, even creates a separate RSS feed)

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -9,6 +9,7 @@ module.exports = {
   favicon: 'images/favicon.ico',
   organizationName: 'jellyfin',
   projectName: 'jellyfin.org',
+  /** @type {import('@docusaurus/types').ThemeConfig} */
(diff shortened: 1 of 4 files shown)


nielsvanvelzen:
image

Example of the dev blog with this PR

Shadowghost (approved):

PR #416, opened 2023-02-21
Add client to "Clients" page

manuel-rw:
Hi 👋

I would like to add my open source project to the list of available clients.
Please let me know if I should change anything or if the commit message is fine.

### DIFF
--- a/src/data/clients.ts
+++ b/src/data/clients.ts
@@ -642,9 +642,9 @@ export const Clients: Array<Client> = [
     ]
   },
   {
-    id: 'discord-music',
-    name: 'Discord Music Bot for Jellyfin',
-    description: 'A Discord bot that allows playing your Jellyfin music library in Discord voice channels.',
+    id: 'discord-music-kgt1',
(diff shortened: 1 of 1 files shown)


nielsvanvelzen (commented):
Thanks for adding! I think we should add something like ` (by [username])` to the names of both discord music bots to avoid confusion between the two.

manuel-rw:
> Thanks for adding! I think we should add something like ` (by [username])` to the names of both discord music bots to avoid confusion between the two.

Yes, that is a good idea.
I will add that as soon as possible.

manuel-rw:
image

This is how it looks on my development server. Do you think that the "by XYZ" should be in the title instead, or is this "enough"?

nielsvanvelzen:
When scanning a page I mostly look at the titles, so IMO the author name should be in the title.

manuel-rw:
Okay, makes sense. I'll also fix the eslint warning...

manuel-rw:
image

nielsvanvelzen (approved):
Thanks!

PR #417, opened 2023-02-21
Make client icons on homepage clickable

iwalton3:
This PR:
 - Allows you to click on all of the client icons on the homepage except web.
 - Persists the filters on the client page into the URL search param.

I excluded web from the clickable links because it would either show an empty page, or, if showing all options would mislead users into thinking jellyfin-vue was the main web client.

### DIFF
--- a/src/pages/downloads/clients/index.tsx
+++ b/src/pages/downloads/clients/index.tsx
@@ -46,16 +46,16 @@ export default function ClientsPage({ recommended = true }: { recommended?: bool
   const [isFiltersVisible, setIsFiltersVisible] = useState(false);
 
   const setFilter = (filter: ClientFilter) => {
-    let search: string[] = [];
+    const search = new URLSearchParams();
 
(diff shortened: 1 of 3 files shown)


src/pages/downloads/clients/index.tsx
```diff
@@ -31,14 +32,35 @@ function toggleValue<Type>(array: Type[], value: Type): Type[] {
...
+
+    if (filter.deviceTypes.length > 0) {
+      search.push(`type=${filter.deviceTypes.join(',')}`);
+    }
+    if (filter.platforms.length > 0) {
+      search.push(`platform=${filter.platforms.join(',')}`);
+    }
+    history.push({
+      search: '?' + search.join('&')
+    });
```
nielsvanvelzen, line 59: Why not use URLSearchParams to construct the query parameters here? Should be more safe because it applies url encoding etc.

nielsvanvelzen (approved):

PR #418, opened 2023-02-22
Update iis.md

kimboslice99:
A couple rules for IIS, redirect to web/ then rewrite to web/index.html, similar trick used on nginx... also prevents back button bug jellyfin/jellyfin-web#1848


nielsvanvelzen (approved):
Thanks!

PR #427, opened 2023-02-28
Explicitly mention provider identifiers in file names

AlexAltea:
As discussed on the General Chat, this was a confusing point to me, as this seems to imply such tags were only available in folders.


nielsvanvelzen (approved):
Thanks!

PR #432, opened 2023-03-05
Update hardware-acceleration.md

icedterminal:
AMD's AMF installation is no longer supported with the current documentation steps. An error is thrown regarding package dependencies / missing packages - depending on what a user tries.

These updated steps rectify that by using the supported method the script now provides. The example command provided installs the Pro variant with AMF for all supported GPUs.


nielsvanvelzen (commented):
Left some suggestions on how to fix the linter issues

icedterminal:
> Left some suggestions on how to fix the linter issues

Thank you

PR #487, opened 2023-03-31
Added directory exclusion docs page

TimGels:
Added a page to document the usage of `.ignore` files. Also included a notice about it currently not working for certain directories.

### DIFF
--- a/docs/general/server/media/excluding-directory.md
+++ b/docs/general/server/media/excluding-directory.md
@@ -1,0 +1,37 @@
+---
+uid: server-media-excluding-directory
+title: Excluding a directory
+---
+
+# Excluding a directory
+
+Directories, and their respective media files, can be excluded from the library scan. This can be achieved by placing a `.ignore` file in the directory you want to exclude. As a result, the directory, and it's subdirectories, will be excluded from being scanned and shown to the end user.
+
+:::tip
+
+When placing a `.ignore` file inside a directory, make sure to refresh/ rescan the metadata afterwards for it to be applied.
+
+:::
+
+The example given below results in the directories `Season 03` and `OST` being ignored:
+
+```txt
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

Shadowghost:
You might want to add a comment describing that adding a .ignore file after a library scan will lead to the files being removed from the library on the next scan.

TimGels:
> You might want to add a comment describing that adding a .ignore file after a library scan will lead to the files being removed from the library on the next scan.

Yea good one. I see how it can be confusing for users when they add the ignore file and expect the directories to become hidden while they still need to do a scan. I will add this tomorrow.

docs/general/server/media/excluding-directory.md
```diff
@@ -0,0 +1,37 @@
...
+    │   ├── ...
+    ├── Season 03
+    │   ├── .ignore
+    │   └── ...
+    └── OST
+        ├── .ignore
+        └── ...
+```
+
+:::note
```
nielsvanvelzen, line 33: We may want to use `:::caution` here

docs/general/server/media/excluding-directory.md
```diff
@@ -0,0 +1,37 @@
+---
+uid: server-media-excluding-directory
+title: Excluding a directory
+---
+
+# Excluding a directory
+
+Directories, and their respective media files, can be excluded from the library scan. This can be achieved by placing a `.ignore` file in the directory you want to exclude. As a result, the directory, and it's subdirectories, will be excluded from being scanned and shown to the end user.
+
+:::note
```
nielsvanvelzen, line 10: This could be a `:::tip`

nielsvanvelzen (approved):

PR #513, opened 2023-04-18
Update to .NET 7

endrl:
Just stumbled upon this


Shadowghost:
Our current stable version is still built with .NET 6 so I'd say we should wait until 10.9 comes out until we change it.

endrl:
You are right. I didn't thought about that. I just wanted to point out that build instructions don't work, as they point to master.

nielsvanvelzen:
The development page can be updated right now because the master branch is already on .NET 7. The build from source instructions actually checks out from master too (and it looks like there are a bunch of redundant steps.. we should really rewrite that page). Anyway, I believe both changes should be fine to merge before 10.9 is released.

Shadowghost (approved):

PR #527, opened 2023-05-04
Add flatpak

istori1:


src/data/downloads.tsx
```diff
@@ -151,6 +151,55 @@ sudo apt install jellyfin`}
...
+      {
+        id: 'flatpak-unstable-button',
+        name: 'Install Instructions',
+        details: (
+          <p>
+            <b>Note:</b> Only stable releases are distributed in Flathub.
+          </p>
+        )
+      }
+    ],
```
nielsvanvelzen, line 194:
```suggestion
    unstableButtons: [],
```

Leave the unstableButtons empty is there is no support for unstable builds.

src/data/downloads.tsx
```diff
@@ -151,6 +151,55 @@ sudo apt install jellyfin`}
...
+    platforms: [Platform.Linux],
+    description: 'Install Jellyfin via Flathub.',
+    stableButtons: [
+      {
+        id: 'flatpak-stable-button',
+        name: 'Install Instructions',
+        details: (
+          <>
+            <p>
+              <b>Note:</b> Make sure Flatpak and Flathub are already setup before installing.
```
nielsvanvelzen, line 169: This note seems a bit out of place, why would someone explicitly opening the flatpak instructions not have flatpak installed/setup?

nielsvanvelzen:
@thornbill right now we have the os types linux, windows, mac and docker. We might want to rename docker to "container" or something so we can add both docker and flatpak in that section. What do you think?

thornbill:
I'm inclined to think it should just be under Linux. 🤷‍♂️ afaik that is the only place where flatpaks are supported unlike docker that can run under other OSes.

nielsvanvelzen (approved):

src/data/downloads.tsx
```diff
@@ -151,6 +151,42 @@ sudo apt install jellyfin`}
...
+      {
+        id: 'flatpak-stable-button',
+        name: 'Install Instructions',
+        details: (
+          <>
+            <pre>
+              <code>{`flatpak install flathub org.jellyfin.JellyfinServer`}</code>
+            </pre>
+            <p>
+              <b>Note:</b> If you are running on an Intel GPU an adtional extension is required for HDR Tone mapping.
```
nielsvanvelzen, line 172:
```suggestion
              <b>Note:</b> If you are running on an Intel GPU an additional extension is required for HDR Tone mapping.
```

nielsvanvelzen (approved):

PR #536, opened 2023-05-17
Caddy Subpath Redirect

felix920506:
set Caddy to redirect example.com/jellyfin to example.com/jellyfin/


nielsvanvelzen:
Thanks!

PR #552, opened 2023-06-06
Update transcoding.md

d6gNMGiKx6k8:
Previously it stated there are three types of transcoding, but there were four items in the list.  The first item was for direct play, so I updated to clarify that there are four types of playback, yet only three of which perform transcoding.

### DIFF
--- a/docs/general/server/transcoding.md
+++ b/docs/general/server/transcoding.md
@@ -20,9 +20,9 @@ This option will manually set the number of threads to use when transcoding. If 
 
 ## Types of Transcoding
 
-There are three types of transcoding. The type being used will be listed in the dashboard when playing a file. They are ordered below from lowest to highest load on the server:
+There are four types of playback; three of which involve transcoding.  The type being used will be listed in the dashboard when playing a file. They are ordered below from lowest to highest load on the server:
 
-- Direct Play: Delivers the file with no modifications. Almost no additional load on the server.
+- Direct Play: Delivers the file without transcoding.  There is no modification to the file and almost no additional load on the server.
 - Remux: Changes the container but leaves both audio and video streams untouched.
 - Direct Stream: Transcodes audio but leaves original video untouched.
 - Transcode: Transcodes the video stream.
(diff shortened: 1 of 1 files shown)


docs/general/server/transcoding.md
```diff
@@ -20,9 +20,9 @@ This option will manually set the number of threads to use when transcoding. If
 
 ## Types of Transcoding
 
-There are three types of transcoding. The type being used will be listed in the dashboard when playing a file. They are ordered below from lowest to highest load on the server:
+There are four types of playback; three of which involve transcoding.  The type being used will be listed in the dashboard when playing a file. They are ordered below from lowest to highest load on the server:
```
Shadowghost, line 23:
```suggestion
There are four types of playback; three of which involve transcoding. The type being used will be listed in the dashboard when playing a file. They are ordered below from lowest to highest load on the server:
```

docs/general/server/transcoding.md
```diff
@@ -20,9 +20,9 @@ This option will manually set the number of threads to use when transcoding. If
...
 
-- Direct Play: Delivers the file with no modifications. Almost no additional load on the server.
+- Direct Play: Delivers the file without transcoding.  There is no modification to the file and almost no additional load on the server.
```
Shadowghost, line 25:
```suggestion
- Direct Play: Delivers the file without transcoding. There is no modification to the file and almost no additional load on the server.
```

Shadowghost (changes requested):

docs/general/server/transcoding.md
d6gNMGiKx6k8: Ha!  I noticed the space after I submitted and that it was inconsistent with this new-age, single space after a sentence, format the kids are using these days.  I'd have updated it previously, but this is my first fork, commit, and pull request in my life.  I wanted to take it slow and not risk screwing things up further in trying to make the update!  Many thanks!

Shadowghost (approved):

docs/general/server/transcoding.md
Shadowghost: It's still wrong, no?

nielsvanvelzen (changes requested):
Feedback is not (yet) resolved

nielsvanvelzen (approved):

PR #580, opened 2023-06-28
Move status link from header to footer

nielsvanvelzen:
It's nice we have it now but it's not relevant for most of the people visiting our website. It may also confuse users thinking we are a streaming service or provide hosted instances. Moving the status link to the footer which is in my opinion the best place for a status page link.


thornbill (approved):

thornbill:
That wasn't the "merge" button I meant to click :sweat_smile:

PR #582, opened 2023-06-29
Update nginx.md

tempookian:


nielsvanvelzen:
There are multiple `proxy_pass` lines across the different config samples and you only changed a single one. Can you describe why you made this change?

tempookian:
my apologies for the lack of explanation. I tried on a server that is being proxied through Cloudflare. The way I changed the line worked. The other methods (lines in the post) I have not tried. Hence, I did not change them

joshuaboniface:
Can you provide more details on why this change is being made and what it fixes? From my understanding, NGiNX will always start the proxy pass with `/` because the incoming URL would always start with `/`, so this shouldn't be necessary.

nielsvanvelzen:
The trailing slash should not be necessary. Closing this PR as stale.

PR #412, opened 2023-02-20
Update caddy.md

jaykup26:
Reddit posts are outdated, but corrections are in the comments.  Author of the post lost his account.  Adding a thorough YouTube guide.

PR #607, opened 2023-07-24
Add Warning message on DNS Tokens

felix920506:
This PR does the following:
1. Add warning to warn against giving Caddy DNS Tokens 
2. Fix some other minor issues with the Caddy page
3. Remove references to other community links

Reasoning for each of the actions:
1. Supplying Caddy with DNS Tokens can be dangerous since this grants Caddy the privileges to edit more DNS records than it needs to in most cases. (Eg. Being able to edit DNS records of the entire domain or Account) This can be dangerous in the event of an attack as compromising this token means attackers can compromise your entire domain or even other domains under the same account
2. Clarify the subpath so it is less confusing
3. The Youtube Video contains steps that can be dangerous (as pointed out in the first point). The Reddit links are outdated (As stated in #412 )

### DIFF
--- a/docs/general/networking/caddy.md
+++ b/docs/general/networking/caddy.md
@@ -30,7 +30,7 @@ example.com {
 Please proceed with caution when using this option:
 
 - This will **NOT** automatically update your DNS records if you have a dynamic IP.
-- This is **NOT** required for automatic SSL to work in most cases.
+- This is **NOT** required for automatic HTTPS to work in most cases.
 - Misconfiguration can lead to **compromised domains and/or accounts**.
 - API keys should only be granted the least permissions required for the application to function.
 


srcrist:
I think it's a good idea to clarify that the API tokens do not enable caddy to automatically update DDNS, and I think both the subpath changes and the youtube removals are fine as well. Pursuant to our conversation yesterday, though, I think the language here regarding the use of API tokens generally is too strong. Services like Cloudflare provide ways to scope limit specific tokens both by purpose and IP, and the security risk of proper usage is minimal. Let's Encrypt's own tooling, like Certbot, also require the same keys.

Rather than discouraging their wholesale use, I would just clarify that there can be security risks to using tokens, and advocate that users ensure that tokens are correctly scope limited and handled carefully, and note that the use of global tokens which provide unfettered access to entire accounts, specifically, should be strongly discouraged.

docs/general/networking/caddy.md
```diff
@@ -13,6 +13,30 @@ You can reverse proxy to Jellyfin either with or without a config file, and eith
...
+    }
+    // highlight-end
+}
+```
+
+Please proceed with caution when using this option:
+
+- This will **NOT** automatically update your DNS records if you have a dynamic IP.
+- This is **only** necessary if you **cannot** open ports 80 and 443
+- Misconfiguration can lead to **compromised domains and/or accounts**.
```
TimGels, line 34: Is there a source or small explanation that you can back this up with? In what way can it lead to a compromise for example. Ignore this comment if this might be too much.

felix920506: I think it would be too long as this section is intended to be a short warning

felix920506: And also I think it should be very obvious simply by looking at what permissions are available and can be enabled for the API keys.

docs/general/networking/caddy.md
```diff
@@ -13,6 +13,30 @@ You can reverse proxy to Jellyfin either with or without a config file, and eith
 
 If you aren't familiar with Caddy yet, check out its [Getting Started](https://caddyserver.com/docs/getting-started) guide.
 
+:::caution
+There are a some guides that have a Caddyfile which includes the `tls` section with a DNS Provider API Key, as shown the following example.
```
Shadowghost, line 17:
```suggestion
There are a some guides that have a Caddyfile which includes a `tls` section with the DNS provider's API key as shown the following example.
```

docs/general/networking/caddy.md
```diff
@@ -13,6 +13,30 @@ You can reverse proxy to Jellyfin either with or without a config file, and eith
...
+        dns <DNS Provider> <API Token>
+    }
+    // highlight-end
+}
+```
+
+Please proceed with caution when using this option:
+
+- This will **NOT** automatically update your DNS records if you have a dynamic IP.
+- This is **only** necessary if you **cannot** open ports 80 and 443
```
Shadowghost, line 33:
```suggestion
- This is **only** necessary if you **cannot** open ports 80 and 443.
```

docs/general/networking/caddy.md
```diff
@@ -13,6 +13,30 @@ You can reverse proxy to Jellyfin either with or without a config file, and eith
...
+- Misconfiguration can lead to **compromised domains and/or accounts**.
+- API Tokens should only be granted the least permissions for the application to function.
```
Shadowghost, line 35:
```suggestion
- API keys should only be granted the least permissions required for the application to function.
```

docs/general/networking/caddy.md
```diff
@@ -63,7 +87,7 @@ To do this, first configure Jellyfin to use a base path.
 If you already have access to the web interface, go to `Admin > Networking` and enter a path like `/jellyfin` in the Base URL field.
 If not, you may instead go to `<Configuration Directory>/network.xml` and modify the value of `<BaseUrl>` according to your needs. For information on the directory location, please consult the [configuration documentation](/docs/general/administration/configuration#configuration-directory).
 You might have to restart the Jellyfin server for this to take effect.
-Then simply give the `reverse_proxy` directive a path matcher.
+Then simply give the `reverse_proxy` directive a path matcher. The path should be the same as the Base URL you entered into Jellyfin's settings. The following example is for a server that is under `example.com/jellyfin`.
```
Shadowghost, line 90:
```suggestion
Then simply give the `reverse_proxy` directive a path matcher. The path should be the same as the `Base URL`  you entered into Jellyfin's settings. The following example is for a server that is under `example.com/jellyfin`.
```

srcrist:
My only remaining thought on the port question is just that it isn't strictly true that the only reason to use the tokens would be if you're unable to forward the ports. You might also need to use DNS verification if you already have a web server like nginx or apache running on the standard ports, preventing caddy from using web server authentication. You also HAVE to use DNS verification if you want a wildcard certificate, as LE won't issue wildcards via web verification.

Personally, I would just remove that line entirely. It isn't really necessary to communicate that there is a risk.

felix920506:
I changed the port section to a simple clarification that this is not required for automatic SSL to work.

docs/general/networking/caddy.md
```diff
@@ -13,6 +13,30 @@ You can reverse proxy to Jellyfin either with or without a config file, and eith
 
 If you aren't familiar with Caddy yet, check out its [Getting Started](https://caddyserver.com/docs/getting-started) guide.
 
+:::caution
+There are a some guides that have a Caddyfile which includes a `tls` section with the DNS provider's API key as shown the following example.
```
Shadowghost, line 17:
```suggestion
There are a some guides that have a Caddyfile which includes a `tls` section with the DNS provider's API key as shown in the following example.
```

docs/general/networking/caddy.md
```diff
@@ -13,6 +13,30 @@ You can reverse proxy to Jellyfin either with or without a config file, and eith
...
+        dns <DNS Provider> <API Token>
+    }
+    // highlight-end
+}
+```
+
+Please proceed with caution when using this option:
+
+- This will **NOT** automatically update your DNS records if you have a dynamic IP.
+- This is **NOT** required for automatic SSL to work in most cases.
```
Shadowghost, line 33:
```suggestion
- This is **NOT** required for automatic TLS to work in most cases.
```

docs/general/networking/caddy.md
```diff
@@ -13,6 +13,30 @@ You can reverse proxy to Jellyfin either with or without a config file, and eith
...
+- Misconfiguration can lead to **compromised domains and/or accounts**.
+- API keys should only be granted the least permissions required for the application to function.
+
+Please read [Let's Encrypt Documentation](https://letsencrypt.org/docs/challenge-types/) for more info.
```
Shadowghost, line 37:
```suggestion
Please read the [Let's Encrypt documentation](https://letsencrypt.org/docs/challenge-types/) for more info.
```

docs/general/networking/caddy.md
```diff
@@ -63,7 +87,7 @@ To do this, first configure Jellyfin to use a base path.
 If you already have access to the web interface, go to `Admin > Networking` and enter a path like `/jellyfin` in the Base URL field.
 If not, you may instead go to `<Configuration Directory>/network.xml` and modify the value of `<BaseUrl>` according to your needs. For information on the directory location, please consult the [configuration documentation](/docs/general/administration/configuration#configuration-directory).
 You might have to restart the Jellyfin server for this to take effect.
-Then simply give the `reverse_proxy` directive a path matcher.
+Then simply give the `reverse_proxy` directive a path matcher. The path should be the same as the `Base URL`  you entered into Jellyfin's settings. The following example is for a server that is under `example.com/jellyfin`.
```
Shadowghost, line 90:
```suggestion
Then simply give the `reverse_proxy` directive a path matcher. The path should be the same as the `Base URL` you entered into Jellyfin's settings. The following example is for a server that is accessible at `example.com/jellyfin`.
```

Shadowghost (approved):

Shadowghost:
Would be nice if you can squash the commits before I merge.

felix920506:
I don't know how I would do such thing

felix920506:
I think you should be able to just squash them while merging

PR #617, opened 2023-08-03
Add Location directive to apache.md to fix issues with WebSocket

Waindor:
I added a Location directive to the vhost configuration to fix issues with WebSocket and reverse proxy. Without this directive, the browser console would say `GET wss://jellyfin.<domain>.<tld>/socket?api_key=<api key> [HTTP/1.1 404 Not Found 4ms]`.

Before adding this directive, marking a TV series episode as watched on the dashboard or anywhere else, the next episode would not show as the next one. You would have to refresh the page to see what the next episode is.

If someone tests this, be aware that it takes 10 seconds before the next episode shows up. If it takes more than that, you have the same problem I had, and should test the code I added.

System:
Debian 12
Jellyfin 10.8.10
Apache 2.4.57

### DIFF
--- a/docs/general/networking/apache.md
+++ b/docs/general/networking/apache.md
@@ -42,6 +42,11 @@ title: Apache
     ProxyPass "/" "http://SERVER_IP_ADDRESS:8096/"
     ProxyPassReverse "/" "http://SERVER_IP_ADDRESS:8096/"
 
+    <Location "/socket">
+        ProxyPass "ws://SERVER_IP_ADDRESS:8096/socket"
+        ProxyPassReverse "ws://SERVER_IP_ADDRESS:8096/socket"
+    </Location>
+
     SSLEngine on
     SSLCertificateFile /etc/letsencrypt/live/DOMAIN_NAME/fullchain.pem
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

thornbill:
How does this differ from the ProxyPass config on lines 39-40?

Shadowghost:
I think the correct way would be to remove L39-L40.

Waindor:
I just tested removing L39-L40, and it still works. I am not sure what the difference betweeen the two are, but I am doing some research on it now.

Waindor:
I have tested and done some more research, and found out that I had reversed the order of L39-L40 and L42-L43, which made Apache send all traffic to "/" instead of hitting "/socket" first. It simply skipped "/socket" since it was in the wrong order. Adding the Location directive like I did made the order not important for some reason. In other words, the documentation is correct, and I will close the pull request. I was not aware that the order was crucial.

All I could find about the difference between using the Location directive and not, is that it can carry a performance penality when not using Location if there are many ProxyPass lines. The advantage of doing as shown in the Jellyfin documentation is that is allows for dynamic control via the Balancer Manager interface. Source: https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#proxypass

PR #652, opened 2023-09-06
Permission error note in docs for docker

mahi160:
Add a permission error note if the config path is in /home/USER_NAME


Shadowghost:
I'm not sure if we should encourage this because there are security concerns about granting read permissions to user homes.

JPVenson:
I agree with ghost here, updating owner for home can lead to huge issues with other software and we generally should not encourage that. But that information would be great to have in the docs alternatively

mahi160:
Yup I go the point! This probably should be path/to/Jellyfin/config instead of user/home right?
Although this happens mostly for the home directory.
I'll make an update there if this seems okay. Thoughts?

JPVenson:
Thats my point, just dont set anything JF related to the home directory. That should be the statement. That you might have to update your permissions elsewhere is another topic.

Sure request a review when you are done.

mahi160 (commented):
Update the folder path and provided a caution.

Shadowghost:
This is not necessarily correct since usually the `/home/<user>` directory has very restrictive permissions, so nobody except the user can access anything within it. Therefore just granting permissions for a user on `/home/<user>/jellyfin_data` doesn't make the folder accessible to the other user.
This is why I said this has security implications, you not only grant another user access to `/home/<user>/jellyfin_data` but also to everything else under `/home/<user>` (at least the other user can list the content)

mahi160:
Okay got it! closing it for now then! thanks for the feedback!

PR #655, opened 2023-09-12
Warning against using Raspberry Pis or other SBCs for hosting JF

felix920506:
While many people like to use these SBCs to host Jellyfin, they are underpowered and don't provide a good experience whenever transcoding is needed as a result.

This PR intends to add a warning message against using these devices for hosting Jellyfin and recommend users to use more powerful hardware.

### DIFF
--- a/docs/general/administration/hardware-acceleration/index.md
+++ b/docs/general/administration/hardware-acceleration/index.md
@@ -27,7 +27,7 @@ The supported and validated video [hardware acceleration (HWA)](https://trac.ffm
 
 :::caution
 
-While hardware acceleration is supported on Raspberry Pi hardware, it is recommended that Jellyfin NOT be hosted on Raspberry Pis or other SBCs. Many hardware acceleration features are not supported and will fallback to software. In addation, they are generally too slow to provide a good experience when transcoding is needed. Please consider getting a more powerful system to host Jellyfin.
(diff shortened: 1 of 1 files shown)


felix920506:
I found a typo myself, will fix later in the day.

thornbill (commented):
Maybe the last sentence could specifically recommend older intel x86 hardware as an affordable alternative?

felix920506:
I think a hardware selection guide deserves its own page rather than being crammed into a RPi warning message. I'll add a short recommendation on hardware as a stopgap over here then change over once a hardware selection guide is complete.

joshuaboniface:
There is a plan (mostly my own) to implement a proper hardware recommendation database and site. For now I'm not sure this is really *needed*, but if it fits I don't really object.

thornbill:
I think this is fine... just letting it sit for a bit after soliciting feedback from the team to make sure no one wants to comment since this _could_ be controversial... 🤷‍♂️

felix920506:
This PR is mostly a reaction to the not insignificant amount of people asking "how to setup hardware acceleration on a RPi" and other questions related to trying to run JF on low power SBCs which are very underpowered. 

Many guides online don't point out the red flags and pit falls for trying to do certain things and running JF on such low power hardware will almost always result in a bad experience.

The people who ask how to transcode on the RPi are likely going to come back for more support once their playback stutters when they encounter a heavy source file that needs transcoding.

thornbill (approved):

PR #666, opened 2023-09-18
Update fail2ban.md

jameskimmel:
This is just a draft to make the doc better. I am not very good at contributing on github, these are just the problem I ran into and I hope to help others.


docs/general/networking/fail2ban.md
```diff
@@ -86,3 +87,28 @@ Assuming you've at least one failed authentication attempt, you can test this ne
 ```bash
 sudo fail2ban-regex /path_to_logs/*.log /etc/fail2ban/filter.d/jellyfin.conf --print-all-matched
 ```
+
+### optional steps for systemd and nftables
```
Shadowghost, line 91:
```suggestion
### Optional steps for systemd and nftables
```

docs/general/networking/fail2ban.md
```diff
@@ -86,3 +87,28 @@ Assuming you've at least one failed authentication attempt, you can test this ne
...
+[sshd]
+enabled = true
+backend=systemd
+
+[DEFAULT]
+banaction = nftables
+banaction_allports = nftables[type=allports]
+```
+
+Without the telling the sshd jail to use systemd as a backend, fail2ban won't even start. The banaction is to send the bans to nftables instead of unused iptables.
```
Shadowghost, line 106:
```suggestion
Without telling the sshd jail to use systemd as a backend, fail2ban won't even start. The `banaction` is set to send the bans to nftables instead of unused iptables.
```

Is this really still necessary on recent Debian/Ubuntu systems?

docs/general/networking/fail2ban.md
```diff
@@ -86,3 +87,28 @@ Assuming you've at least one failed authentication attempt, you can test this ne
...
+And without adding the banaction default, fail2ban will try to use iptables that got replaced by nftables.
```
Shadowghost, line 107:
```suggestion
In a similar manner, without adding the `banaction` default, fail2ban will try to use iptables that got replaced by nftables.
```

docs/general/networking/fail2ban.md
jameskimmel: It definitely does still not work out of the box, but I don't know what the best solution for that would be. See last entry here:
https://github.com/fail2ban/fail2ban/discussions/3575

jameskimmel:
There were some problems I was unable to solve to get fail2ban working on debian and nftables. Since fail2ban does not fully support IPv6 anyway, I will postpone updating the docs.

PR #675, opened 2023-09-25
arch linux package now in extra

feinedsquirrel:
The Arch Linux "Extra" repo now has `jellyfin-server` and `jellyfin-web`.
The [Jellyfin wiki page](https://wiki.archlinux.org/title/Jellyfin) has been updated.

### DIFF
--- a/docs/general/installation/linux.md
+++ b/docs/general/installation/linux.md
@@ -16,9 +16,10 @@ To enable the web UI after installing `jellyfin-web`, make sure to remove the `-
 
 ## Arch Linux
 
-The `Extra` repository contains builds for both [`jellyfin-server`](https://archlinux.org/packages/?name=jellyfin-server) and [`jellyfin-web`](https://archlinux.org/packages/?name=jellyfin-web).
+The `Extra` repository contains builds for both [`jellyfin-server`](https://archlinux.org/packages/?name=jellyfin-server) and [`jellyfin-web`](https://archlinux.org/packages/?name=jellyfin-web). `jellyfin-server` includes a hard dependency on [`jellyfin-ffmpeg`](https://archlinux.org/packages/?name=jellyfin-ffmpeg).
 
-The Jellyfin binary can be installed from the AUR as [`jellyfin-bin`](https://aur.archlinux.org/packages/jellyfin-bin/) and can be built at the tip of the master branch with [`jellyfin-git`](https://aur.archlinux.org/packages/jellyfin-git/).
+Both packages, server and web, can also be built from source at the tip of the master branch using [`jellyfin-git`](https://aur.archlinux.org/packages/jellyfin-git/).
+Debian builds of the Jellyfin server and web binaries can be installed from the AUR at [`jellyfin-bin`](https://aur.archlinux.org/packages/jellyfin-bin/).
 
 ## Fedora
 
--- a/docs/general/installation/linux.md
+++ b/docs/general/installation/linux.md
@@ -16,9 +16,9 @@ To enable the web UI after installing `jellyfin-web`, make sure to remove the `-
 
 ## Arch Linux
 
-Jellyfin can be found in the AUR as [`jellyfin`](https://aur.archlinux.org/packages/jellyfin/), [`jellyfin-bin`](https://aur.archlinux.org/packages/jellyfin-bin/) and [`jellyfin-git`](https://aur.archlinux.org/packages/jellyfin-git/).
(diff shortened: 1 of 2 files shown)


thornbill:
It seems the extras package is **not** including Jellyfin's custom ffmpeg build... which is not great 😕

feinedsquirrel:
I am not the maintainer, nor an Arch Linux trusted user. I noticed that it had moved to the official repos, and thought I'd get the ball rolling.

Something that may be possible is to have the package maintainer add an `*.install` script informing of the existence of the `jellyfin-ffmpegX` packages in the AUR. Looking now, it seems strange that `jellyfin-ffmpeg` is only v4, and not "current", while `jellyfin-ffmpeg6` appears to be the latest version. Arch convention seems to be the `jellyfin-ffmpeg` would always be "latest", and older versions get their version number attached. Discussion with a Trusted User might clarify whether this is a recommended change. @nyanmisaka should probably be involved in that discussion.

Last thought I had was wondering if the jellyfin-ffmpeg alterations get merged upstream. From some reddit comments, and looking myself, jellyfin's version is generally kept close to upstream, and just gets a few patches applied on top. If those patches regularly get merged upstream, then it shouldn't matter that the Arch package links to the system `ffmpeg`, since it stays with upstream stable reliably (in my experience).

At a minimum, I can remove the old content in this documentation, and replace it with a note that Arch users may consider installing `jellyfin-ffmpegX` from the AUR.

Finally, I have not looked at the download wizard, but from the wording of @enter-a-random-username 's comment, it sounds like there is more there than simply altering something related to Arch. Should it be a separate pull-request? (and thinking about it more, it may not even be in this repo, not sure.)

nyanmisaka:
Many changes in jellyfin-ffmpeg are not suitable for upstream, or would take forever, so we decided to maintain a fork to meet our needs. The existence of `jellyfin-ffmpeg{5,6}` was due to changes in the upstream ffmpeg API that broke compatibility with jellyfin-server. At that time, I never considered that jellyfin-server AUR would be adopted by extra.

Maybe I should get in touch with their maintainer to add jellyfin-ffmpeg into the jellyfin-server PKGBUILD, like what emby-server PKGBUILD does. But I didn't find the an email or other contact information.

https://gitlab.archlinux.org/archlinux/packaging/packages/jellyfin-server/-/blob/main/PKGBUILD
https://gitlab.archlinux.org/archlinux/packaging/packages/emby-server/-/blob/main/PKGBUILD

nyanmisaka:
https://github.com/grawlinson

Well I found it, thanks to him using the same username on github.
I'll try to get in touch with him and find a good place for jellyfin-ffmpeg.

feinedsquirrel:
Not that you need my approval, but putting `jellyfin-ffmpeg` into the `*-server` pkgbuild seems like a good idea. And thank you for clarifying the differences of that fork. The reddit comments I had found were not super detailed.

I held off for now making a recommendation to install jellyfin-ffmpeg from the aur alongside the Arch "Extra" version until we hear the outcome of the discussion between @nyanmisaka and @grawlinson. If the decision is made to keep them separate, then we can add a phrase in this documentation, and to the Jellyfin page on the Arch Wiki that a `*-server` installation may be well served by also installing the package from the AUR.

grawlinson:
I see someone has pinged me! Can we move this to a separate issue (as it's not relevant to this PR) and discuss it there?

feinedsquirrel:
@nyanmisaka might have input as well on where a separate issue should be opened; he's also messaged you directly, I believe. I wonder if an arch flyspray issue would be best, since it primarily involves how it is packaged in the arch repo.

nyanmisaka:
> I see someone has pinged me! Can we move this to a separate issue (as it's not relevant to this PR) and discuss it there?

Let's move here. https://github.com/jellyfin/jellyfin/issues/10315

feinedsquirrel:
@enter-a-random-username I've taken a stab at the download wizard.

I looked through the other options in the file as a guide.
I wonder if instead of linking to the package in Arch-Extra, it should  link to the Arch Wiki page.
Anyone have thoughts on this?

Shadowghost (approved):

feinedsquirrel:
`jellyfin-bin` has now been removed from the AUR, so I removed it from the doc. I also explicitly stated `-server-git` and `-web-git` are also available.

I'm not sure if anything needs to be mentioned about older versions of `jellyfin-ffmpeg` in the AUR. Let me know if I should add anything. Otherwise, I think this is about ready, as the linked issues have been resolved.

PR #694, opened 2023-10-08
Add script-src unsafe-eval to CSP for latest master branch changes

### LINKED ISSUE #693: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' 'unsafe-inline'
To get various required fixes, I'm trying to switch from branch `release-10.8.z` to `master`. I've re-built `jellyfin` and `jellyfin-web` successfully but launching the web browser in either Chrome or Firefox shows a CSP error. For example, from Chrome:

```
Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "default-src https: data: blob: http://image.tmdb.org; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline' 'unsafe-eval' https://www.gstatic.com https://www.youtube.com blob:; worker-src 'self' blob:; connect-src 'self'; object-src 'none'; frame-ancestors 'self'".

    at ./index.jsx (main.jellyfin.bundle.js?8b2376313fc177f319ea:348:1)
    at __webpack_require__ (runtime.bundle.js?8b2376313fc177f319ea:31:42)
    at __webpack_exec__ (main.jellyfin.bundle.js?8b2376313fc177f319ea:8551:61)
    at main.jellyfin.bundle.js?8b2376313fc177f319ea:8552:1378
    at __webpack_require__.O (runtime.bundle.js?8b2376313fc177f319ea:70:23)
    at main.jellyfin.bundle.js?8b2376313fc177f319ea:8553:56
    at webpackJsonpCallback (runtime.bundle.js?8b2376313fc177f319ea:316:39)
    at main.jellyfin.bundle.js?8b2376313fc177f319ea:9:53
first.js:1 App Version: 3.9.0
```

kevgrig:
Fixes #693

### DIFF
--- a/docs/general/networking/nginx.md
+++ b/docs/general/networking/nginx.md
@@ -80,7 +80,7 @@ server {
     # Enforces https content and restricts JS/CSS to origin
     # External Javascript (such as cast_sender.js for Chromecast) must be whitelisted.
     # NOTE: The default CSP headers may cause issues with the webOS app
-    #add_header Content-Security-Policy "default-src https: data: blob: http://image.tmdb.org; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline' https://www.gstatic.com https://www.youtube.com blob:; worker-src 'self' blob:; connect-src 'self'; object-src 'none'; frame-ancestors 'self'";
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

nielsvanvelzen:
I think we should fix jellyfin-web to NOT use eval instead. There is a reason the CSP directive starts with unsafe.

Shadowghost:
Agreed. I thought it was intentional on the jf-web side.

Shadowghost:
Shoud be fixed on web.

PR #716, opened 2023-10-21
Major cleanup of Linux install instructions

joshuaboniface:
This is an attempt to clean up and streamline the various Linux package install instructions.

1. (For Debuntu) The main Downloads page now provides just the most basic instructions for using the script install; users are directed to the docs page for more advanced instructions. We don't need walls of text or commands in the quickstart.

2. The Debian/Ubuntu docs are cleaned up and unified: a) The Debian and Ubuntu sections were mostly identical, so combine them and show only the (single) relevant difference. b) Add a clear command for integrity checking the script. c) Expand some aspects of the manual process for better checking, and link to the script rather than provide a list of codenames that must be constantly updated. d) Cleans up some wording.

3. The Fedora and CentOS docs are cleaned up: a) Use consistent monospace for package names and capitalization elsewhere. b) Clean up the steps a bit. c) Clarify the source and when a repo will come. d) For CentOS mention that the process is "similar" to Fedora (better than the nothing that was there before).

4. Use consistent links to the download repository throughout; linking back to the downloads page, which presumably someone *came from*, is unfriendly. Just link to the right place.

### DIFF
--- a/docs/general/installation/linux.md
+++ b/docs/general/installation/linux.md
@@ -175,7 +175,7 @@ If you would prefer to install everything manually, the full steps are as follow
    :::note
 
    The supported values for the above variables are:
-   
+
    * `${VERSION_OS}`: One of `debian` or `ubuntu`; if it is not, use the closest one for your distribution.
    * `${VERSION_CODENAME}: One of our supported [Debian](https://github.com/jellyfin/jellyfin-metapackages/blob/master/install-debuntu.sh#L7) or [Ubuntu](https://github.com/jellyfin/jellyfin-metapackages/blob/master/install-debuntu.sh#L8) release codenames. These can change as new releases come out and old releases are dropped, so check the script to be sure yours is supported.
    * `${DPKG_ARCHITECTURE}`: One of our [supported architectures](https://github.com/jellyfin/jellyfin-metapackages/blob/master/install-debuntu.sh#L6). Microsoft does not provide a .NET for 32-bit x86 Linux systems, and hence Jellyfin is **not** supported on the `i386` architecture.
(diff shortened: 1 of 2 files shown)


src/data/downloads.tsx
```diff
@@ -57,36 +57,8 @@ export const Downloads: Array<Download> = [
...
-sudo apt update
-sudo apt install jellyfin`}
-              </code>
-            </pre>
-            <p>
-              <b>Note:</b>: If you are running a non-Debian, non-Ubuntu derivative, ensure the <code>$DISTRO</code> and{' '}
-              <code>$CODENAME</code> are valid Debian or Ubuntu values!
-            </p>
+            <p>If you do not have <code>curl</code> installed, you can use <code>wget -O-</code> instead of <code>curl</code>.</p>
+            <p>For more advanced users, the full steps can be [found in the docs](https://jellyfin.org/docs/general/installation/linux#debuntu).</p>
```
joshuaboniface, line 61: I'm iffy on this link. I presume it would be correct but I'm not exactly sure how well Docusaurus handles this.

Shadowghost: You can do relative redirects too afaik

joshuaboniface: Oh it's more the # part, given that the actual section has a longer title :D

docs/general/installation/linux.md
```diff
@@ -264,7 +147,9 @@ If you don't trust the script or want to do everything manually, the full steps
 
    :::
 
-3. Download the GPG signing key (signed by the Jellyfin Team):
+   On Debian, you can also enable the `non-free` components of your base repositories for additional FFMpeg dependencies, but this is optional.
```
thornbill, line 150:
```suggestion
   On Debian, you can also enable the `non-free` components of your base repositories for additional FFmpeg dependencies, but this is optional.
```

thornbill (approved):

PR #725, opened 2023-10-28
Fix spacing and link in downloads

joshuaboniface:
Corrects some faults with #716 

1. Missing spaces between the words "use" and "wget" and between "of" and "curl". Prettier insists I remove this extra space but that is nonsensical.
2. Link should be an `<a>` tag as markdown-style did not work.


joshuaboniface:
I don't care what `prettier` says, it's suggestions make no sense.

thornbill (approved):

PR #759, opened 2023-12-02
Update nginx.md

Ladrien-git:
Updating the NGINX config to disable insecure TLS 1.0 and 1.1 version

### DIFF
--- a/docs/general/networking/nginx.md
+++ b/docs/general/networking/nginx.md
@@ -42,6 +42,9 @@ server {
     ## The default `client_max_body_size` is 1M, this might not be enough for some posters, etc.
     client_max_body_size 20M;
 
+    #Disable TLS 1.0 and 1.1
+    ssl_protocols TLSv1.3 TLSv1.2;
+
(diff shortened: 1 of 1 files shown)


felix920506 (changes requested):
There are many older devices out there that might not support TLS 1.2+ and this setting might break those clients. There should at least be a warning

Maybe change it to
```
    # Uncomment next line to Disable TLS 1.0 and 1.1 (Might break older devices)
    # ssl_protocols TLSv1.3 TLSv1.2;
```

felix920506 (approved):
Looks good to me

nielsvanvelzen (approved):

DX37:
There's no point in this change. `/etc/letsencrypt/options-ssl-nginx.conf` already contain this line. After uncommeting that line in nginx config it would print warning about duplicating.
image

DX37:
That's applies to Let's Encrypt usage, so maybe at least leave a warning about it?

PR #791, opened 2023-12-16
Update docker volume passthrough info

emanueljg:
I don't actually know what update changed this (might be nvidia drivers update, or whatever) as I updated my distro from one bi-yearly stable version to another (NixOS) and a lot has changed since then, but I recently have to add `nvidia-caps` to the volumes for it to work again. Would be great for someone with better knowledge of how nvidia/docker/jellyfin works to repro what update specifically adds the `/dev/nvidia-caps` volume requirement.


Shadowghost:
Line 369 already lists that mount.

emanueljg:
Ah, my bad. Thank you!

PR #806, opened 2023-12-25
use AutoUpdate option instead of label in podman systemd example

luebke-dev:
use AutoUpdate option instead of label in podman systemd example


nielsvanvelzen (approved):
Thanks!

PR #811, opened 2023-12-30
add demo user & password under the "See it in action" button

ColorCookie-dev:
the user and password wasn't mentioned near the demo button causing confusion.

In the index page, the demo button is present, however when clicking it, neither it show the user and password to the user, nor the demo server shows the user and password. For first time user, it will be very confusing. I also had to guess the credentials.

### DIFF
--- a/src/components/common/Hero.scss
+++ b/src/components/common/Hero.scss
@@ -67,6 +67,13 @@
   margin-bottom: 0;
 }
 
+#demo_userpass_span {
+  opacity: 0.75;
+  line-height: 1.7;
+  font-size: 0.9em;
+  font-style: italic;
+}
+
 .hero__buttons {
   display: flex;
   justify-content: center;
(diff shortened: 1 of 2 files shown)


Shadowghost:
The login credentials are on the demo server below the login field. If you are using script blockers you do not see them.

ColorCookie-dev:
Oh, I understand. I apologize for not testing on other browsers. Do you know why this is happening? I used Mullvad Browser, and on the default security level (not the safest). Why would the text be blocked by script blockers?

nielsvanvelzen:
A lot of "privacy" browsers/extensions just block random elements on a page that they think could be advertisements/cookiebanners/trackers etc. this can often lead to false-positives.

Closing this pull request as the instructions are on the demo server and we have no interest in adding it to the jellyfin.org homepage at this time.

PR #820, opened 2024-01-08
clarified series file name

JPVenson:


Shadowghost (approved):

docs/general/server/media/shows.md
```diff
@@ -15,25 +15,25 @@ In order to help with identifying a series, Jellyfin can make use of media provi
...
 ```txt
 Shows
-├── Series (2010)
+├── Series Name A (2010)
 │   ├── Season 00
 │   │   ├── Some Special.mkv
-│   │   ├── Episode S00E01.mkv
-│   │   └── Episode S00E02.mkv
+│   │   ├── Series Name A S00E01.mkv
+│   │   └── Series Name A S00E02.mkv
```
nielsvanvelzen, line 22: Previous example indicated episodes can have names, which is removed in this example.

JPVenson: How did the previous example indicate that? it just called for `Episode`.

nielsvanvelzen: Yes, which I assume was a placeholder for the episode name

JPVenson: As far as i know the episodename was not supposed to be there but appended like
`Stargate SG1 S01E01 Children of the gods.mkv`
or
``Stargate SG1 S01E01 - Children of the gods.mkv``

JPVenson: https://github.com/jellyfin/jellyfin/blob/master/Emby.Naming/Common/NamingOptions.cs#L344

this is the default scheme and if i read the expressions correctly, where a seriesname is defined it always starts the filename. With all the misinterpretations of the scheme i would say that its best we show the case that is known to work best/most reliable and from all examples i saw is that this scheme `SeriesName (YEAR) Title.container` works in 99% of all cases

nielsvanvelzen: Fair enough

PR #841, opened 2024-02-04
Add missing column for linter

1337joe:
We've been missing the column closing pipe for two years but the linter is just now complaining.


thornbill (approved):
Probably caused by a markdownlint update at some point.

nielsvanvelzen:
We should add markdown-cli as dev delendency so it is pinned. Right now the lint workflow always downloads the latest version.

PR #842, opened 2024-02-04
Added new Client to "Clients" page

appsynergydev:
Hi, I have a Jellyfin client I would like to add to the clients page. 

Thank you.

Screenshot 2024-02-04 at 19 14 45


src/data/clients.ts
```diff
@@ -637,6 +637,30 @@ export const Clients: Array<Client> = [
     ],
     recommended: true
   },
+  {
+    id: 'phyn',
+    name: 'Phyn',
+    description: 'Multiplatform video and audio playback client for Jellyfin with Jellyseerr support. Available on iOS, iPadOS and macOS. Android coming soon...',
```
nielsvanvelzen, line 643: - We do not advertise piracy tools (Jellyseerr) from out documentation
- The "Android coming soon" part is not worth mentioning here

src/data/clients.ts
```diff
@@ -637,6 +637,30 @@ export const Clients: Array<Client> = [
     ],
     recommended: true
   },
+  {
```
nielsvanvelzen, line 640: New entries should be added at the end.

nielsvanvelzen (changes requested):

src/data/clients.ts
appsynergydev: Thank for the feedback. I have moved it to the end.

src/data/clients.ts
appsynergydev: thanks for the feedback. I have fixed it

nielsvanvelzen:
Thanks, the changes look fine now. We're discussing the inclusion of this client internally which will take a bit longer.

appsynergydev:
No worries. Thank you

nielsvanvelzen:
After some discussion with the other team members (including the core team ) we decided to not include this client. Jellyfin is a FOSS product and adding a proprietary closed-source client with a subscription model does not fit our philosophy.

Some other concerns that were brought up are:
- It's unknown who the actual author is, it feels like they are hiding their identity
- Author made a second GitHub account for this specific pull request which is shady
- The app promotes piracy

I will close the pull request.

appsynergydev:
Thank you for the feedback.

I understand your concerns.

From what I gather, the defining reason for this rejection was because of the subscription model. However, the core functionality of the app is totally free. The subscription is just for add-ons which are totally optional. You also do have a subscription based client on your client list so I don't really understand the rationale.

As for your other concerns, those can be easily rectified. The new account for one was because the AppsynergyTS (the company) wanted everything done in their name for what I assume is IP protection reasons but they didn't have a Github account. 

I guess my question is, do I resubmit another request after addressing your concerns?

nielsvanvelzen:
The main reason is that the app is closed source and encourages piracy due to inclusion of the jellyseerr support.

> The new account for one was because the AppsynergyTS (the company) wanted everything done in their name for what I assume is IP protection reasons but they didn't have a Github account.

Can you explain this in more detail? Are you being paid to develop this app for a company?

appsynergydev:
> The main reason is that the app is closed source and encourages piracy due to inclusion of the jellyseerr support.

I undersand. 

> Can you explain this in more detail? Are you being paid to develop this app for a company?

No, not paid. I was only asked to develop the app landing page and then make the changes in this PR.  However, I will ask the other dev to reach out instead if that will clarify the situation.

Thank you for your time.

PR #856, opened 2024-02-12
Inform users of known issues in the latest Windows graphics drivers

nyanmisaka:
- Inform users of known issues in the latest Windows graphics drivers

They have been asked many times. And it seems like Intel and AMD will need several iterations of driver versions to fix them.


nielsvanvelzen (commented):
These pages have excessive use of admonitions. I don't think adding more rainbow puke will help.

nyanmisaka:
> These pages have excessive use of admonitions. I don't think adding more rainbow puke will help.

Indeed. Very few users have the patience to read through this.
But at least when a user asks next time we can directly show him the link😅

PR #858, opened 2024-02-18
nginx: Remove Origin-Agent-Cluster header

FintasticMan:
With this header set like this, the webOS client may show a white screen when trying to play media. See jellyfin/jellyfin-webos#205.


nielsvanvelzen:
I think it is fine to remove this header from our documentation. It's an experimental feature only implemented by chrome based browsers and there is a chance it will be dropped/changed over time because it is considered experimental.

FintasticMan:
Yes, seeing as it's experimental and is shown to cause issues, I agree that it doesn't make sense to encourage setting it by default. I'll update this PR to just remove it from the guide.

PR #865, opened 2024-02-21
Correcting and updating instructions for nginx proxy manager

solidsnake1298:
Correcting and updating instructions for nginx proxy manager based on user feedback.  Verified in test setup.

### DIFF
--- a/docs/general/networking/nginx.md
+++ b/docs/general/networking/nginx.md
@@ -393,13 +393,13 @@ error_page 460 http://your-page-telling-your-limit/;
 
 Create a proxy host and point it to your Jellyfin server's IP address and http port (usually 8096)
 
-Enable "Cache Assets", "Block Common Exploits", and "Websockets Support". Configure the access list if you intend to use them. Otherwise leave it on "publicly accessible".
+Enable "Block Common Exploits", and "Websockets Support". Configure the access list if you intend to use them. Otherwise leave it on "publicly accessible".
 
-In the "Custom Locations" tab, create two HTTP locations using the same IP and port as the step before:
+In the "Advanced" tab, enter the following in "Custom Nginx Configuration".  This is optional, but recommended if you intend to make Jellyfin accessible outside of your home.
 
 ```config
-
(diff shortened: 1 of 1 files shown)


docs/general/networking/nginx.md
```diff
@@ -423,39 +423,4 @@ LOCATION: /
...
-    add_header X-Content-Type-Options "nosniff";
-
-    # Content Security Policy
-    # See: https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
-    # Enforces https content and restricts JS/CSS to origin
-    # External Javascript (such as cast_sender.js for Chromecast) must be whitelisted.
-    # NOTE: The default CSP headers may cause issues with the webOS app
-    #add_header Content-Security-Policy "default-src https: data: blob: http://image.tmdb.org; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline' https://www.gstatic.com/cv/js/sender/v1/cast_sender.js https://www.gstatic.com/eureka/clank/95/cast_sender.js https://www.gstatic.com/eureka/clank/96/cast_sender.js https://www.gstatic.com/eureka/clank/97/cast_sender.js https://www.youtube.com blob:; worker-src 'self' blob:; connect-src 'self'; object-src 'none'; frame-ancestors 'self'";
-
-```
```
nielsvanvelzen, line 449: Is this not needed anymore? Normally without additional configurations Nginx does not work with WebSockets.

nielsvanvelzen: > The playback tracker in the dashboard updated as expected.

That does not use websockets though. The best way to test sockets is probably with the remote control feature, so try to play/pause/seek a different device. And make sure that it still works after 5 minutes to verify there is no timeout in Nginx.

solidsnake1298: Confirmed that enabling "Websocket Support" on the details page is sufficient to open and maintain websockets.  I was able to remote control another session as expected.

nielsvanvelzen (approved):

PR #875, opened 2024-02-29
feat: add hwa doc for apple platforms

gnattu:
Update the docs to include hardware acceleration docs for Apple platforms in the upcoming 10.9 release.

### DIFF
--- a/docs/general/administration/hardware-acceleration/apple.md
+++ b/docs/general/administration/hardware-acceleration/apple.md
@@ -11,11 +11,11 @@ This tutorial guides you on setting up full video hardware acceleration on Apple
 
 [VideoToolbox](https://developer.apple.com/documentation/videotoolbox) is the only available method on macOS.
 
-To achieve full acceleration, [Metal](https://developer.apple.com/metal/) and OpenCL is also required.
+To achieve full acceleration, [Metal](https://developer.apple.com/metal/) and OpenCL are also required.
 
 ## Tone-mapping Methods
 
-Hardware accelerated HDR/DV to SDR tone-mapping is supported on most Macs from 2017 and later, with the exception of the MacBook Air (13-inch, 2017).
+Hardware accelerated HDR to SDR tone-mapping is supported on all Macs from 2017 and later, with the exception of the MacBook Air (13-inch, 2017).
 
 There are two different methods that can be used. Pros and cons are listed below:
 
@@ -71,9 +71,7 @@ There is no hardware accelerated path for AV1 on macOS at the moment. Although t
 
 ## macOS Setups
 
-Although there is no hard requirement for the macOS version, we recommend **macOS 12 and later**.
-
-Jellyfin's custom ffmpeg may work on earlier versions, but the functionality cannot be guaranteed.
+**MacOS 12 and later** are officially supported. Older versions might work, but are not supported.
 
 ### Configure
 
--- a/docs/general/administration/hardware-acceleration/apple.md
+++ b/docs/general/administration/hardware-acceleration/apple.md
@@ -73,7 +73,7 @@ On legacy Intel Macs, you may encounter performance issues with tone-mapping usi
 
 <sup>1</sup> Using prefer speed encoder preset.
 
-<sup>2</sup> The simultaneous session count is a soft limit. You can run more sessions if you are willing to accept the per-session performance trade-offs.
+<sup>2</sup> The simultaneous session count is a soft limit. You can run more sessions if you want, but the transcoding performance of each session may be reduced to a point where video playback starts to stutter.
 
 ## macOS Setups
 
--- a/docs/general/administration/hardware-acceleration/apple.md
+++ b/docs/general/administration/hardware-acceleration/apple.md
@@ -43,7 +43,7 @@ To have native Apple Silicon support, Jellyfin server 10.9.0+ and `jellyfin-ffmp
 
 While hardware acceleration via VideoToolbox might work on older series Macs, it is not officially supported.
 
-<sup>*</sup> On very old Macs, VideoToolbox may fallback to software decoding/encoding. Macs from 2017 and later is strongly recommend.
+<sup>*</sup> VideoToolbox may fall back to software decoding/encoding when the task exceeds its hardware capability. The older your Mac, the more this will occur. Macs from 2017 and later are strongly recommended.
 
 ### Transcode H.264
 
--- a/docs/general/administration/hardware-acceleration/apple.md
+++ b/docs/general/administration/hardware-acceleration/apple.md
@@ -43,7 +43,7 @@ To have native Apple Silicon support, Jellyfin server 10.9.0+ and `jellyfin-ffmp
 
 While hardware acceleration via VideoToolbox might work on older series Macs, it is not officially supported.
 
-If your Mac does not have an internal display, you may need to connect it to a monitor or use an HDMI dummy plug to prevent the GPU from being throttled.
+If your Mac does not have an internal display, you may need to connect it to a monitor or use a dummy plug to prevent the GPU from being throttled.
 
 <sup>*</sup> VideoToolbox may fallback to software decoding/encoding or stop working when the task exceeds its hardware capability. The older your Mac, the more this will occur. Macs from 2017 and later, especially Apple Silicon Macs are strongly recommended.
 
--- a/docs/general/administration/hardware-acceleration/apple.md
+++ b/docs/general/administration/hardware-acceleration/apple.md
@@ -43,6 +43,8 @@ To have native Apple Silicon support, Jellyfin server 10.9.0+ and `jellyfin-ffmp
 
 While hardware acceleration via VideoToolbox might work on older series Macs, it is not officially supported.
 
+If your Mac does not have an internal display, you may need to connect it to a monitor or use an HDMI dummy plug to prevent the GPU from being throttled.
+
 <sup>*</sup> VideoToolbox may fallback to software decoding/encoding or stop working when the task exceeds its hardware capability. The older your Mac, the more this will occur. Macs from 2017 and later, especially Apple Silicon Macs are strongly recommended.
 
 ### Transcode H.264
--- a/docs/general/administration/hardware-acceleration/apple.md
+++ b/docs/general/administration/hardware-acceleration/apple.md
@@ -63,15 +63,17 @@ There is no hardware accelerated path for AV1 on macOS at the moment. Although t
 
 ### Performance Consideration
 
-An Apple Silicon-based Mac is preferred in most cases. Even an entry-level M1 can handle three 4K 24fps Dolby Vision HEVC 10-bit transcoding tasks simultaneously while performing tone-mapping to SDR.<sup>*</sup>
+An Apple Silicon-based Mac is preferred in most cases. Even an entry-level M1 can handle three 4K 24fps Dolby Vision HEVC 10-bit transcoding tasks simultaneously while performing tone-mapping to SDR.<sup>1</sup> <sup>2</sup>
 
-The "Max" variant chips come with an additional video encoding engine. VideoToolbox can utilize this extra engine even when there is only a single transcoding session, enabling support for 4K 120fps transcoding and tone-mapping.<sup>*</sup>
+The "Max" variant chips come with an additional video encoding engine. VideoToolbox can utilize this extra engine even when there is only a single transcoding session, enabling support for 4K 120fps transcoding and tone-mapping.<sup>1</sup>
 
 The "Ultra" variant chips feature 2 video decoding engines and 4 video encoding engines, effectively doubling the capability compared to the "Max" variant chips.
 
 On legacy Intel Macs, you may encounter performance issues with tone-mapping using Metal on 4K videos if your Mac doesn't have an AMD GPU. However, it is still adequate for transcoding SDR videos.
 
-<sup>*</sup> Using prefer speed encoder preset.
+<sup>1</sup> Using prefer speed encoder preset.
+
+<sup>2</sup> The simultaneous session count is a soft limit. You can run more sessions if you are willing to accept the per-session performance trade-offs.
 
 ## macOS Setups
 
--- a/docs/general/administration/hardware-acceleration/apple.md
+++ b/docs/general/administration/hardware-acceleration/apple.md
@@ -27,9 +27,9 @@ There are two different methods that can be used. Pros and cons are listed below
 
 2. **VideoToolbox Native**
 
-   - Pros - Lower power consumption, more performant, good visual quality without manual fine-tuning.
+   - Pros: Lower power consumption, less dependency on GPU performance, good visual quality without manual fine-tuning.
 
-   - Cons - No tuning options, might work badly on Intel Macs.
+   - Cons: Lack of tuning options, potential performance issues on Intel Macs.
 
 :::note
 
--- a/docs/general/administration/hardware-acceleration/apple.md
+++ b/docs/general/administration/hardware-acceleration/apple.md
@@ -43,7 +43,7 @@ To have native Apple Silicon support, Jellyfin server 10.9.0+ and `jellyfin-ffmp
 
 While hardware acceleration via VideoToolbox might work on older series Macs, it is not officially supported.
 
-<sup>*</sup> VideoToolbox may fallback to software decoding/encoding when the task exceeds its hardware capability. The older your Mac, the more this will occur. Macs from 2017 and later are strongly recommended.
+<sup>*</sup> VideoToolbox may fallback to software decoding/encoding or stop working when the task exceeds its hardware capability. The older your Mac, the more this will occur. Macs from 2017 and later are strongly recommended.
 
 ### Transcode H.264
 
--- a/docs/general/administration/hardware-acceleration/apple.md
+++ b/docs/general/administration/hardware-acceleration/apple.md
@@ -23,7 +23,7 @@ There are two different methods that can be used. Pros and cons are listed below
 
    - Pros: Supports Dolby Vision P5, detailed fine-tuning options.
 
-   - Cons: Slower on low end GPU, especially on Intel iGPUs.
+   - Cons: Slower on low end GPUs, especially on Intel iGPUs.
 
 2. **VideoToolbox Native**
 
--- a/docs/general/administration/hardware-acceleration/apple.md
+++ b/docs/general/administration/hardware-acceleration/apple.md
@@ -43,7 +43,7 @@ To have native Apple Silicon support, Jellyfin server 10.9.0+ and `jellyfin-ffmp
 
 While hardware acceleration via VideoToolbox might work on older series Macs, it is not officially supported.
 
-<sup>*</sup> VideoToolbox may fallback to software decoding/encoding or stop working when the task exceeds its hardware capability. The older your Mac, the more this will occur. Macs from 2017 and later are strongly recommended.
+<sup>*</sup> VideoToolbox may fallback to software decoding/encoding or stop working when the task exceeds its hardware capability. The older your Mac, the more this will occur. Macs from 2017 and later, especially Apple Silicon Macs are strongly recommended.
 
 ### Transcode H.264
 
--- a/docs/general/administration/hardware-acceleration/apple.md
+++ b/docs/general/administration/hardware-acceleration/apple.md
@@ -81,8 +81,7 @@ On legacy Intel Macs, you may encounter performance issues with tone-mapping usi
 
 ### Configure
 
-- Enable VideoToolbox in Jellyfin and uncheck unsupported codecs.
-  - Disable `MPEG4`, `MPEG2`, and `VC-1`, as VideoToolbox lacks a hardware-accelerated decoding path for these codecs.
+- Enable VideoToolbox in the Jellyfin Dashboard under the Playback section and deselect unsupported codecs for your Mac.
 - Check `Enable VideoToolbox Tone mapping` if you want to use VideoToolbox native tone-mapping.
 - Check `Enable Tone mapping` if you want to use Metal-based tone-mapping.
 - Optionally, select an `Encoding Preset`. The `veryslow`, `slower`, `slow`, and `medium` presets prioritize quality, while `fast`, `faster`, `veryfast`, `superfast`, and `ultrafast` prioritize speed. The default `Auto` setting prioritizes speed.
--- a/docs/general/administration/hardware-acceleration/apple.md
+++ b/docs/general/administration/hardware-acceleration/apple.md
@@ -43,7 +43,7 @@ To have native Apple Silicon support, Jellyfin server 10.9.0+ and `jellyfin-ffmp
 
 While hardware acceleration via VideoToolbox might work on older series Macs, it is not officially supported.
 
-<sup>*</sup> VideoToolbox may fall back to software decoding/encoding when the task exceeds its hardware capability. The older your Mac, the more this will occur. Macs from 2017 and later are strongly recommended.
+<sup>*</sup> VideoToolbox may fallback to software decoding/encoding when the task exceeds its hardware capability. The older your Mac, the more this will occur. Macs from 2017 and later are strongly recommended.
 
 ### Transcode H.264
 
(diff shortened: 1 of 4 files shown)


docs/general/administration/hardware-acceleration/apple.md
```diff
@@ -0,0 +1,94 @@
...
+
+# HWA Tutorial On Mac
+
+This tutorial guides you on setting up full video hardware acceleration on Apple Macs via VideoToolbox.
+
+## Acceleration Methods
+
+[VideoToolbox](https://developer.apple.com/documentation/videotoolbox) is the only available method on macOS.
+
+To achieve full acceleration, [Metal](https://developer.apple.com/metal/) and OpenCL is also required.
```
felix920506, line 14:
```suggestion
To achieve full acceleration, [Metal](https://developer.apple.com/metal/) and OpenCL are also required.
```

docs/general/administration/hardware-acceleration/apple.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Tone-mapping Methods
+
+Hardware accelerated HDR/DV to SDR tone-mapping is supported on most Macs from 2017 and later, with the exception of the MacBook Air (13-inch, 2017).
```
felix920506, line 18:
```suggestion
Hardware accelerated HDR to SDR tone-mapping is supported on all Macs from 2017 and later, with the exception of the MacBook Air (13-inch, 2017).
```

docs/general/administration/hardware-acceleration/apple.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Select System Hardware
+
+Hardware accelerated transcoding is supported on all Macs that support [VideoToolbox](https://developer.apple.com/documentation/videotoolbox). This includes most 2011 and later Macs.<sup>*</sup>
+
+Full acceleration is available on most Macs from 2017 and later, with the exception of the MacBook Air (13-inch, 2017).
+
+To have native Apple Silicon support, Jellyfin server 10.9.0+ and the latest `jellyfin-ffmpeg6` is required.
+
+<sup>*</sup> On very old Macs, VideoToolbox may fallback to software decoding/encoding. Macs from 2017 and later is strongly recommend.
```
felix920506, line 48:
```suggestion
Hardware accelerated transcoding is supported on all Macs that support [VideoToolbox](https://developer.apple.com/documentation/videotoolbox). This includes most 2011 and later Macs.<sup>*</sup>

Full acceleration is available on most Macs from 2017 and later, with the exception of the MacBook Air (13-inch, 2017).

To have native Apple Silicon support, Jellyfin server 10.9 or newer and `jellyfin-ffmpeg6` is required.

<sup>*</sup> On very old Macs, VideoToolbox may fallback to software decoding/encoding. Macs from 2017 and later is strongly recommend.
```

The exact minimum version of jellyfin-ffmpeg6 should be included.
Also, you mentioned that VideoToolbox may fallback to software for older macs. How old exactly are we talking about here?

docs/general/administration/hardware-acceleration/apple.md
```diff
@@ -0,0 +1,94 @@
...
+
+:::note
+
+While hardware acceleration via VideoToolbox might work on older series Macs, it is not officially supported.
+
+:::
```
felix920506, line 54: This really doesn't have to be an admonition and can be part of the body with some rewording

docs/general/administration/hardware-acceleration/apple.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Transcode H.264
+
+AVC / H.264 8-bit is still widely used due to its excellent compatibility. All Intel GPUs that support QSV can decode and encode it.
```
felix920506, line 58: There exist macs out there that don't have Intel Integrated graphics (eg. Mac Pro, iMac Pro). What about these macs?

docs/general/administration/hardware-acceleration/apple.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Transcode AV1
+
+There is no hardware accelerated path for AV1 on macOS at the moment. Although the M3 series added AV1 decoding support, [ffmpeg does not support it yet](https://trac.ffmpeg.org/ticket/10642).
+
+## macOS Setups
+
+Although there is no hard requirement for the macOS version, we recommend **macOS 12 and later**.
+
+Jellyfin's custom ffmpeg may work on earlier versions, but the functionality cannot be guaranteed.
```
felix920506, line 76:
```suggestion
**MacOS 12 and later** are officially supported. Older versions might work, but are not supported.
```

I would disagree with not having a hard minimum OS version requirement that we support. Or else we might get some guy with ancient hardware and ancient OS versions coming to matrix or the forum for support.

docs/general/administration/hardware-selection.md
```diff
@@ -43,16 +43,12 @@ Intel ARC GPUs are recommended when upgrading an existing system to be used as a
...
-
-It is recommended that Intel-based macs be used with Windows or Linux installed to host Jellyfin. Many hardware acceleration features aren't available on MacOS due to the custom [jellyfin-ffmpeg](https://github.com/jellyfin/jellyfin-ffmpeg) fork not being available.
-
-:::
-
 ### Low Power Applications
 
 For low power applications, Intel 12th gen or newer Atom CPUs with integrated graphics are recommended. It is also recommended that [Low Power Encoding](/docs/general/administration/hardware-acceleration/intel/#low-power-encoding) be setup.
 
+Alternatively, you can use an Apple Silicon Mac for even lower power consumption.
```
felix920506, line 50: Doubt. Does an Apple Silicon Mac Mini consume measurably less power than an N100 based mini PC?

docs/general/administration/hardware-acceleration/apple.md
```diff
@@ -0,0 +1,94 @@
...
+
+   - Pros: Lower power consumption, less dependency on GPU performance, good visual quality without manual fine-tuning.
+
+   - Cons: Lack of tuning options, potential performance issues on Intel Macs.
+
+:::note
+
+When both methods are enabled, VideoToolbox Native will be used for most videos, and OpenCL will only be used as a fallback for Dolby Vision Profile 5 videos.
+
+:::
```
felix920506, line 38: This doesn't have to be an admonition.

docs/general/administration/hardware-acceleration/apple.md
```diff
@@ -0,0 +1,94 @@
...
+
+   - Pros - Supports Dolby Vision P5, detailed fine-tuning options.
+
+   - Cons - Slower, requires more powerful GPUs.
+
+2. **VideoToolbox Native**
+
+   - Pros: Lower power consumption, less dependency on GPU performance, good visual quality without manual fine-tuning.
+
+   - Cons: Lack of tuning options, potential performance issues on Intel Macs.
```
felix920506, line 32: Is there any reason for using dashes for OpenCL but colons for VideoToolbox Native? They should be the same and follow the style of other documentation pages.
Cons for VideoToolbox Native should also include lack of support for DoVi P5.

felix920506 (changes requested):

docs/general/administration/hardware-selection.md
```diff
@@ -17,7 +17,7 @@ For a Jellyfin server, the following is recommended:
 - CPU (Without dGPU): Intel Pentium G4560, Intel Core i3-7100 or better. (Intel 7th gen or newer Pentium or better, excluding J and N series)
 - RAM: 8GB or more
 - Storage: 60GB SSD storage for Jellyfin files and transcoding cache.
-- Graphics: Intel HD 6xx (7th gen integrated graphics) or newer, Nvidia GTX 16 / RTX 20 series or newer (excluding GTX 1650). Intel is recommended over Nvidia. AMD and Apple Silicon are not recommended.
+- Graphics: Intel HD 6xx (7th gen integrated graphics) or newer, Nvidia GTX 16 / RTX 20 series or newer (excluding GTX 1650). Intel is recommended over Nvidia. AMD is not recommended.
```
felix920506, line 20: The reasons Apple Silicon wasn't recommended when this guide was written are as follows:
1. No jellyfin-ffmpeg support
2. Poor encoding quality compared to QSV, NVENC and even AMF during my testing.

TimGels: for 2; That is not the case anymore then I assume?

gnattu: The upstream ffmpeg had a very broken VideoToolbox implementation until very recently. I back-ported  some patches to our own ffmpeg to make at least the full-range color works. But to verdict this I still need to run benchmarks to give us numbers.

docs/general/administration/hardware-selection.md
gnattu: N100-based mini PCs are less power-efficient than you might expect. In fact, many of them have higher power consumption idling than an M1 Mac mini during transcoding due to poor firmware.

docs/general/administration/hardware-selection.md
gnattu: Ran a quick test against an 30M bitrate 1080p movie, compressing to 3000K HEVC.

Here's the VMAF test result:

- NVENC: 97.845684
- QSV: 97.743323
- VideoToolbox: 97.579240

So the score is indeed lower, but not to an extent where "not recommend" should be labeled, at least on this sample and this bitrate. I will run additional tests to see if things are different.

docs/general/administration/hardware-acceleration/apple.md
gnattu: It is hard to tell "how old". Maybe I can just say "when the encoding task exceeds then capability of the hardware"?

docs/general/administration/hardware-selection.md
TimGels: Yea that delta between VideoToolbox and QSV is not something a user should worry about. I agree with the removal of not recommending apple silicon then, if this is the quality users can expect.

gnattu: I run an additional sample where all encoders performs bad at 3000K HEVC:

- NVENC: 59.085323
- QSV: 58.456923
- VideoToolbox: 59.149807

And in this case VT is slightly better in score.  But it does not make any sense to human as the visual quality is just so bad for all of them.

gnattu: With the same sample not giving ideal score, but now use the 3000K H264:

- NVENC: 58.908556
- QSV: 58.832961
- VideoToolbox: 58.740689

This turn VideoToolbox back to the lowest, but still no meaningful delta.

docs/general/administration/hardware-acceleration/apple.md
felix920506: That could work as well

felix920506 (approved):
Looks good to me

docs/general/administration/hardware-acceleration/apple.md
```diff
@@ -0,0 +1,84 @@
...
+
+Hardware accelerated transcoding is supported on all Macs that support [VideoToolbox](https://developer.apple.com/documentation/videotoolbox). This includes most 2011 and later Macs.<sup>*</sup>
+
+Full acceleration is available on most Macs from 2017 and later, with the exception of the MacBook Air (13-inch, 2017).
+
+To have native Apple Silicon support, Jellyfin server 10.9.0+ and `jellyfin-ffmpeg6` 6.0.1-4 or higher is required.
+
+While hardware acceleration via VideoToolbox might work on older series Macs, it is not officially supported.
+
+<sup>*</sup> VideoToolbox may fallback to software decoding/encoding when the task exceeds its hardware capability. The older your Mac, the more this will occur. Macs from 2017 and later are strongly recommended.
```
nyanmisaka, line 46: Even if the `allow_sw` option is [disabled by default](https://github.com/FFmpeg/FFmpeg/blob/c31d2326560087b80f8523b27060ad6388369cac/libavcodec/videotoolboxenc.c#L2849), will it automatically fall back to software encoding on partially supported devices?

gnattu: Both encoding and decoding have fallback behaviors. For encoding it will error out, but for decoding, the default option is to [allow software decoding for HEVC videos](https://github.com/FFmpeg/FFmpeg/blob/c31d2326560087b80f8523b27060ad6388369cac/libavcodec/videotoolbox.c#L816).

nyanmisaka: One more question: Is there an artificial limit on the number of concurrent encoding sessions on Apple like NVIDIA NVENC?

gnattu: There is no artificial limit, but there are some hard limits on certain models due to hardware limitations. On some ancient Intel Macs, the limit is only 1, as Intel could not perform concurrent encoding at that time. This should not be an issue on modern Macs anymore; I've run approximately 40 sessions without encountering any problems.

nyanmisaka: Good to know. I remember the Plex documentation mentioned a limit of up to 1 concurrent encoding. It might be worth add this in our documentation to avoid confusion.

gnattu: The problem is, I don't know exactly on which model this was an issue and since when this limitation was no longer applicable.

nyanmisaka: Just put what we already know into the docs. And emphasize that we have only verified this on Apple Silicon, newer Intel based Macs may also support this but use it at your own risk.

This is much better than Plex documents anyway.

> macOS is only capable of hardware-accelerated encoding of 1 video at a time. This is a platform limitation from Apple.

https://support.plex.tv/articles/115002178853-using-hardware-accelerated-streaming/

gnattu: Then I think just change "VideoToolbox may fallback to software decoding/encoding when the task exceeds its hardware capability." to " VideoToolbox may fallback to software decoding/encoding or stop working when the task exceeds its hardware capability." is enough. Users should not care about a historical limitation too much.

nyanmisaka: That should be enough. We will add automatic fallback to software transcoding on the server side in the future.

gnattu:
Updated the doc to match current development and platform verification status.

docs/general/administration/hardware-acceleration/apple.md
```diff
@@ -0,0 +1,102 @@
...
+
+Hardware accelerated HDR to SDR tone-mapping is supported on all Macs from 2017 and later, with the exception of the MacBook Air (13-inch, 2017).
+
+There are two different methods that can be used. Pros and cons are listed below:
+
+1. **Metal**
+
+   - Pros: Supports Dolby Vision P5, detailed fine-tuning options.
+
+   - Cons: Slower on low end GPU, especially on Intel iGPUs.
```
felix920506, line 26:
```suggestion
   - Cons: Slower on low end GPUs, especially on Intel iGPUs.
```

docs/general/administration/hardware-acceleration/apple.md
```diff
@@ -0,0 +1,102 @@
...
+
+Macs from 2017 and later, excluding the MacBook Air (13-inch, 2017), support decoding and encoding HEVC.
+
+### Transcode AV1
+
+There is no hardware accelerated path for AV1 on macOS at the moment. Although the M3 series added AV1 decoding support, [ffmpeg does not support it yet](https://trac.ffmpeg.org/ticket/10642).
+
+### Performance Consideration
+
+An Apple Silicon-based Mac is preferred in most cases. Even an entry-level M1 can handle three 4K 24fps Dolby Vision HEVC 10-bit transcoding tasks simultaneously while performing tone-mapping to SDR.<sup>1</sup> <sup>2</sup>
```
felix920506, line 66: What are the cases that Apple silicon macs aren't preferred?

docs/general/administration/hardware-acceleration/apple.md
```diff
@@ -0,0 +1,102 @@
...
+
+The "Max" variant chips come with an additional video encoding engine. VideoToolbox can utilize this extra engine even when there is only a single transcoding session, enabling support for 4K 120fps transcoding and tone-mapping.<sup>1</sup>
+
+The "Ultra" variant chips feature 2 video decoding engines and 4 video encoding engines, effectively doubling the capability compared to the "Max" variant chips.
```
felix920506, line 70: So Max has 2x media perf of normal and Pro, and Ultra has 2x perf of Max?

docs/general/administration/hardware-acceleration/apple.md
```diff
@@ -0,0 +1,102 @@
...
+
+On legacy Intel Macs, you may encounter performance issues with tone-mapping using Metal on 4K videos if your Mac doesn't have an AMD GPU. However, it is still adequate for transcoding SDR videos.
+
+<sup>1</sup> Using prefer speed encoder preset.
+
+<sup>2</sup> The simultaneous session count is a soft limit. You can run more sessions if you are willing to accept the per-session performance trade-offs.
```
felix920506, line 76:
```suggestion
<sup>2</sup> The simultaneous session count is a soft limit.
```
Telling users they can run more sessions means users overloading the media engine with too many streams asking why their video is stuttering

docs/general/administration/hardware-acceleration/apple.md
```diff
@@ -0,0 +1,102 @@
...
+
+Hardware accelerated transcoding is supported on all Macs that support [VideoToolbox](https://developer.apple.com/documentation/videotoolbox). This includes most 2011 and later Macs.<sup>*</sup>
+
+Full acceleration is available on most Macs from 2017 and later, with the exception of the MacBook Air (13-inch, 2017).
+
+To have native Apple Silicon support, Jellyfin server 10.9.0+ and `jellyfin-ffmpeg6` 6.0.1-5 or higher is required.
+
+While hardware acceleration via VideoToolbox might work on older series Macs, it is not officially supported.
+
+<sup>*</sup> VideoToolbox may fallback to software decoding/encoding or stop working when the task exceeds its hardware capability. The older your Mac, the more this will occur. Macs from 2017 and later are strongly recommended.
```
felix920506, line 46: The info in later sections strongly suggest that Apple SIlicon Macs are preferred. Should be listed here as well.

felix920506 (changes requested):
Had another look at the this PR, a few minor items.

docs/general/administration/hardware-acceleration/apple.md
gnattu: In a very marginal case, such as someone owning a high-end Intel Mac Pro with multiple GPUs added and disregarding power consumption entirely, it's possible that such a setup might perform better on multi-session transcodes and tone-mapping. However, since I haven't tested this kind of hardware at all, I'm not very sure about its performance. There are also other possible edge cases, such as someone wanting to use the CPU based encoding for specific reasons, where CPU-based HEVC encoding is much better supported on x86_64 platforms. Additionally, there may be other cases that I cannot think of at the moment. Those cases are too edge that I don't even want to document them here.

docs/general/administration/hardware-acceleration/apple.md
gnattu: Not quite. The Max only doubles the encoder of the normal and Pro variants. The Ultra doubles both the decoder and encoder of the Max. In real-life performance, it could be close to the 2x increase as you described, as the decoder decodes faster than the encoder can encode. My M1 Max runs at 1.75x the performance of the M1 when encoding 4K videos.

docs/general/administration/hardware-acceleration/apple.md
gnattu: That's why I said "if you are willing to accept the per-session performance trade-offs".

Perhaps changing it into a more clear statement to tell the user what will happen is better than just removing this sentence like:

"You can run more sessions if you want, but the transcoding performance of each session may be reduced to a point where video playback starts to stutter."

docs/general/administration/hardware-acceleration/apple.md
felix920506: Fair

docs/general/administration/hardware-acceleration/apple.md
felix920506: Could work as well, but make it clear to the users what the consequences of overloading the media engine are

docs/general/administration/hardware-acceleration/apple.md
```diff
@@ -0,0 +1,102 @@
...
+<sup>2</sup> The simultaneous session count is a soft limit. You can run more sessions if you want, but the transcoding performance of each session may be reduced to a point where video playback starts to stutter.
+
+## macOS Setups
+
+**macOS 12 and later** are officially supported. Older versions might work, but are not supported.
+
+### Configure
+
+- Enable VideoToolbox in Jellyfin and uncheck unsupported codecs.
+  - Disable `MPEG4`, `MPEG2`, and `VC-1`, as VideoToolbox lacks a hardware-accelerated decoding path for these codecs.
```
nyanmisaka, line 85: TBH you can eliminate these legacy hw codecs in the server and web. I don't really believe Apple will introduce them in the future.

gnattu: We also need to perform some server-side hacks to revert those options if the user enabled them before.

These options have been available for a long time and for a reason. Previously, ffmpeg allowed software fallback for all codecs, and these legacy codecs would use VideoToolbox's software fallback route. However, now ffmpeg only allows fallback for HEVC, and these codecs will error out because VT session creation would fail.

Is such change still good for 10.9?

nyanmisaka: We need a codec panel that can be checked automatically. At that time, the existing decoder/encoder settings will be discarded and Jellyfin will be automatically set based on the hardware probing results.

So deleting these codecs should have no impact.
https://github.com/jellyfin/jellyfin/blob/1124da5428b4cb0e26133f636b27634723026513/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L6167-L6175

felix920506 (approved):
Other than the points Nyan brought up, LGTM

docs/general/administration/hardware-acceleration/apple.md
gnattu: Updated doc and pushed changes to server: jellyfin/jellyfin#11217 and web: jellyfin/jellyfin-web#5325

nyanmisaka (approved):

docs/general/administration/hardware-acceleration/apple.md
```diff
@@ -0,0 +1,103 @@
...
+
+Hardware accelerated transcoding is supported on all Macs that support [VideoToolbox](https://developer.apple.com/documentation/videotoolbox). This includes most 2011 and later Macs.<sup>*</sup>
+
+Full acceleration is available on most Macs from 2017 and later, with the exception of the MacBook Air (13-inch, 2017).
+
+To have native Apple Silicon support, Jellyfin server 10.9.0+ and `jellyfin-ffmpeg6` 6.0.1-5 or higher is required.
+
+While hardware acceleration via VideoToolbox might work on older series Macs, it is not officially supported.
+
+If your Mac does not have an internal display, you may need to connect it to a monitor or use an HDMI dummy plug to prevent the GPU from being throttled.
```
felix920506, line 46:
```suggestion
If your Mac does not have an internal display, you may need to connect it to a monitor or use a dummy plug to prevent the GPU from being throttled.
```
There are dummy plugs for other display interfaces.

docs/general/administration/hardware-acceleration/apple.md
```diff
@@ -0,0 +1,103 @@
...
+
+2. Open the "Activity Monitor" and search for ffmpeg.
+
+3. If ffmpeg is not using a few hundred percent CPU, then hardware acceleration is working.
+
+:::note
+
+It would be normal if you see the GPU usage close to 0. For pure transcoding, everything is performed on a dedicated accelerator, and that will not count as GPU usage under macOS.
+
+:::
```
felix920506, line 103:
```suggestion
3. If ffmpeg is not using a few hundred percent CPU, then hardware acceleration is working. It is normal to see the GPU usage close to 0. For pure transcoding, everything is performed on a dedicated accelerator, and that will not count as GPU usage under macOS.
```
Doesn't need to be an admonition

docs/general/administration/hardware-acceleration/apple.md
gnattu: Well I can accept this suggestions, but I have to say it is either HDMI or USB-C for desktop Macs after 2017 and the USB-C one cost more usually.

felix920506:
It looks like there are merge conflicts on this one.

PR #878, opened 2024-02-29
Converted Twitter to X

TimGels:
Twitter is no more. We should use X now.

### DIFF
--- a/src/components/contact/XCorpCard.tsx
+++ b/src/components/contact/XCorpCard.tsx
@@ -15,7 +15,7 @@ const XCorpCard = () => (
       Follow us on X for release announcements and other updates, along with general musings.
     </div>
     <div className='card__footer'>
-      <a href='https://www.x.com/jellyfin' className='button button--block button--x'>
+      <a href='https://www.x.com/jellyfin' className='button button--block button--xcorp'>
         @jellyfin
       </a>
     </div>
--- a/src/components/contact/XCorpCard.tsx
+++ b/src/components/contact/XCorpCard.tsx
@@ -1,0 +1,25 @@
+import { SiX } from '@icons-pack/react-simple-icons';
+import React from 'react';
+
+import './ContactCard.scss';
+
+const XCorpCard = () => (
+  <div className='card card--contact'>
+    <div className='card__header' style={{ display: 'flex' }}>
+      <h3 className='margin-bottom--none' style={{ flexGrow: 1 }}>
+        X
+      </h3>
+      <SiX />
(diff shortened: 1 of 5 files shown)


felix920506 (approved):
~~What should we call tweets now that twitter is no more~~

JPVenson:
i am generally fine with changing the texts but would argue for sake of readability to keep twitter for code side of things

JPVenson (changes requested):
i am generally fine with changing the texts but would argue for sake of readability to keep twitter for code side of things

TimGels:
> i am generally fine with changing the texts but would argue for sake of readability to keep twitter for code side of things

what about xcorp?

TimGels:
> Yes, I agree with @JPVenson on that. I don't care what Elon wants to call his site, but `button-x` is so generic it's unreadable. Let's leave everything that isn't actually user-facing as-is. Or use `x-formerly-twitter`.

https://en.wikipedia.org/wiki/X_Corp. xcorp seems to be the new legal entity. i like that more than still dragging twitter along

JPVenson:
`button-xcorp` is _fine_ with me. Just not _just_ `x`

src/components/contact/XCorpCard.tsx
```diff
@@ -0,0 +1,25 @@
...
+      <h3 className='margin-bottom--none' style={{ flexGrow: 1 }}>
+        X
+      </h3>
+      <SiX />
+    </div>
+    <div className='card__body'>
+      Follow us on X for release announcements and other updates, along with general musings.
+    </div>
+    <div className='card__footer'>
+      <a href='https://www.x.com/jellyfin' className='button button--block button--x'>
```
JPVenson, line 18:
```suggestion
      <a href='https://www.x.com/jellyfin' className='button button--block button--xcorp'>
```

JPVenson (changes requested):

felix920506 (approved):
LGTM

src/components/contact/XCorpCard.tsx
TimGels: Good catch

JPVenson (approved):
lgtm

joshuaboniface (approved):

nielsvanvelzen:
We should just remove it, the Twitter (yes Twitter because look at the domain) account isn't even active anymore. A lot of stuff we posted on other social channels (even our Telegram channel!) was never posted to Twitter.

Also, you've changed the meta tags to invalid ones.

Issue #447, opened 2023-03-15
Developer website task collection


nielsvanvelzen:
I've started work on a developer site in the "developers" branch some time ago. This list collects all tasks that need to happen before & after launch of the site. We'll launch the site with minimal content and we can add more over time. The main tasks left are UX related. If you want to work on one of these things please comment/notify in Matrix to avoid several people working on the same thing.

## Before launch

- [ ] Somehow merge the developer docs and developer blog, some ideas:
  - Add recent blog posts to sidebar for docs, use that sidebar for blog posts
  - Add a second header menu beneath the existing one to easily switch between blog/docs, this menu could also contain the current Jellyfin version or something
- [ ] Show announcement banner indicating the developer site is still a work in progress
      This should ONLY show on the pages for the development site, not on the end-user docs/blog
- [x] Add an developer index page (#466)

## After launch

- [ ] Add page for Authentication (how to get an access token with QuickConnect/AuthenticateByName/etc)
- [ ] Add page for getting started with plugin dev
  - [ ] Add page describing the plugin repository format (maybe add a JSON schema?)
  - [ ] Add page describing the plugin meta format (maybe add a JSON schema?)
- [ ] Add page listing our SDK's
- [ ] Add page with API changelog (based on the kotlin-sdk openapi-generator compare tool)
- [ ] Add page listing contact persons for various roles (app publishing, security issues) and maintainers (SDKs & clients)
- [ ] Add page for the playback APIs

ferferga:
Can we also have some specific docs for playback requests maybe? The whole get mediasource and start a playback session it's huge black box, even to me.

nielsvanvelzen:
> Can we also have some specific docs for playback requests maybe? The whole get mediasource and start a playback session it's huge black box, even to me.

I've added it but I have no clue how they work myself either so that's why I didn't list it originally (the list is basically the to do I made for myself posted as an issue :p)

mcarlton00:
In regards to the authentication, I have this gist for username/password that can probably just be forklifted into wherever in the docs in belongs: https://gist.github.com/mcarlton00/f7bd7218828ed465ce0f309cebf9a247

nielsvanvelzen:
We can probably re-use most of the Kotlin SDK's authentication page too: https://kotlin-sdk.jellyfin.org/guide/authentication.html

nielsvanvelzen:
Closing due to lack of interest

sevenrats:
[...]

PR #890, opened 2024-03-07
Added a work in progress banner under the /developers page

PeachesMLG:
Delivers the announcement banner as per: https://github.com/jellyfin/jellyfin.org/issues/447

Adds the banner under all pages in the development directory
image

Keeping the ones in the normal directory the same
image

Wasnt exactly sure what our color schemes were so I copied a colour from another component I thought would work.

Happy for any feedback around this


src/theme/Root.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { useLocation } from '@docusaurus/router';
+import Banner from '../components/banner/Banner';
+
+// The root view wrapper for the entire app
+export default function Root({ children }: { children: React.ReactNode }) {
+  const isInDevelopersSection = useLocation().pathname.startsWith('/developers');
```
PeachesMLG, line 7: I dont like this since it feels very hacky, but it works and I couldnt find a better way to do this.

Happy for any feedback if anyone knows a better way to do this

grhallenbeck: First time I'm seeing this repo, but couldn't it go in the index for the developers section here?
https://github.com/jellyfin/jellyfin.org/blob/developers/content/developers/pages/index.tsx

I think that gets rendered when the route is under `/developers`, so we wouldn't need to check the base path in the location then.

grhallenbeck: Oh wait, I see. There's a `pages` and `content/docs` under developers. I wonder if the `docsRootComponent` config option might be a way to do this type of thing?
https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs#docsRootComponent

PeachesMLG: Sure, but then this wont show under this page, and ill still be left with the same issue trying to determine which is the normal and which is the developers
image

nielsvanvelzen: Yeah this is the main concern I had. Ideally we'd be able to somehow know we're in the developer part of the site without looking at the path, but I guess we can use this for now. The banner is meant to be temporary anyway.

PR #895, opened 2024-03-11
Refactor code to reduce duplication

PeachesMLG:
Refactored code in the developers page to reduce duplication by utilizing reusable components


nielsvanvelzen:
Can you format the files? The repository contains a prettier config so styling is consistent.

PeachesMLG:
I did format it, guess my ide didnt pick up prettier, any reason why we arnt using npmrc like we are in jellyfin-web? We could also then setup github actions to flag when its not correct

PR #896, opened 2024-03-11
Add publish workflow to deploy to Cloudflare Pages

nielsvanvelzen:
Draft until it works


nielsvanvelzen:
Closing for now as the workflow is quite outdated now so it needs to be updated anyway.

PR #900, opened 2024-03-13
Add AmpFin to the clients list

rasmuslos:
AmpFin is now available on the AppStore, so I thought I might add it to the clients list.

### DIFF
--- a/src/data/clients.ts
+++ b/src/data/clients.ts
@@ -863,5 +863,28 @@ export const Clients: Array<Client> = [
         url: 'https://fintunes.app'
       }
     ]
+  },
+  {
+    id: 'ampfin',
+    name: 'AmpFin',
+    description: 'iOS 17 native Jellyfin music player',
+    clientType: ClientType.ThirdParty,
+    deviceTypes: [DeviceType.Mobile],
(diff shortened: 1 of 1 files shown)


src/data/clients.ts
```diff
@@ -863,5 +863,28 @@ export const Clients: Array<Client> = [
...
       }
     ]
+  },
+  {
+    id: 'ampfin',
+    name: 'AmpFin',
+    description: 'iOS 17 native Jellyfin music player',
+    clientType: ClientType.ThirdParty,
+    deviceTypes: [DeviceType.Mobile],
+    licenseType: LicenseType.Proprietary,
```
nielsvanvelzen, line 873: Is this correct?

rasmuslos: Yes, it Is. The app is licensed under the Mozilla Public License 2 but the Common Clause also applies. So the app is only source available, and I think calling it open source would be misleading. Proprietary is a more accurate description.

nielsvanvelzen (approved):

nielsvanvelzen:
We've decided to not include this client in our listing due to the combination of it's proprietary license and the monetization.

I will close the pull request.

PR #906, opened 2024-03-14
Refactored the ContributeGuide.tsx file

PeachesMLG:


src/components/node/Node.tsx
PeachesMLG, line 1: This file is still fairly bulky, but I dont think there is much way around that

src/data/contributeData.tsx
PeachesMLG, line 1: This file is still fairly bulky, but I dont think there is much way around that

src/data/contributeData.tsx
```diff
@@ -0,0 +1,284 @@
+import { NodeData } from '../components/node/Node';
+import React from 'react';
+import Link from '@docusaurus/Link';
+
+export const contributeData: NodeData = {
+  links: [
```
gnattu, line 6: I'm sorry, but I really don't think a huge `Nodedata` object is more maintainable and easier to work with than JSX. This will cause extra pain if we ever want to change this page.

src/components/node/Node.tsx
```diff
@@ -0,0 +1,38 @@
...
+  body?: ReactNode;
+  links: NodeLink[];
+}
+
+interface NodeLink {
+  title: string;
+  node: NodeData;
+}
+
+export default function Node({ data }: { data: NodeData }) {
```
gnattu, line 14: This is a terrible name for this "component" and I think it is meaningless to have a component like this because the idea passing "node data" is just terrible. If you really want a re-usable component, made a "Tabs" component which switch between displaying multiple panels of content and use saner syntax like:

 ```jsx
<Tabs id="TabsExample" onChange={this.handleTabChange} selectedTabId="code">
    <Tab id="code" title="Code" panel={<CodePanel />} />
    <Tab id="translations" title="Translations" panel={<TranslationsPanel />} />
    <Tab id="other" title="Other" panel={<OtherPanel />} />
</Tabs>
```

gnattu (changes requested):
I personally do not like what it is in its current form. It seems to achieve the goal of reducing the original file's line count, but it introduces unnecessary complexity and makes the code even harder to maintain.

nielsvanvelzen:
This file is moved in the developers branch, unless actual content changes are made i don't see the point of making changes to it as it only complicates rebasing the dev branch.

PeachesMLG:
I dont believe it has moved in the developer branch? still appears to be there?

https://github.com/jellyfin/jellyfin.org/blob/master/src/components/contribute/ContributorGuide.tsx

src/data/contributeData.tsx
PeachesMLG: I disagree to be honest, it means if we want to change any of the stucture of a "node" we can do so very easily in the Node.tsx file, I also dont see how this is harder to edit than the previous one.

Doing it this way also means that we can unit test this alot easier than before.

src/components/node/Node.tsx
PeachesMLG: I made a new pr here: https://github.com/jellyfin/jellyfin.org/pull/907

See if this option looks better

PR #917, opened 2024-03-22
Update linux.md fedora install instructions

grecko123:
Add essential step of reloading the firewall rules, so as to apply the newly loaded changes.


docs/general/installation/linux.md
```diff
@@ -96,7 +96,13 @@ However [`rpmfusion`](https://rpmfusion.org/) provides both `jellyfin-server` an
 
    :::
 
-6. Go to `localhost:8096` or `ip-address-of-jellyfin-server:8096` to finish setup in the web UI
+6. Reload the firewall to apply the new rules:
+
+   ```sh
+   sudo firewall-cmd --list-services
```
felix920506, line 102: I don't think this is the right command for reloading firewalld

felix920506 (changes requested):

docs/general/installation/linux.md
Shadowghost: Correct one should be `sudo firewall-cmd --reload`

joshuaboniface:
So I want to jump in this one and suggest, that we also update it to leverage the RPMFusion packages/repo instead of our, since the latter goes away with 10.9.0.

sigaloid:
+1, this blocked me from setting it up for a little while

docs/general/installation/linux.md
grecko123: wow- i do know that :joy:  

and i dont know how I pushed this, must have been in a rush

felix920506 (approved):

PR #924, opened 2024-03-26
doc: add instructions on how to test unstable builds on macOS

gnattu:
This add instructions on how to test the unstable builds on macOS. 

Testing running master branch should not be too much different from other platform using command line, if we have that one ready I can just add a link to the doc.


docs/general/testing/macos.md
```diff
@@ -0,0 +1,71 @@
+---
+uid: testing-macos
+title: macOS
+description: Test unstable builds on macOS.
+sidebar_position: 2
+---
+
+# Test unstable builds on macOS
+
+Testing unstable builds requires you to be confident with the command line, or at least not afraid of it.
```
felix920506, line 10:
```suggestion
Testing unstable builds requires you to not be afraid of using the command line.
```

docs/general/testing/macos.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Get Unstable Server
+
+Unlike the release version, the unstable builds for macOS will not come with a packaged DMG or application. Instead, you have to run them from the command line.
```
felix920506, line 14:
```suggestion
Unlike the stable releases, the unstable builds for macOS do not come with a packaged installer or application. Instead, they have to be ran from the command line.
```

docs/general/testing/macos.md
```diff
@@ -0,0 +1,71 @@
...
+
+The latest unstable builds for packaged jellyfin can be obtained here:
+
+- [Intel Macs](https://repo.jellyfin.org/?path=/server/macos/latest-unstable/amd64)
+- [Apple Silicon Macs](https://repo.jellyfin.org/?path=/server/macos/latest-unstable/arm64)
+
+There is no difference in content between the `tar.gz` and the `tar.xz` archives; they are only compressed in different formats. Either one will work.
```
felix920506, line 21:
```suggestion
There is no difference in content between the `tar.gz` and the `tar.xz` archives. They are only compressed in different formats and either one will work.
```

docs/general/testing/macos.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Get Jellyfin FFmpeg
+
+The unstable version doesn't come with FFmpeg by default, so you need to obtain FFmpeg manually.
```
felix920506, line 25:
```suggestion
The unstable builds don't come with FFmpeg bundled. Instead, the binaries have to be downloaded separately.
```

docs/general/testing/macos.md
```diff
@@ -0,0 +1,71 @@
...
+
+Jellyfin's custom FFmpeg is available [here](https://repo.jellyfin.org/?path=/ffmpeg/macos).
+
+Select the version corresponding to the unstable version of the server.
+
+Intel Macs should download the `AMD64` binaries.
+
+Apple Silicon Macs should download the `ARM64` binaries.
```
felix920506, line 33: I see the difference between AMD64 and ARM64 mentioned twice. Maybe it would be cleaner to mention it once on the top?

docs/general/testing/macos.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Prepare The Unstable Server
+
+After you have downloaded both the server and FFmpeg, you should extract them in a folder you like.
+
+Before running, we need to remove quarantine flag and do an ad-hoc sign to the server binaries:
```
felix920506, line 39:
```suggestion
After downloading both the server and FFmpeg, extract them to a folder you like. Before running, the quarantine flag needs to be removed and an ad-hoc signature needs to be created for the binaries.
```

docs/general/testing/macos.md
```diff
@@ -0,0 +1,71 @@
...
+
+Open a command line interface. You can use the `Terminal.app` come with macOS, then run following commands:
+
+```shell
+cd /path/to/jellyfin
+xattr -rd com.apple.quarantine .
+codesign -fs - --deep jellyfin
+```
```
felix920506, line 47:
```suggestion
Open a terminal in the directory containing the extracted contents of the server package. This can be done by right-clicking on the directory and selecting `open a terminal here`. Run the following commands in the terminal:
```shell
xattr -rd com.apple.quarantine .
codesign -fs - --deep jellyfin
```
```

I don't actually know the exact string for that option since my system language isn't English but you get the idea.

File: docs/general/testing/macos.md [hunk 8]
```diff
@@ -0,0 +1,71 @@
...
+## Prepare The FFmpeg
+
+We also need to perform the similar thing to our ffmpeg binaries, but this time we only have to remove the quarantine flag as the ad-hoc signature will be automatically created on first launch.
+
+Open another command line interface, then run the following commands:
+
+```shell
+cd /path/to/jellyfin-ffmpeg
+xattr -rd com.apple.quarantine .
+```
```
felix920506, line 58: If this section and the section before it are doing the same thing, can they be combined into one section?

docs/general/testing/macos.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Run The Unstable Server
+
+At this point, you should be able to run unstable jellyfin with the specified ffmpeg in the command line:
```
felix920506, line 62:
```suggestion
At this point, the binaries should be ready for testing. Run the following command in a terminal to start Jellyfin:
```
There should probably be a reminder to backup any existing Jellyfin installs in case of the need to downgrade to a previous Jellyfin version

docs/general/testing/macos.md
```diff
@@ -0,0 +1,71 @@
...
+
+```shell
+/path/to/jellyfin-folder/jellyfin --ffmpeg /path/to/ffmpeg-folder/ffmpeg
+```
```
felix920506, line 66: Won't it cause problem if the user doesn't first cd to /path/to/jellyfin ?

docs/general/testing/macos.md
```diff
@@ -0,0 +1,71 @@
...
+
+After pressing enter, you should see Jellyfin up and running, with logs appearing in the console.
```
felix920506, line 68:
```suggestion
After running the command, Jellyfin should startup in the terminal window.
```

docs/general/testing/macos.md
```diff
@@ -0,0 +1,71 @@
...
+
+To stop it, you press `CTRL+C` in the terminal window.
```
felix920506, line 70:
```suggestion
To stop Jellyfin, press `CONTROL+C` in the terminal window, and Jellyfin will shutdown gracefully.
```
the CTRL key is labeled `control` on mac keyboards.

docs/general/testing/macos.md
gnattu: There will be not "Open a terminal here" option on standard macOS install.

felix920506: Oh sorry I forgot that I have an automator script installed for that

docs/general/testing/macos.md
gnattu: we are using absolute path to the jellyfin binary, so this won't be a issue.

docs/general/testing/macos.md
gnattu: This is actually the first time it appears as human-readable form. The former one is in a link and the user will just see "Intel" and "Apple Silicon"

felix920506: What I meant was put the Apple Silicon -> ARM64, Intel -> AMD64 on top and make both the upper level links

felix920506:
This branch is out of date with master, and the changes that are in master but not here includes changes to the testing category. This page should ideally be under `server` in testing rather than alongside it.

gnattu:
> This branch is out of date with master, and the changes that are in master but not here includes changes to the testing category. This page should ideally be under `server` in testing rather than alongside it.

I rebased from master, but to make this possible, we need to create a folder for the server and make the current `server.md` its `index.md`. Such a change is not appropriate for me to make in this branch.

felix920506:
Nah just do it here in this PR

felix920506 (approved):

PR #934, opened 2024-03-30
Add warning for network path deprecation in Kodi native mode documentation

nielsvanvelzen:
Considered removing the entire section but that would be confusing as the client still contains the native mode.

### DIFF
--- a/docs/general/clients/kodi.md
+++ b/docs/general/clients/kodi.md
@@ -95,6 +95,12 @@ Native mode accesses your media files directly from the filesystem, bypassing th
 
 To use Native mode, first set up your libraries in Jellyfin with a remote path.
 
+:::caution
+
+Starting from Jellyfin 10.9 it is no longer possible to set the shared network folder.
+
+:::
+
 1. In the Jellyfin server, navigate to the Libraries section of the admin dashboard.
    - Select an existing library (or create a new one)
    - Select the media folder


docs/general/clients/kodi.md
```diff
@@ -95,6 +95,12 @@ Native mode accesses your media files directly from the filesystem, bypassing th
 
 To use Native mode, first set up your libraries in Jellyfin with a remote path.
 
+:::caution
+
+Starting from Jellyfin 10.9 it is no longer possible to set the shared network folder.
```
felix920506, line 100: Will existing setups stop working after upgrading the server to 10.9? If so then that should be documented.

nielsvanvelzen: nope, but we may remove the paths entirely in the future although nothing is decided.

felix920506 (approved):

Shadowghost (approved):

nielsvanvelzen:
The future release label means "do not merge until the next release"

Shadowghost:
Sorry, was on mobile and missed it.
Nonetheless in cases like this where the release is explicitly referenced I don't see a reason why we shouldn't merge it ahead of actual release.

nielsvanvelzen:
Yeah for this specific PR it doesn't matter that much, but generally we want to wait with merging documentation for new versions until that version is available.

PR #937, opened 2024-04-04
Add Subtitle Extract plugin

ericswpark:
From https://github.com/jellyfin/jellyfin-plugin-subtitleextract/issues/22

This PR adds the Subtitle Extract plugin to the documentation and also encourages new Jellyfin administrators to install them.

Honestly I think the SE plugin should be integrated into Jellyfin given the horrible experience users will get without it, but I remember seeing a PR that integrated it into Jellyfin getting closed because it relied on some side effect? Can't find the URL for it at the moment but I remember @crobibero commented on it, and he did spin it out to the SE plugin.

Finally, I'm not sure about the wording for recommending individual plugins -- I do get that some people will find this recommendation useless, but I honestly do believe that a majority if not all Jellyfin admins would want to have this plugin by default. Without it, the UX of enabling and viewing subtitles becomes really annoying for viewers as it can take more than a couple of seconds to sometimes a couple of minutes (depending on server load, file size that impacts how long ffmpeg takes to map the subtitle stream out from the container, etc.) for the subtitle to show. Please suggest any alternative wording if you think there's a better option!

### DIFF
--- a/docs/general/server/plugins/index.mdx
+++ b/docs/general/server/plugins/index.mdx
@@ -163,6 +163,18 @@ Download subtitles from the internet to use with your media files from [Open Sub
 
 - [GitHub](https://github.com/jellyfin/jellyfin-plugin-opensubtitles)
 
+#### Subtitle Extract
+
+[![Language](https://img.shields.io/github/languages/top/jellyfin/jellyfin-plugin-subtitleextract.svg)](https://github.com/jellyfin/jellyfin-plugin-subtitleextract)
+[![Contributors](https://img.shields.io/github/contributors/jellyfin/jellyfin-plugin-subtitleextract.svg)](https://github.com/jellyfin/jellyfin-plugin-subtitleextract)
+[![License](https://img.shields.io/github/license/jellyfin/jellyfin-plugin-subtitleextract.svg)](https://github.com/jellyfin/jellyfin-plugin-subtitleextract)
+
+Plugin to automatically extract embedded subtitles in media. This avoids delayed subtitles during streaming if the client does not support direct play and requests subtitles.
+
+**Links:**
+
+- [GitHub](https://github.com/jellyfin/jellyfin-plugin-subtitleextract)
+
 #### Playback Reporting
 
 [![Language](https://img.shields.io/github/languages/top/jellyfin/jellyfin-plugin-playbackreporting.svg)](https://github.com/jellyfin/jellyfin-plugin-playbackreporting)


felix920506:
Adding the plugin to the list of plugins is fine. However we are still discussing internally whether we want a recommended plugins sections or not.

ericswpark:
@felix920506 that's what I figured. Feel free to leave the PR open until the discussion is resolved or close it altogether for an alternate solution.

felix920506:
We have decided against having a recommended plugins section in our docs. I will review your other changes in this PR shortly.

docs/general/server/plugins/index.mdx
```diff
@@ -163,6 +169,21 @@ Download subtitles from the internet to use with your media files from [Open Sub
...
+Plugin to automatically extract embedded subtitles in media. This avoids delayed subtitles during streaming if the client does not support direct play and requests subtitles.
+
+
+**Links:**
+
+- [GitHub](https://github.com/jellyfin/jellyfin-plugin-subtitleextract)
+
+
+
 #### Playback Reporting
```
felix920506, line 187: Why are there so many blank lines?

felix920506 (changes requested):

docs/general/server/plugins/index.mdx
ericswpark: Fixed

ericswpark:
@felix920506 I've removed the recommended section and removed the extra blank lines.

felix920506 (approved):

ericswpark:
@felix920506 @crobibero just wondering what you guys meant by the plugin being added as a "troubleshooting tip" and it depending on a "side effect." Because on my server, with files as small as 20 GB the subtitles take a while to mux out with ffmpeg. I've seen wait times around 30+ seconds to 90 seconds even, and even longer than that if there are other processes competing for I/O.

Granted this is only for sessions that can't direct play the file, but that's nearly all of the play sessions from our server as we use a mix of different devices with apps/browsers used to access Jellyfin. In addition, if we are playing outside the home network and the bitrate goes over our uplink speed then it'll always transcode and we'll run into the subtitle delay. What exactly do you guys mean with the above? Is this unexpected behavior on Jellyfin's side, or something else that I'm not understanding clearly here?

felix920506:
There is a troubleshooting page in the documentation.
https://jellyfin.org/docs/general/administration/troubleshooting

ericswpark:
@felix920506 right, but I was just curious as to why this plugin is considered a troubleshooting step. Unless I'm mistaken, all Jellyfin admins will run into this with giant files that have subtitles muxed in.

Unless the intended behavior of Jellyfin is show "fetching additional data" for several minutes while the server works on extracting the subtitle data over and over again when it is requested, I don't think this plugin should be deemed a "troubleshooting step."

felix920506:
This isn't really that common of an issue among the JF users that show up in the various troubleshooting channels.
It is a troubleshooting step in that most people will think that it is a performance issue when they see it.

PR #938, opened 2024-04-05
doc: updated safari codec support

gnattu:
This update reflects the current Safari codec support status. However, the iOS and Swiftfin sections have not been updated because I'm not entirely sure about their status. It's unlikely that VP9 will work on iOS clients because it is almost exclusively implemented for Google's YouTube, as the app requires a special entitlement to use the VP9 decoder(VT will report codec not supported for apps missing this entitlement). The AV1 decoder does not have such limitation, but I don't have a compatible device to test the real situation with our app, but it should work when used with system player.

I'm also a little bit confused about the usage of 🔶 and ✅, because both are used when platform quirks apply.


felix920506:
Swiftfin uses libvlc for playback. Does that not support VP9 on iOS?

gnattu:
> Swiftfin uses libvlc for playback. Does that not support VP9 on iOS?

Yes, but through software decoding.

felix920506 (approved):

PR #954, opened 2024-04-17
Add warning for containers on not linux

felix920506:
Explicitly tell users to not use containers on Windows or macOS.

Using containers on these platforms can be hard to troubleshoot and will lead to an inferior setup (eg. no hw accel)

### DIFF
--- a/docs/general/installation/container.md
+++ b/docs/general/installation/container.md
@@ -24,6 +24,12 @@ Additionally, there are several third parties providing unofficial container ima
 
 [Docker](https://www.docker.com/) allows you to run containers on Linux, Windows and MacOS.
 
+:::warning
+
+While it is possible to run Docker on Windows or macOS, it is an **UNSUPPORTED** configuration for Jellyfin. Please install Jellyfin natively if you wish to use Windows or macOS.
+
+:::
+
 The basic steps to create and run a Jellyfin container using Docker are as follows.
 
 1. Follow the [official installation guide to install Docker](https://docs.docker.com/engine/install).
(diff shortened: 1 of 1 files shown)


docs/general/installation/container.md
```diff
@@ -24,6 +24,12 @@ Additionally, there are several third parties providing unofficial container ima
 
 [Docker](https://www.docker.com/) allows you to run containers on Linux, Windows and MacOS.
 
+:::warning
+
+While it is possible to run Docker on Windows or macOS, it is an **UNSUPPORTED** configuration for Jellyfin. Please install Jellyfin natively if you wish to use Windows or macOS.
```
nielsvanvelzen, line 29: I wouldn't say it's unsupported as docker/podman work fine on these platforms. The only pain point would be hardware acceleration.

joshuaboniface: Maybe "not recommended" instead? Curious, is this something that comes up a lot and requires a disclaimer, or is it just an opinion thing?

JPVenson:
I think that warning is missleading.

Jellyfin does not have windows docker containers, but running Jellyfin in WSL/WSL2 is nothing out of the ordenary.

docs/general/installation/container.md
```diff
@@ -26,7 +26,7 @@ Additionally, there are several third parties providing unofficial container ima
 
 :::warning
 
-While it is possible to run Docker on Windows or macOS, it is not recommended for Jellyfin, as certain features would be unavailable. Please install Jellyfin natively if you wish to use Windows or macOS.
+While it is possible to run Jellyfin in Docker on Windows or macOS, it is not supported. Please install Jellyfin natively if you wish to use [Windows](/docs/general/installation/windows) or [macOS](/docs/general/installation/macos).
```
JPVenson: please add reasons

felix920506: added

JPVenson (approved):

PR #956, opened 2024-04-17
fix traefik v2.x doc: x-xss-protection description/value mismatch

flo1t:
description says it should be 1, but the value is 0


liampwll:
This should likely be removed entirely as it's not supported by any current browser: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection#browser_compatibility

liampwll:
See also: #372 #373

nielsvanvelzen:
This change looks fine to me (it corrects the line to what the comment says it should do) so I'll merge it as-is. I do agree however that we may want to remove this option entirely across all our configurations, perhaps you are interested in creating a pull request for that @liampwll?

nielsvanvelzen (approved):

PR #963, opened 2024-04-20
Hardware Selection Guide Cleanup

felix920506:
** Please do NOT merge until after 10.9 **

Cleaning up the hardware selection page and general updating.


docs/general/administration/hardware-selection.md
```diff
@@ -88,89 +86,56 @@ When using Intel ARC Graphics, Resizable BAR is recommended. Disabling it will r
...
+It is recommended to add more memory on Windows 11 due to the OS being heavier.
 
 ### Graphics Cards (GPUs)
 
 With modern systems, the media engines are usually located on the GPUs. Therefore, the available hardware acceleration options are determined by the choice of GPU.
 
-Encoder Quality: Intel > Nvidia > AMD > Apple
-
-Intel graphics is recommended over Nvidia graphics because of the quality. AMD or Apple Silicon are not recommended since they produce poor quality H.264 and H.265(HEVC) output.
+Encoder Quality: Apple ≥ Intel > Nvidia >>> AMD
```
gnattu, line 97: The same as above. This is not constantly true and is depending on the input videos.

docs/general/administration/hardware-selection.md
```diff
@@ -88,89 +86,56 @@ When using Intel ARC Graphics, Resizable BAR is recommended. Disabling it will r
...
 
-:::note A Note on AMD AV1 Encoders
+Intel graphics is recommended over Nvidia graphics because of the quality. AMD is not recommended because of poor quality H.264 and H.265(HEVC) output.
```
gnattu, line 99: The reason to recommended Intel should not only be based on quality as the difference of quality is relatively small on modern hardware. But it is still much cheaper and has more mature open-source GPU driver available on Linux(although broken at the moment)

docs/general/administration/hardware-selection.md
```diff
@@ -88,89 +86,56 @@ When using Intel ARC Graphics, Resizable BAR is recommended. Disabling it will r
...
+Intel CPUs with a model description that ends with F don't have integrated graphics. These are intended to be used with a dedicated graphics card. Please beware if you intend to use Intel integrated graphics.
 
-:::
-
-:::note Intel 12th Gen, 13th Gen and ARC Support
-
-If you are planning to use Linux with Intel 12/13th Gen integrated graphics or ARC, these GPUs only work on Linux Kernel 6.2 or newer. Please check your distribution to make sure it has a supported Linux Kernel version. It is also recommended that `jellyfin-ffmpeg6` be installed instead of the default `jellyfin-ffmpeg5` on these hardware platforms.
-
-:::
+If you are planning to use Linux with Intel 12/13th Gen integrated graphics or ARC, these GPUs only work on Linux Kernel 6.2 or newer. Please check your distribution to make sure it has a supported Linux Kernel version.
```
gnattu, line 119: A link to the "known limitations" section of Intel acceleration could also be helpful as Intel breaks their drivers from time to time and will have more constraints on the kernel version.

docs/general/administration/hardware-selection.md
```diff
@@ -9,55 +9,55 @@ This page covers what you need to know in order to select appropriate hardware f
...
-
-:::
+- Intel "Atom" CPUs: Intel J/M/N/Y series low power CPUs up to 11th gen use a different architecture than higher end parts, leading to subpar performance despite what their names might suggest. Please be careful about these parts when shopping for a system to run Jellyfin.
+- Prebuilt NAS devices: Prebuilt NAS devices often have low powered CPUs underpowered for a good Jellyfin experience. Please lookout for the specs
```
gnattu, line 57: The most problematic part is not the hardware, as the CPU is actually good enough for many of them, and some have already transitioned to using Alder Lake-N or similar processors. The more problematic part is the software stack required to install Jellyfin on them, which is usually very vendor-specific and may break Jellyfin, especially the hardware acceleration. If that happens, Jellyfin cannot provide much help for that specific hardware-software combination and cannot fix it.

gnattu (commented):
Left a few comments that I think is not quite accurate.

It's very risky to provide power consumption numbers, as the comparison needs to be conducted in a highly controlled lab environment to make any sense, especially for loaded power consumption. This figure will vary significantly depending on the workload and user configuration. Detailed lab testing is really out of the scope of this document.

docs/general/administration/hardware-selection.md
felix920506: Many still have 11th gen or below Intel Atom or even RTL something. I will change this section to be watch out but not out right not recommend.

docs/general/administration/hardware-selection.md
felix920506: Linked.

docs/general/administration/hardware-selection.md
```diff
@@ -68,13 +74,11 @@ The CPU will be responsible for many tasks, notably:
...
-
-:::caution
+Audio codecs are very lightweight and most CPUs should be able to handle them without issues (Ryzen 5950X FLAC -> AAC transcoding runs at ~110x real time). Please note that they will only use one core.
 
-Newer codecs can also require software decoding if there are no available hardware decoders. High bitrate HEVC content such as commonly found in 4k HDR can be very demanding even on modern CPUs. Please considering adding a modern GPU with 10bit HEVC decoding support. HDR to SDR tone-mapping is also currently not available without hardware acceleration.
+Video in unsupported codecs are usually older formats that are easier to decode. Assuming hardware acceleration is properly configured, any modern CPU with 4 threads should be able to handle the workload.
 
-:::
+However, newer codecs can also require software decoding if there are no available hardware decoders. HEVC, VP9 and AV1 can be **VERY** demanding even on modern CPUs if hardware acceleration isn't available. This is why a hardware platform capable of HEVC 10bit hardware decoding is strongly recommended.
```
gnattu, line 81: We have some software optimization for AV1 specifically. AV1 has a better optimized dav1d decoder, and we are enabling it in our custom ffmpeg. This decoder outperforms other software modern codecs like HEVC and VP9, and can provide good enough playback for most videos on modern CPUs. An Apple M1 can software decode up to 8K at 60fps (using all cores), although actual transcoding performance may be lower as the CPU has to be used for other tasks. As long as your CPU isn't too underpowered, you can decode 4K AV1 with this decoder without affecting playback too much, even an i3-7100 will do the trick (at the cost of 100% CPU load) if you are using hardware encoding with this software decoder.

This is helpful for users with older hardware that does not support AV1 but is still capable of decoding other codecs. We can let our users know that they don't have to panic if their hardware isn't new enough to support AV1 decoding because this codec has a better decoder and will have less impact on performance. Of course, it's still hard to beat hardware decoders and it will be less power-efficient. However, users can choose between buying a new GPU or saving money and using current hardware, especially when the GPU is integrated.

docs/general/administration/hardware-selection.md
```diff
@@ -9,55 +9,61 @@ This page covers what you need to know in order to select appropriate hardware f
...
-Atom CPUs in this guide refer to any Intel Atom CPUs, as well as other low powered Intel CPUs such as J/M/N/Y series from other product lines. (eg. Pentium N5105, Pentium N6000, Core i3 N305). While these CPUs are branded as products from other product lines, they use the same microarchitecture as Intel's low powered chipsets and may be significantly less powerful and have less features than their names might otherwise suggest.
+### Hardware Encoder Quality
 
-:::
-
-:::note These are Recommended Specs
+Different vendors have different hardware encoder implementations and produce different results. Usually, newer generations within the same vendor will provide better results. The following is a quick comparison on the quality between vendors on modern products.
 
-These specs are the **recommended** specs to run Jellyfin. They are not minimum requirements, and it is certainly possible to run Jellyfin on lower end hardware.
+Apple ≥ Intel ≥ Nvidia >>> AMD
```
gnattu, line 26: I personally still think `≈` is more accurate than `≥`, but if you want to use `≥` it is still OK, but it would be better to add a comment that the result may vary due to encoder settings, and this currently reflects what Jellyfin produces

docs/general/administration/hardware-selection.md
felix920506: 4K AV1 pinning i3-7100 is still quite heavy considering what type of devices we usually see users running JF on so I am against removing AV1 from the list. 

I would add something like "The AV1 software decoder is more efficient than HEVC and VP9, and AV1 media isn't common yet, so it shouldn't be an issue to not have AV1 decoding for now."

gnattu: That decoder (dav1d) is not enabled in all ffmpeg builds and is not the default one comes with ffmpeg, so you may need to mention that as well. Our jellyfin-ffmpeg is built with that decoder.

docs/general/administration/hardware-selection.md
gnattu: image

I just see a table from an audio codec testing website, and this could suite our needs

gnattu: We can make a similar table to define what does these symbols mean. We don't have to use the exact symbol they are using, just make a clear definition could be good enough.

felix920506: I don't think ≥ is that confusing and I've seen it in many other places

docs/general/administration/hardware-selection.md
```diff
@@ -53,6 +53,7 @@ For users with expensive power or running servers on battery power:
 
 - Intel 12th gen N series platforms
 - Apple M Series mac mini
+- Intel-Based [Framework Laptop Mainboards](https://frame.work/marketplace/mainboards)
```
gnattu, line 56: Why? We really should not endorse specific vendor made boards like this, and they are not better than a lot of other vendors.

gnattu (changes requested):

docs/general/administration/hardware-selection.md
felix920506: 1. It is the only laptop manufacturer that officially supports using the mainboard outside the enclosure
2. It only uses a few watts at idle
3. It is something that our community members may have lying around
4. It is normal Intel CPUs which we already recomment

thats why I felt it was an appropriate inclusion in this section

gnattu: > It is the only laptop manufacturer that officially supports using the mainboard outside the enclosure

Why you want to use a laptop motherboard outside a laptop enclosure in the first place?

> It only uses a few watts at idle

Same for a lot of other NUCs using laptop CPUs, and you want to list all of them?

Fun fact: Using less power than an Alderlake-N processor at idle doesn't necessarily mean that the board or the SoC on that board is power-efficient enough. It just means that Alderlake-N based systems are poorly made and nobody bothers to fine-tune the power consumption.

Fun fact 2: Both RK3588 and M1 can achieve single-digit (< 10) watt power consumption during transcoding (without fancy processing like tone mapping). Being good at idling is not the only reason they are recommended. Although the real-world case is unlikely to be this ideal, but it is still way better than what the Intel Laptop chips can archive. 

> It is something that our community members may have lying around

Same for a lot of other NUCs, laptops, etc.

> It is normal Intel CPUs which we already recommend

But this is the Low-Power section, right? And we are recommending a series of CPUs that are known to be the worst of their class in power consumption, which looks absurd to me. Intel laptops after Alderlake consistently perform as the worst in battery life, which means the actual power consumption during typical loads is not something we can recommend.

felix920506: > Why you want to use a laptop motherboard outside a laptop enclosure in the first place?

For a home server, many parts of a laptop are simply not needed and the option to put it in an alternative enclosure can save space

> Being good at idling is not the only reason they are recommended.

A server is likely going to be sitting idle most of the time. Many heavy tasks are only ran once in a few months at most by default. The heaviest task a JF server is likely to see during normal use is maybe an hour of OpenCL tonemapping

gnattu: > For a home server, many parts of a laptop are simply not needed and the option to put it in an alternative enclosure can save space
> A server is likely going to be sitting idle most of the time. Many heavy tasks are only ran once in a few months at most by default. The heaviest task a JF server is likely to see during normal use is maybe an hour of OpenCL tonemapping

Then we go back to "Same for a lot of other NUCs using laptop CPUs", why this vendor and why not using an NUC but a bare board from a laptop?

The NUCs are not more power-hungry, are cheaper, and actually come with smaller motherboard. If you think the power consumption of the Intel laptop chips are good enough for recommendation, then why not recommending those but this vendor's laptop motherboard? It is almost the least reasonable choice.

Also, a lot of users won't run the server with Jellyfin only, and multitasking is one of the easiest way to make the Intel chips behave poorly in terms of power consumption because the chip is harder to get into the deep sleep state to reduce power consumption.

felix920506: Fair

gnattu (approved):

nielsvanvelzen (approved):

nyanmisaka (approved):

docs/general/administration/hardware-selection.md
```diff
@@ -9,55 +9,61 @@ This page covers what you need to know in order to select appropriate hardware f
...
-- CPU (Without dGPU): Intel Pentium G4560, Intel Core i3-7100 or better. (Intel 7th gen or newer Pentium or better, excluding J and N series)
-- RAM: 8GB or more
-- Storage: 60GB SSD storage for Jellyfin files and transcoding cache.
-- Graphics: Intel HD 6xx (7th gen integrated graphics) or newer, Nvidia GTX 16 / RTX 20 series or newer (excluding GTX 1650). Intel is recommended over Nvidia. AMD and Apple Silicon are not recommended.
+Some component recommendations do not change regardless of the hardware configuration:
 
-:::note Intel "Atom" CPUs
+- Storage: 100GB SSD for Jellyfin files and transcoding cache. (Consider adding more on Windows 11)
+- Networking: Gigabit Ethernet Adapter or faster, WiFi or Powerline not recommended.
+- Internet Connection: At least 20mbps upload bandwidth (for remote access)
```
JPVenson, line 20: this also is highly user depended and will not really fit most. 20mbps will be consumed pretty quickly if no bandwith limit is configured.

docs/general/administration/hardware-selection.md
```diff
@@ -9,55 +9,61 @@ This page covers what you need to know in order to select appropriate hardware f
...
 
-- CPU (With dGPU): Intel Core i5-2300, AMD Ryzen 3 1200 or better. (Geekbench 6 Multicore 1500 or better). *CPU vendor will not have a meaningful impact on quality and/or speed of transcoding.*
-- CPU (Without dGPU): Intel Pentium G4560, Intel Core i3-7100 or better. (Intel 7th gen or newer Pentium or better, excluding J and N series)
-- RAM: 8GB or more
-- Storage: 60GB SSD storage for Jellyfin files and transcoding cache.
-- Graphics: Intel HD 6xx (7th gen integrated graphics) or newer, Nvidia GTX 16 / RTX 20 series or newer (excluding GTX 1650). Intel is recommended over Nvidia. AMD and Apple Silicon are not recommended.
+Some component recommendations do not change regardless of the hardware configuration:
 
-:::note Intel "Atom" CPUs
+- Storage: 100GB SSD for Jellyfin files and transcoding cache. (Consider adding more on Windows 11)
```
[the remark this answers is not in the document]
felix920506: I put in 100GB to account for the need for OS files etc. I'll add in the rule of thumb part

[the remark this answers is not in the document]
JPVenson: ok yea thats fair

docs/general/administration/hardware-selection.md
felix920506: I put in 20mb because that is a reasonable number to handle a single (maybe 2) 4k stream or a couple of 1080p streams. I will add in the bandwidth limit recommendation.

JPVenson (approved):

PR #973, opened 2024-04-29
doc: update MPEG-4 decoder support

sj14:
MPEG-4 playback is supported when using the libVLC player on Android TV


sj14:
That's true, on the other side, I was happy to see that it worked. Maybe we should just add to the footnote that LibVLC is experimental?

nielsvanvelzen:
The LibVLC implementation will actually be removed in a future release as part of a big rewrite for all player related code. Eventually codec support will be automatically detected making this table quite obsolete.

sj14:
Ah, ok. Thanks for the information. I'm closing the PR.

PR #975, opened 2024-05-02
Clarify Linux RKMPP docker instructions

felix920506:
Clarify that the normal Linux instructions are also needed when installing via Docker

Fixes https://github.com/jellyfin/jellyfin/issues/11475

### DIFF
--- a/docs/general/administration/hardware-acceleration/rockchip.md
+++ b/docs/general/administration/hardware-acceleration/rockchip.md
@@ -179,7 +179,7 @@ Root permission is required.
 
 ### Configure With Linux Virtualization
 
-Before proceeding, you should complete the [Configure on Linux Host](#configure-on-linux-host) section above.
+Before proceeding, please complete the [Configure on Linux Host](#configure-on-linux-host) section above.
 
 #### Official Docker
 
--- a/docs/general/administration/hardware-acceleration/rockchip.md
+++ b/docs/general/administration/hardware-acceleration/rockchip.md
@@ -134,8 +134,6 @@ Root permission is required.
 
 4. Add the `jellyfin` user to the `render` and `video` group, then restart the `jellyfin` service:
 
-   You can skip this step if you intend to run Jellyfin in a container (eg. Docker)
-
    :::note
 
    On some distros, the group may be `input`.
@@ -179,7 +177,7 @@ Root permission is required.
 
 ### Configure With Linux Virtualization
 
-Before proceeding, please complete the [Configure on Linux Host](#configure-on-linux-host) section above.
+Before proceeding, please complete **steps 2 and 5** in the [Configure on Linux Host](#configure-on-linux-host) section above.
 
 #### Official Docker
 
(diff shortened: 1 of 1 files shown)


docs/general/administration/hardware-acceleration/rockchip.md
```diff
@@ -177,6 +179,8 @@ Root permission is required.
 
 ### Configure With Linux Virtualization
 
+Before proceeding, please complete the [Configure on Linux Host](#configure-on-linux-host) section above.
```
gnattu, line 182: I think link to the OpenCL runtime section would be enough

gnattu: More precisely:

> Install the ARM Mali OpenCL runtime (libmali) on host:

felix920506: Don't they also need steps 2 and 3? Those look like required steps to me as they are verifying things like kernel version and configuring permissions for applications to use the GPU.

gnattu: For RK3588 the earliest BSP supports that SoC is the 5.10 LTS.

The user uid used in this section is not necessarily the "jellyfin" created in that section, and users will still have permission errors if they just copied what is provided. This is an advanced-user only document so we don't have to list everything in detail IMO. 

What's more, a lot of vendor-provided kernels enabled the rockchip hack to let every user access the rockchip hardware transcoding devices and makes even the permission configuration unnecessary. 

You can add on sentence like if you are facing permission issues, please check if the user used in the docker container has required permissions described in the previous section, and provide the link.

nyanmisaka: The `--privileged` docker flag already avoids permission issues. If the user doesn't want to use this, they can still use `getent group video | cut -d: -f3` and `--group-add="122" \ # Change this to match your "video" host group id and remove this comment`.

nyanmisaka:
This DEB package contains both Mali firmware and userspace driver. For official Docker images that have pre-installed DEB packages, users only need to install the DEB package on the host for firmware. So just link to `Configure On Linux Host` in this step.

https://github.com/jellyfin/jellyfin.org/blob/master/docs/general/administration/hardware-acceleration/rockchip.md?plain=1#L212

docs/general/administration/hardware-acceleration/rockchip.md
BergLucas: Hi,

I just wanted to say that I still had permission issues with the `--privileged` docker flag set. From what I read in the [docker documentation](https://docs.docker.com/reference/cli/docker/container/run/#privileged), the `--privileged` docker flag just allows to expose more files from the host to the container and gives more access to the container. However, the permissions of the exposed files from the host still apply so if the container user doesn't have the right UID or GID, he won't be able to access the exposed files.

In the end, I decided to follow most of the steps from the section `Configure On Linux Host` in order to create a jellyfin user and give the right permissions to the devices. Then, in docker, I just manually specified the devices, the UID and GID of the jellyfin user and the docker flag `--group-add="video"` and everything worked. I did not use the `--privileged` docker flag because I think that there's no point at both specifying the devices manually and using this flag, and it is safer not to use it.

felix920506:
I think it would be a good idea to tell users to install that anyway so that we get less complaints from 3rd party containers

nyanmisaka (approved):

PR #1011, opened 2024-05-28
Update caddy.md

DevilsDesigns:
added videos links for tutorials

### DIFF
--- a/docs/general/networking/caddy.md
+++ b/docs/general/networking/caddy.md
@@ -98,3 +98,13 @@ reverse_proxy /jellyfin/* 127.0.0.1:8096
 
 With that config, Caddy will only proxy requests that start with `/jellyfin/`.
 Note the trailing slash - that is optional, but recommended.
+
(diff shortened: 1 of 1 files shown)


felix920506:
Please provide a more descriptive title.

DevilsDesigns:
Can you give me some examples of what you would like. Or would it be better to add a description about what these tutorials go over?

nielsvanvelzen (commented):
I don't think we need to link any video tutorials. This page is meant to explain the usage of Caddy with Jellyfin, which it does. To get familiar with Caddy itself we already link to the official "getting started" documentation from Caddy.

DevilsDesigns:
> I don't think we need to link any video tutorials. This page is meant to explain the usage of Caddy with Jellyfin, which it does. To get familiar with Caddy itself we already link to the official "getting started" documentation from Caddy.

so should i not pursue continueing to edit this doc?

PR #1040, opened 2024-07-01
Changing default reverse proxy config to work locally

NeutroGe:
Hello, i think most users run jellyfin on the same local server as they run their reverse nginx proxy, so it seems logical that the nginx config files are set this way by default instead of using the dns resolve mechanism as it currently does, so i'm proposing this small change to reflect this. It should allow people to get their reverse proxy working more easily and less troubleshooting questions about it.


felix920506:
Please provide a title more descriptive of the changes made.

NeutroGe:
> Please provide a title more descriptive of the changes made.



> Please provide a title more descriptive of the changes made.

done!

docs/general/networking/nginx.md
```diff
@@ -47,10 +47,10 @@ server {
     # ssl_protocols TLSv1.3 TLSv1.2;
 
     # use a variable to store the upstream proxy
-    # in this example we are using a hostname which is resolved via DNS
-    # (if you aren't using DNS remove the resolver line and change the variable to point to an IP address e.g `set $jellyfin 127.0.0.1`)
-    set $jellyfin jellyfin;
-    resolver 127.0.0.1 valid=30s;
+    set $jellyfin 127.0.0.1;
```
felix920506, line 50: Please add a comment to guide users to change the IP address to something that isn't localhost, such as "replace 127.0.0.1 with another IP if jellyfin isn't on the same host"
There are many setups that have Jellyfin and the reverse proxy on different machines.

docs/general/networking/nginx.md
```diff
@@ -144,10 +144,10 @@ server {
     #server_name jellyfin.local;
 
     # use a variable to store the upstream proxy
-    # in this example we are using a hostname which is resolved via DNS
-    # (if you aren't using DNS remove the resolver line and change the variable to point to an IP address e.g `set $jellyfin 127.0.0.1`)
-    set $jellyfin jellyfin;
-    resolver 127.0.0.1 valid=30s;
+    set $jellyfin 127.0.0.1;
```
felix920506, line 147: Same as above

felix920506 (changes requested):

joshuaboniface (approved):
I like the change and agree with the idea but second @felix920506's comment about suggesting people change this to something else if needed.

docs/general/networking/nginx.md
NeutroGe: It was what the default config file was suggesting if not using a reverse proxy on a different machine, i didn't change anything there. I already added a comment paragraph to clarify for the case when it's not hosted on the same server.
What you are saying is another change of the base config file, if you want it i suggest you make your own merge request instead of polluting mine.

docs/general/networking/nginx.md
NeutroGe: Same as above

docs/general/networking/nginx.md
joshuaboniface: But this is a functional change: previously, it would set `$jellyfin` to `jellyfin`, i.e. a hostname; now it sets it to `127.0.0.1`. The requested change is to add another comment saying `if Jellyfin is on another machine, change 127.0.0.1 to that machine's IP`.

NeutroGe: dude, you're the owner, it was just a suggestion, do what you want, i created this 2 months ago, it's just a small change anyway, i'm deleting this it's too much trouble for such a small thing.

NeutroGe:
So much nonsense for such a small change, last time i try to help here

docs/general/networking/nginx.md
joshuaboniface: Well, you're asking to submit a change request. We're telling you to make enhancements to that change you're suggesting. If you're not interested in making your change compliant with what we want, then fine, it can stay how it is. Sorry a bit of feedback is too much for you.

PR #1043, opened 2024-07-07
Update server downloads page

nielsvanvelzen:
The flatpak change came up in chat a few times

- Remove server flatpak as it is outdated (still stuck on 10.8 with no maintainer)
- Mark Arch Linux as OS Package - we don't maintain it ourselves but it is in the official arch repo
- Mark fedora as community package - rpmfusion is not a default repo on fedora
- Move Gentoo above fedora for consistent ordering (order: official - os package - community)

VSCode also decided to reformat the file because of the `"editor.formatOnSave": true` setting. I'm not gonna fight that so the diff is messy.

### DIFF
--- a/src/data/downloads.tsx
+++ b/src/data/downloads.tsx
@@ -298,7 +298,7 @@ sudo apt install jellyfin`}
     id: 'arch',
     name: 'Arch Linux',
     osTypes: [OsType.Linux],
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
image

Bond-009 (approved):

felix920506 (approved):

PR #1064, opened 2024-08-02
Revert "Add docs for stereo downmix options"

nielsvanvelzen:
Reverts jellyfin/jellyfin.org#1063

We should not publish 10.10 documentation before release.


Shadowghost (approved):
My bad, thanks for reverting.

PR #1068, opened 2024-08-05
remove outdated note about image extractor

RayBB:
There were two parts to the note:
1. The image extractor can't be [turned off](https://github.com/jellyfin/jellyfin/issues/2355) in Jellyfin at the moment 
2. which is causing [performance issues](https://github.com/jellyfin/jellyfin/issues/2600).

The first one, has since been renamed and it [says that the extractor can be turned off](https://github.com/jellyfin/jellyfin/issues/2355#issuecomment-853934544). The second one is closed.

Overall, I think it doesn't add anything to the docs to link to these issues here because there is already a great note warning to turn off image extraction. 

Edit: Also removed note to animostiy22's repo as the github account is 404 now.

Thanks to the maintainers for all your hard work!


nielsvanvelzen (approved):

nielsvanvelzen:
Thanks!

PR #1077, opened 2024-08-14
Add dedicated dotnet tab for dotnet portable build

gnattu:
We are listing these portable builds under all OS tabs, which may confuse users because each OS also has its own platform-specific builds that we refer to as “portable” in our documentation. To avoid confusion, separate the tabs accordingly.

This also updated the description as the .NET Core runtime is now just .NET runtime

### DIFF
--- a/src/data/downloads.tsx
+++ b/src/data/downloads.tsx
@@ -16,7 +16,8 @@ export enum OsType {
   Docker,
   Linux,
   MacOS,
-  Windows
+  Windows,
+  DotNet,
 }
 
 export type Button = {
@@ -275,11 +276,11 @@ sudo apt install jellyfin`}
   {
     id: 'portable',
     name: 'Portable',
-    osTypes: [OsType.Linux, OsType.MacOS, OsType.Windows],
+    osTypes: [OsType.DotNet],
     status: DownloadStatus.Official,
     features: [],
     platforms: [Platform.DotNet],
-    description: 'The portable version can be run on any system with a .NET Core runtime.',
+    description: 'The portable version can be run on any system with a .NET runtime.',
     stableButtons: [
       {
         id: 'portable-manual-stable-link',
--- a/src/pages/downloads/server.tsx
+++ b/src/pages/downloads/server.tsx
@@ -62,6 +62,12 @@ export default function DownloadsPage({ osType = OsType.Linux }: { osType?: OsTy
                 >
                   macOS
                 </Link>
+                <Link
+                  to='/downloads/dotnet'
+                  className={clsx('pills__item', { 'pills__item--active': osType === OsType.DotNet })}
+                >
+                  DotNet
+                </Link>
               </div>
             </div>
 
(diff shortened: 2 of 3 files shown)


src/data/downloads.tsx
```diff
@@ -275,11 +276,11 @@ sudo apt install jellyfin`}
   {
     id: 'portable',
     name: 'Portable',
-    osTypes: [OsType.Linux, OsType.MacOS, OsType.Windows],
+    osTypes: [OsType.DotNet],
     status: DownloadStatus.Official,
     features: [],
     platforms: [Platform.DotNet],
-    description: 'The portable version can be run on any system with a .NET Core runtime.',
+    description: 'The portable version can be run on any system with a .NET runtime.',
```
joshuaboniface, line 283: Good idea, though I think it should be even more explicit:

+description: 'The .NET portable version can be run on any system with a .NET runtime using `dotnet jellyfin.dll`.',

joshuaboniface (approved):

src/data/downloads.tsx
gnattu: Well it seems like we cannot use markdown in this description, I change this to 'The .NET portable version can be run on any system with a .NET runtime using jellyfin.dll', which also explicitly states the user has to use the dll directly for this version.

JPVenson (approved):

src/data/downloads.tsx
```diff
@@ -16,7 +16,8 @@ export enum OsType {
   Docker,
   Linux,
   MacOS,
-  Windows
+  Windows,
+  DotNet,
```
felix920506, line 20:
```suggestion
  .NET,
```
dotnet branding guidelines for reference
https://github.com/dotnet/brand/blob/main/dotnet-brand-guidelines.pdf

gnattu: Internal enum should not care about branding that much as that is just like source code variable name

gnattu: Btw I don’t think you can use the dot in enum name at all

felix920506: Sorry wrong place, it should be below

src/pages/downloads/server.tsx
```diff
@@ -62,6 +62,12 @@ export default function DownloadsPage({ osType = OsType.Linux }: { osType?: OsTy
                 >
                   macOS
                 </Link>
+                <Link
+                  to='/downloads/dotnet'
+                  className={clsx('pills__item', { 'pills__item--active': osType === OsType.DotNet })}
+                >
+                  DotNet
```
felix920506, line 69:
```suggestion
  .NET Core
```

gnattu: It should be .NET as .NET Core no longer exists after 6.0

felix920506 (approved):

thornbill:
@joshuaboniface might like to review this. He had a lot of input on the current options listed.

PR #1053, opened 2024-07-20
Add dedicated page for hwa issues

felix920506:
the list of known issues and limitations can make the page hard to read if it gets too long.
This PR moves those things into a dedicated page.

PR #1084, opened 2024-08-23
Fix broken link

felix920506:
Fix a broken link in hardware selection guide introduced in https://github.com/jellyfin/jellyfin.org/pull/1053


maru801 (commented):
Looks good 👍

joshuaboniface (approved):

PR #1113, opened 2024-09-10
refactor(web-config): remove mention to stable config as config.json source

ferferga:
This is out of date already and with the faster cadence of releases we plan to have, this can easily become a burden to maintain.

@thornbill Does web break if the json file contains keys that it's not aware of (that's the only incompatibility I can see, some new setting introduced in an update)? Or it's able to just pick the ones that it expects correctly?

Marking as a draft until we have OK from Bill in that regard

### DIFF
--- a/docs/general/clients/web-config.md
+++ b/docs/general/clients/web-config.md
@@ -9,7 +9,7 @@ title: Jellyfin Web Configuration
 
 The Jellyfin Web default interface can be configured using the `config.json` file in the webroot. Where this is and how to edit it depends on the installation method.
 
-We recommend obtaining the [stable](https://github.com/jellyfin/jellyfin-web/blob/release-10.8.z/src/config.json) or the [unstable](https://github.com/jellyfin/jellyfin-web/blob/master/src/config.json) default version of the file to pre-populate your configuration directory before starting Jellyfin for the first time; unlike most other components of this directory, it will not be created automatically.
+We recommend obtaining the [unstable](https://github.com/jellyfin/jellyfin-web/blob/master/src/config.json) default version of the file to pre-populate your configuration directory before starting Jellyfin for the first time; unlike most other components of this directory, it will not be created automatically.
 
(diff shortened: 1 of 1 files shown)


thornbill:
Changes to plugins will likely be problematic (cause failed imports).

ferferga:
@thornbill Any chance the wrong imports can gracefully fail so it doesn't matter or that's a difficult thing to tackle right now?

thornbill:
I did some testing and failing to load a plugin should not cause issues, but any change to the path of existing plugins would still cause breakages of course.

BotBlake:
This change has been merged by https://github.com/jellyfin/jellyfin.org/pull/1113

felix920506:
@BotBlake what are you talking about? This thread IS #1113

ferferga:
@BotBlake ping

BotBlake:
@ferferga I copied the wrong PR link, but this is already in master. PR was approved and merged by josh.

BotBlake:
Was merged into master with https://github.com/jellyfin/jellyfin.org/pull/1360
(This time its the correct PR)

ferferga:
@BotBlake I'd argue my PR is not exactly the same thing. What I wanted to achieve here was making the distinction between both version-agnostic, so we don't need to come over here and update it for every release (it's something easy to forget and we indeed did multiple times).

Regardless, I'm not sure which is the best course of action for this anyway, given Bill told us that there could be some breakage. 

Any input regarding this is appreciated, updating the version number is just a stopgap at the end of the day.

BotBlake:
@ferferga You are right. Sorry, I just saw what it updated and came to an conclusion based on that!

Maybe it is possible to add a dynamic link somehow.
Just how https://github.com/jellyfin/jellyfin/releases/latest automatically gets redirected to the latest release.

BotBlake:
Good news: 
This should be solve able by 
https://github.com/jellyfin/jellyfin.org/pull/1596

PR #1065, opened 2024-08-02
Re-apply downmix docs

gnattu:
This reverts commit dff5fa4fc2d1f6f8cd74325e1ffcada68a49a8c0.

Applies #1063 again.

Do not merge until 10.10

PR #1120, opened 2024-09-14
10.10 hwa / transcode docs update

felix920506:


docs/general/administration/hardware-selection.md
```diff
@@ -64,7 +71,7 @@ These hardware platforms might lead to a poor Jellyfin experience. Please be car
...
 - Prebuilt NAS devices: The software environment on most Prebuilt NAS devices often causes 3rd party software to be hard to install and not work properly. They also often have low end processors that are too slow for a good Jellyfin experience (Intel Atom, Realtek ARM CPUs etc.).
-- Most SBCs (Single Board Computers): Most SBCs (Including Raspberry Pis and **especially the Pi 5**) are too slow to provide a good Jellyfin experience. If You really want to run Jellyfin on a SBC, please look at models based on the following platforms: Rockchip RK3588 / RK3588S, Intel Core, Intel 12th gen N series
+- Most SBCs (Single Board Computers): Most SBCs (Including Raspberry Pis and **especially the Pi 5**) are too slow to provide a good Jellyfin experience since they often lack proper support for hardware acceleration. If You really want to run Jellyfin on a SBC, please look at models based on the following platforms: Rockchip RK3588 / RK3588S, Intel Core, Intel 12th gen N series
```
Gauvino, line 74:
```suggestion
- Most SBCs (Single Board Computers): Most SBCs (Including Raspberry Pis and **especially the Pi 5**) are too slow to provide a good Jellyfin experience since they often lack proper support for hardware acceleration. If You really want to run Jellyfin on an SBC, please look at models based on the following platforms: Rockchip RK3588 / RK3588S, Intel Core, Intel 12th gen N series
```

Gauvino (changes requested):

nyanmisaka:
Maybe clean up the WIP software tonemap in the hardware transcoding page too? It's outdated and in the wrong place.

https://github.com/jellyfin/jellyfin.org/blob/a2cb03c5c9533c2c58e78aa4383b9286933bacd0/docs/general/administration/hardware-acceleration/index.md?plain=1#L183-L185

docs/general/administration/hardware-selection.md
```diff
@@ -50,6 +50,13 @@ If you are planning to use a dedicated graphics card (including upgrading an old
 
 Intel Drivers are much easier to install on Linux, with many distributions including them by default. If you do not need CUDA for other applications, it is highly recommended that you stick with Intel Graphics on Linux.
 
+### Servers without GPUs
+
+Not having a GPU is **NOT** recommended for Jellyfin. The following specs are provided for reference only, for software tone-mapping 4K content to SDR:
+
+- CPU: AMD Ryzen 9 5950X, Intel Core i7-12700K or better
```
gnattu, line 57: This only applies to Dolby Vision, videos does not require dolby vision reshaping do not need 16core zen3, and the 16core requirement is 60fps playback, so of the goal is 24fps the requirement can also be lower. Also, if users are having 12700K they can just use the integrated graphics.

The requires CPU power differs vastly for use cases and probably should not be documented like this because 4K60 DV tone mapping + transcoding is probably the most demanding task.

docs/general/administration/hardware-selection.md
```diff
@@ -50,6 +50,21 @@ If you are planning to use a dedicated graphics card (including upgrading an old
...
+<details>
+- RAM: 16GB
+
+For software tone-mapping 4K 24fps Dolby Vision content to SDR:
+CPU: AMD Ryzen 7 5700X, Intel Core i5-12400F or better
+
+For 4K 60fps Dolby Vision content:
+CPU: AMD Ryzen 9 5950X, Intel Core i7-12700F or better
+
+</details>
```
gnattu, line 66: I'd say this is still inaccurate. 12700F has no way to guarantee 4K60 dolby vision and choosing CPU solely by dolby vision content performance is still very biased because as I said earlier it is probably the worst thing we can have. Dolby Vision videos has no fallback (like profile 5 videos) will have a significant impact on performance to a point that you will need 8 core zen3 for 4K24 playback while normal HDR10 videos or dolby vision videos with fallback (like profile 8.1) can be tone mapped with only 4 core Zen3 and still get 4K24 playback. These numbers are encoding with libx264 veryfast, using libx265 or libsvtav1 may reduce the framerate even further, so don't try to make a "recommended spec" because we cannot as it is way too scenario specific. 

We also need to be clear about the following points why CPU only is not recommended:

- Transcoding videos with CPU is performance demanding, and it is hard to gurantee all clients are compatible with all media sources
- Advanced codecs like HEVC and AV1 is even slower
- HDR to SDR tone mapping can make things worse, depending on the inputs you may end up with need 16 core zen3 level performance, or even no desktop CPUs currently can handle(if you want to use libx265 for dolby vision 4K60)

So if we have to do hardware selection guide on CPUs, we just say as much CPU cores as possible because the range is just too wide. The CPU only needs raspberry pi level performance if no video transcoding is needed, and a 4-6 core zen 3 can handle most use cases, including HDR10 4k24 tone mapping with libx264 encoder, but the CPU requirement goes up significantly when libx265/libsvtav1 is used and/or dolby vision reshaping is required.

gnattu (changes requested):
I'd say if we cannot reasonably make spec recommendation then we don't, just list the performance requirement for typical workload and let the user to choose CPUs themselves. Encoding presets are tunable and it can go beyond even the fastest desktop CPU's capability that will eventually require HEDT/Server CPUs and it will be very ridiculous to recommend those models in this guide.

docs/general/administration/hardware-selection.md
```diff
@@ -97,6 +101,10 @@ In BIOS settings of motherboards, it may also be called `Smart Access Memory` or
 
 When using Intel ARC Graphics, Resizable BAR is recommended. Disabling it will result in a 10% reduction in transcoding performance. However, given how fast the media engine on Intel ARC is, this will not be a problem for the vast majority of users.
 
+#### Software HDR to SDR Tone-mapping
+
+If there is no GPU available, the CPU can also be used to tone-map HDR content to SDR. **This is very demanding on hardware, therefore a GPU is ALWAYS recommended.** When using software tone-mapping, a software encoder will always be used. Tone-mapping 4K 60fps Dolby Vision content to SDR H264 requires a Ryzen 9 5950X for faster than real time transcoding. If you would like to output in H265 or AV1, you may end up in situations where no current desktop CPUs have enough performance.
```
gnattu, line 106:
```suggestion
If there is no GPU available, the CPU can also be used to tone-map HDR content to SDR. **This can be very demanding on CPU, therefore a GPU is ALWAYS recommended.** When using software tone-mapping, a software encoder will always be used. Tone-mapping 4K 60fps Dolby Vision content to SDR H264 requires a Ryzen 9 5950X for faster than real time transcoding. If you would like to output in H265 or AV1, you may end up in situations where no current desktop CPUs have enough performance.
```

gnattu: Ideally this should be moved to a dedicated document about tone mapping but this is good enough for now.

gnattu (approved):

docs/general/administration/hardware-acceleration/index.md
```diff
@@ -180,13 +180,13 @@ Dolby Vision (P5 & P8) to SDR tone-mapping is supported in Jellyfin 10.8 and req
...
 :::
 
-| OS/Platform    | NVIDIA NVENC | AMD AMF | Intel QSV | Intel VA-API | AMD VA-API |Apple VideoToolbox| Rockchip RKMPP | Software |
-| -------------- | ------------ | ------- | --------- | ------------ | ---------- | ---------------- | -------------- | -------- |
-| Windows        | ✔️           | ✔️      | ✔️        | N/A          | N/A        | N/A              | N/A           | WIP      |
-| Windows Docker | ✔️           | N/A     | N/A       | N/A          | N/A        | N/A              | N/A            | WIP      |
-| Linux          | ✔️           | ✔️      | ✔️        | ✔️           | ✔️        | N/A              | ✔️             | WIP      |
-| Linux Docker   | ✔️           | ✔️      | ✔️        | ✔️           | ✔️        | N/A              | ✔️             | WIP      |
-| macOS          | N/A          | N/A     | N/A       | N/A          | N/A        | ✔️               | N/A            | WIP      |
+| OS/Platform    | NVIDIA NVENC | AMD AMF | Intel QSV | Intel VA-API | AMD VA-API | Apple VideoToolbox | Rockchip RKMPP | Software |
```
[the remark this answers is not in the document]
gnattu: After looking at this again I think we can remove this table altogether? It does not make any sense to have a table like this now because every platform supports tone mapping on every possible hardware acceleration methods for that platform now and this table does not provide extra information.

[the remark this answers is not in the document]
felix920506: Sounds fair

docs/general/administration/hardware-acceleration/amd.md
```diff
@@ -435,7 +429,7 @@ Root permission is required.
          image: jellyfin/jellyfin
          user: 1000:1000
          group_add:
-           - "122" # Change this to match your "render" host group id and remove this comment
+           - '122' # Change this to match your "render" host group id and remove this comment
```
gnattu, line 432: I'm not sure about all the double quote to single quote changes in yaml blocks because they do have different meanings. Although the changes looks okay and should still work to me but I'm not sure if those changes are necessary

docs/general/server/transcoding.md
```diff
@@ -26,3 +23,7 @@ There are four types of playback; three of which involve transcoding. The type b
 - Remux: Changes the container but leaves both audio and video streams untouched.
 - Direct Stream: Transcodes audio but leaves original video untouched.
 - Transcode: Transcodes the video stream.
+
+## HDR to SDR Tone-mapping
+
+When the source video is in HDR, it will need to be tone-mapped to SDR when transcoding, as Jellyfin currently doesn't support HDR to HDR tone-mapping, or passing through HDR metadata. While this can be done in software, it is very slow and you may encounter situations where no modern consumer CPUs can transcode in real time. Therefore, a GPU is always recommended, where even a basic Intel iGPU can handle the load without problems.
```
gnattu, line 29: Not that ideal. We've received reports that N100 GPU is struggling on 4K60 dolby vision videos, but it is as fast as 8 core zen3 at dolby vision reshaping so still much more cost effective

docs/general/server/transcoding.md
```diff
@@ -9,10 +9,7 @@ These settings will relate to backend options that modify how the server transco
 
 ## Hardware Acceleration
 
-If your hardware supports this you can enable [hardware acceleration](/docs/general/administration/hardware-acceleration) for much faster transcoding. Some of the supported methods are listed below.
-
-- VAAPI
-- NVENC
+If your hardware supports this you can enable [hardware acceleration](/docs/general/administration/hardware-acceleration) for much faster transcoding. Some of the supported methods are listed below. Jellyfin supports hardware accelerated transcoding on all modern GPUs from Intel, AMD, Nvidia, Apple, Rockchip and Raspberry Pi if the required hardware is available.
```
gnattu, line 12:
```suggestion
If your hardware supports this you can enable [hardware acceleration](/docs/general/administration/hardware-acceleration) for much faster transcoding. Jellyfin supports hardware accelerated transcoding on all modern GPUs from Intel, AMD, Nvidia, Apple, Rockchip and Raspberry Pi if the required hardware is available.
```

docs/general/administration/hardware-acceleration/amd.md
```diff
@@ -15,7 +15,7 @@ On Windows **AMF** is the only available method.
 
 On Linux there are two methods:
 
-- **VA-API** - Prefered on all GPUs, full acceleration on Vega+ GPUs, open source.
+- **VA-API** - Prefered on all GPUs, full acceleration on Polaris(RX400/500)+ GPUs, open source.
```
Shadowghost, line 18:
```suggestion
- **VA-API** - Preferred on all GPUs, full acceleration on Polaris(RX400/500)+ GPUs, open source.
```

docs/general/administration/hardware-acceleration/index.md
```diff
@@ -63,15 +63,15 @@ The transcoding pipeline usually has multiple stages, which can be simplified to
...
+Partial acceleration may result in higher CPU usage and lower transcoding speed.
 
 :::
 
-Jellyfin 10.8 supports full acceleration on mainstream Intel, NVIDIA and AMD (Windows only) GPUs.
-
-Jellyfin 10.9 enables full acceleration for:
+Jellyfin supports full acceleration for:
 
+- Mainstream Intel, Nvidia GPUs on Windows and Linux
```
Shadowghost, line 72:
```suggestion
- Mainstream Intel and Nvidia GPUs on Windows and Linux
```

docs/general/administration/hardware-acceleration/index.md
```diff
@@ -180,21 +180,15 @@ Dolby Vision (P5 & P8) to SDR tone-mapping is supported in Jellyfin 10.8 and req
...
 - Avoid H.264 / AVC 10-bit videos
 
   :::tip
 
-  The hardware decoding of H.264 / AVC 10-bit (High 10 profile) video is not supported by any Intel, NVIDIA and AMD GPU. Jellyfin will always fallback to software decoding for it. Consider upgrading such video to H.265 / HEVC 10-bit (Main 10 profile).
+  The hardware decoding of H.264 / AVC 10-bit (High 10 profile) video is not supported by any Intel, NVIDIA and AMD GPU.
+- It is only supported by Apple Silicon and Rockchip. Jellyfin will fall back to software decoding for it when there is
+- no hardware decoder available. Consider upgrading such video to H.265 / HEVC 10-bit (Main 10 profile).
 
   :::
```
Shadowghost, line 193: This is not properly rendered. The second point is also a continuation of the sentence of the first

docs/general/administration/hardware-selection.md
```diff
@@ -50,11 +50,15 @@ If you are planning to use a dedicated graphics card (including upgrading an old
 
 Intel Drivers are much easier to install on Linux, with many distributions including them by default. If you do not need CUDA for other applications, it is highly recommended that you stick with Intel Graphics on Linux.
 
+### Servers without GPUs
+
+Not having a GPU is **NOT** recommended for Jellyfin, as video transcoding on the CPU is very performance demanding. HDR to SDR tone-mapping can make the situation even worse. Depending on your configuration, you may end up in situations where a Ryzen 9 5950X cannot handle even a single video stream. Please read the detailed section below.
```
Shadowghost, line 55:
```suggestion
Not having a GPU is **NOT** recommended for Jellyfin, as video transcoding on the CPU is very performance demanding. HDR to SDR tone-mapping can make the situation even worse. Depending on your configuration, you may end up in situations where a Ryzen 9 5950X cannot handle even a single video stream. Please read the section below for more details.
```

Maybe link to the section directly.

docs/general/administration/hardware-selection.md
```diff
@@ -64,7 +68,7 @@ These hardware platforms might lead to a poor Jellyfin experience. Please be car
...
 - Prebuilt NAS devices: The software environment on most Prebuilt NAS devices often causes 3rd party software to be hard to install and not work properly. They also often have low end processors that are too slow for a good Jellyfin experience (Intel Atom, Realtek ARM CPUs etc.).
-- Most SBCs (Single Board Computers): Most SBCs (Including Raspberry Pis and **especially the Pi 5**) are too slow to provide a good Jellyfin experience. If You really want to run Jellyfin on a SBC, please look at models based on the following platforms: Rockchip RK3588 / RK3588S, Intel Core, Intel 12th gen N series
+- Most SBCs (Single Board Computers): Most SBCs (Including Raspberry Pis and **especially the Pi 5**) are too slow to provide a good Jellyfin experience since they often lack proper support for hardware acceleration. If You really want to run Jellyfin on an SBC, please look at models based on the following platforms: Rockchip RK3588 / RK3588S, Intel Core, Intel 12th gen N series
```
Shadowghost, line 71: Should probably not use the same phrase twice in one line

felix920506: How should I change this?

gnattu:
This is going to conflict with #1065 due to the transcoding page change. Can we have something like a 10.10 branch so that we can merge all future docs in to resolve conflicts there?

nyanmisaka:
There are some `jellyfin-ffmpeg6` package names that need to be updated to `jellyfin-ffmpeg7`, is there a better way to avoid this in the future?

docs/general/administration/hardware-selection.md
solidsnake1298: Change ```Most SBCs (Single Board Computers):``` to ```Single Board Computers (SBC):```.  It avoids what Ghost mentioned, using the same phrase twice in close proximity, and is more clear since it uses the full phrase before using the abbreviation.

felix920506:
@nyanmisaka you will have to talk to @joshuaboniface for that

felix920506:
@nyanmisaka I updated version numbers, please review

PR #1129, opened 2024-09-24
Adding host fonts for docker containers

solidsnake1298:
Adding documentation for adding host system fonts to a Docker container.

### DIFF
--- a/docs/general/administration/configuration.md
+++ b/docs/general/administration/configuration.md
@@ -119,7 +119,7 @@ The system fonts installed on the server are used for burning in subtitles and r
 
 If you are running Jellyfin in Docker, it is possible to pass fonts you have on the host system into the Jellyfin container.  Simply add two additional volumes for /usr/share/fonts and /usr/share/fontconfig and map them to the same path inside the container.
 
-**Docker Compose Example**
+Docker Compose Example
 
 ```yml
     volumes:
@@ -127,7 +127,7 @@ If you are running Jellyfin in Docker, it is possible to pass fonts you have on 
       - /usr/share/fontconfig:/usr/share/fontconfig
 ```
 
-**Docker Command Line Example**
+Docker Command Line Example
 
 ```sh
  --volume /usr/share/fonts:/usr/share/fonts
--- a/docs/general/administration/configuration.md
+++ b/docs/general/administration/configuration.md
@@ -115,7 +115,7 @@ Jellyfin uses fonts to render text in many places.
 
 The system fonts installed on the server are used for burning in subtitles and rendering cover images. How to install them depends on the operating system.
 
-### Server Fonts in Docker
+### Server Side System Fonts in Docker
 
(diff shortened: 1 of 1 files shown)


docs/general/administration/configuration.md
```diff
@@ -115,6 +115,25 @@ Jellyfin uses fonts to render text in many places.
 
 The system fonts installed on the server are used for burning in subtitles and rendering cover images. How to install them depends on the operating system.
 
+### Server Fonts in Docker
```
felix920506, line 118: We should specify that it is server side system fonts for docker, since server fonts could also mean "fallback fonts"

felix920506 (changes requested):

p0358:
Hey, can you take a look at my PR #1136? I wasn't aware of yours when creating mine, but now they kinda conflict with each other. I believe that my method of just mounting a folder of custom fonts into a subdirectory of `/usr/local/share/fonts` is pretty nice and useful, since it can allow users to easily add new fonts for subtitle burn-in on top of what's already existing in the container. Meanwhile mounting a host path instead has a few potential cons instead. My use-case comes from discussion at jellyfin/jellyfin#12511, where user might want to specify some path for fallback fonts that'd be streamed to web clients, and the same path could also be easily mounted into the Docker container, to also let the same fonts be used during burn-in with clients that do that instead of rendering them themselves.

So let me know what you think about it. In worst case scenario we could just end up mentioning both possibilities, but I believe just mounting a custom dir of fonts might be better if we were to choose just one method, as it's easier to add new font files there, instead of having to install an APT packet of fonts on host, provided one even exists (which for some obscure fonts that translators use in subtitles I fear might usually not be the case)...

solidsnake1298:
> Hey, can you take a look at my PR #1136? I wasn't aware of yours when creating mine, but now they kinda conflict with each other. I believe that my method of just mounting a folder of custom fonts into a subdirectory of `/usr/local/share/fonts` is pretty nice and useful, since it can allow users to easily add new fonts for subtitle burn-in on top of what's already existing in the container. Meanwhile mounting a host path instead has a few potential cons instead. My use-case comes from discussion at [jellyfin/jellyfin#12511](https://github.com/jellyfin/jellyfin/pull/12511), where user might want to specify some path for fallback fonts that'd be streamed to web clients, and the same path could also be easily mounted into the Docker container, to also let the same fonts be used during burn-in with clients that do that instead of rendering them themselves.
> 
> So let me know what you think about it. In worst case scenario we could just end up mentioning both possibilities, but I believe just mounting a custom dir of fonts might be better if we were to choose just one method, as it's easier to add new font files there, instead of having to install an APT packet of fonts on host, provided one even exists (which for some obscure fonts that translators use in subtitles I fear might usually not be the case)...

@felix920506 - I think we should choose #1136 over my PR.

felix920506:
Ok then, closing in favor of #1136

PR #1148, opened 2024-10-05
Add Raspberry Pi V4L2 deprecation notice

gnattu:
I know this is frustrating, but it's necessary for now. Raspberry Pi's V4L2 acceleration never achieved full hardware acceleration, and its CPU is simply too underpowered for partial acceleration. With the latest generation of Raspberry Pi lacking a hardware encoder entirely, continuing development for hardware acceleration on this platform seems futile. 

This would still need more inputs from the team though. 

This removes Raspberry Pi V4L2 as a supported acceleration method and adds a deprecation notice for it.


nyanmisaka (approved):
I agree with this. Unlike the decode-only video in the player, and the encode-only video in the recorder. Transcoding is a very CPU intensive task. The CPU not only has to decode and encode the video, but also has to handle the video filters in between (pixel format/layout conversion, down-scale video resolution, overlay subtitles, HDR2SDR tone-mapping or even transpose the video).

So on the RPi, which lacks full hardware acceleration, any transcoding from >1080p or 10bit or HEVC/AV1 video will not meet real-time performance requirements - at least 24fps. Not to mention that audio transcoding also happens on the CPU, exacerbating the problem. Of course, if RPi 6 or later adds hardware encoder and video filter support that is capable of what we need, it is possible to reconsider bringing back V4L2 support.

Being honest with users is necessary. Plex team even suggest their paid user who just bought an RPi5 return it and switch to Intel N100 boxes for the same reason. https://forums.plex.tv/t/loading-plex-server-on-pi5-problems/891324/5

Continuing to state "support" when the hardware simply isn't capable will only lead users to have false expectations when choosing hardware, and then complain that it was a software issue when the RPi runs Jellyfin poorly. Also, IMO ideally the server should disable video transcoding by default once an under-powered CPU/GPU is detected.

joshuaboniface (approved):

gnattu:
For 10.10. Mention this in 10.10 release note and merge by then

JPVenson (approved):

joshuaboniface:
@gnattu Can you update this from a draft? I believe we're on-track for 10.10 this weekend so want this to be ready to merge.

PR #1156, opened 2024-10-12
Add Jellychord to clients list

felix920506:
Add Jellychord to clients list

### DIFF
--- a/src/data/clients.ts
+++ b/src/data/clients.ts
@@ -720,6 +720,22 @@ export const Clients: Array<Client> = [
     ]
   },
   {
+    id: 'jellychord',
+    name: 'Jellychord',
+    description: 'A modern Discord music bot for Jellyfin using slash commands, written in python.',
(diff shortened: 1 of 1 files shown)


src/data/clients.ts
```diff
@@ -719,6 +719,22 @@ export const Clients: Array<Client> = [
       }
     ]
   },
+  {
```
nielsvanvelzen, line 722: New additions should be added to the bottom of the list

nielsvanvelzen (changes requested):

src/data/clients.ts
joshuaboniface: I think I agree will @thornbill's ordering.

nielsvanvelzen: Yep should be fine considering it was already discussed 👍

nielsvanvelzen (approved):

PR #1157, opened 2024-10-12
docs: fix audiobooks folder structure example for book

prahal:
A book should be under an Author, even for Audiobooks.

---
error in Server guide > Media > Books graph - Audiobooks "Books" folder should be under "Author" not "Audiobooks" parent

In https://jellyfin.org/docs/general/server/media/books, I believe
```
Books
├── Audiobooks
│   ├── Author
│   │   ├── Book1.flac
│   │   └── Book2.flac
│   └── Book
│       ├── Chapter1.flac
│       └── Chapter2.flac
└── Books
    └── Author
        ├── Book1.epub
        ├── Book2.epub
        ├── Book
        │   ├── Book1.epub
        │   ├── cover.ext
        │   └── metadata.opf
        └── Book3.mp3
```
should be
```
Books
├── Audiobooks
│   └── Author
│       ├── Book1.flac
│       ├── Book2.flac
│       └── Book
│           ├── Chapter1.flac
│           └── Chapter2.flac
└── Books
    └── Author
        ├── Book1.epub
        ├── Book2.epub
        ├── Book
        │   ├── Book1.epub
        │   ├── cover.ext
        │   └── metadata.opf
        └── Book3.mp3
``

That is in "Server guide > Media > Books" under Audiobooks, the "Books" folder should be under "Author" not "Audiobooks" parent folder.

### DIFF
--- a/docs/general/server/media/books.md
+++ b/docs/general/server/media/books.md
@@ -10,12 +10,12 @@ The most common organization scheme for books is separation by Audiobook then by
 ```txt
 Books
 ├── Audiobooks
-│   ├── Author
-│   │   ├── Book1.flac
-│   │   └── Book2.flac
-│   └── Book
-│       ├── Chapter1.flac
-│       └── Chapter2.flac
+│   └── Author
+│       ├── Book1.flac
+│       ├── Book2.flac
+│       └── Book
+│           ├── Chapter1.flac
+│           └── Chapter2.flac
 └── Books
     └── Author
         ├── Book1.epub


felix920506:
Are you sure that simply throwing a book into the Audiobooks folder won't work?

prahal:
> Are you sure that simply throwing a book into the Audiobooks folder won't work?

No. But this is not what this bug report is about.

I believe that not putting a non audiobook book would also not prevent it from showing up. Still the example for Books always tells to put a book under an author. Why provide a non recommanded example because it could still work in the example to show users how to structure the folders  (and not provide the recommended structure for audiobooks in this example to only show the non recommended one?).

The bug report was not about the example not getting the book to show up but only about a bug in the example for the recommended structure.

I cannot spend time checking if indeed the audiobook would show up even if not under an Author for at least a few days,  but it is beyond the scope of this bug. Maybe another bug report should be open about the recommended structure not displaying all working cases. This is not the aim of this bug.

I reported this one because to me there was a "typo" in the recommended structure example in the wiki.
By the way, as in the example the book folder does not have the author in its name how would the author be known without this book folder being inside its author folder?

felix920506:
Please also resolve the merge conflicts

BotBlake:
Original issue resolved by https://github.com/jellyfin/jellyfin.org/pull/1320
Additionally this is deprecated by https://github.com/jellyfin/jellyfin.org/pull/1573

Thanks to everyone for the contributions!

prahal:
Well, my opinion is that we should document only the schemes that we support "and" plan to support in the foreseeable future, not the schemes that "work" nowadays. If someone wants to add a UI with metadata support for Audiobooks, now that we have advertised all possible imaginable schemes as supported, they will have to break support for any of these schemes, thus breaking libraries that users customized following the officially supported Jellyfin schemes. 
This seems like shooting oneself in the foot to me.

Nobody asked us for most of these schemes that we already know we won't be able to support if someone steps up to add a UI... why advertise them as supported?

Is there any way we could have a UI without an author parent folder for Audiobooks?

BotBlake:
Hi @prahal 
Currently, Audiobooks are in a Folder View.
That way, you will be abled to have Author parent folders simply by adding the media in a Subfolder.

Also, Jellyfin does SUPPORT every possible configuration of folder structures for audiobooks. That is due to the fact that it literally only provides a folder view.

I have made sure to add the requested example in the current version of jellyfin.org

What is the exact change that you are proposing? 

Kind regards.

prahal:
> Hi @prahal Currently, Audiobooks are in a Folder View. That way, you will be abled to have Author parent folders simply by adding the media in a Subfolder.
> 
> Also, Jellyfin does SUPPORT every possible configuration of folder structures for audiobooks. That is due to the fact that it literally only provides a folder view.
> 
> I have made sure to add the requested example in the current version of jellyfin.org
> 


I believe we have a language barrier issue. Not that I am not a native English speaker. By "supported", I mean that we will "support" these folder structures in the future. We all known that every folder structure that you could add to the documentation for audiobook would display fine because there is no folder structure requirements actually at all. So anything would display.


> What is the exact change that you are proposing?

I am advocating for not advertising all currently viewable structure (because in fact there is actually no structure at all and everything is possible). But sticking to what a "future UI" can support.
By the way, I am also pointing the fact that there are no user request for schemes without an author parent folder. At least none that I saw. What is the point in explicitly stating any imaginable structure to be supported (I agree we have no requirement for any structure, but that does not mean that we should advertise any possible scheme we can imagine being supported. Thus is future burden you are putting on future developer working on Audiobook support. I am asking why increase the burden when no user ever asked for these structure without an Author parent folder that are not needed by any user?
When you develop software, you want to decrease future work, no increase it. Especially when there is no user request for such burden.

I am not telling that explicitly telling users that they can do every and each folder structure that could currently be shown by the UI (well in fact it does not require any structure at all, so any imaginable scheme would display) would not display in the current UI. I am telling you are adding a ton of work to future development for ... no reason, just because currently they are working.
I am telling that advertising a something like a folder structure as supported is not the same at all as the fact that this folder structure currently displaying in Jellyfin. We are signing a contract with users that we will support what we advertise, we are not telling that it currently work, but we are telling that it will work in the future. Which is wrong as if there ever is a Jellyfin UI for Audiobook then most of the scheme you documented in the official documentation will not be able to be supported.

The documentation is not there to document all the hacks that could work today, but what is meant to be supported by the project. A lot of APIs in a lot of project are not documented or documented as private. Because they are meant to be broken in the future and are thus told not to be used by users of the API and the software. They are just there because there was no time to do better for now, as there is no UI for audiobook in Jellyfin currently.

So I am advocating for the folder structure that could not be supported by any future to not be advertised as supported in the documentation.
On the other hand, I am just a sporadic contributor in the project so all I can do is give you advice from the field. Never advertise what you don't intend to support in the public interface documentation. And supported folder structure is public interface.
Doing the opposite is making Jellyfin unmaintainable in the future. I agree that in the present this will change nothing.

BotBlake:
Hi @prahal 

thank you for explaining!
I think that I do understand you:
You want to enforce a more strict folder structure in the Docs to have people create decent folder structures.
That way, if Audiobook support will ever use something other then Folder view, it would work out of the Box!

I do understand that approach!
However at the moment nobody is actively working on trying to get Audiobooks away from folder view.
Therefore this "future change" will not be happening soon.

Additionaly, since nobody is even working on it, we dont know in what ways the folder structure would be allowed.

The documentation should represent what is officially supported.
And - at the moment - we do OFFICIALLY SUPPORT folder view for audiobooks.
It is not "a hack".

I hope you can get behind this.
If you want to continue discussing about the Ideal folder structure, please open an `issue` here on the GitHub, OR join our `jellyfin-documentation` room on Matrix/ Discord/ IRC.

Kind regards and thanks for sharing your oppinion!

PR #1161, opened 2024-10-20
Add deprecation notice for Intel <11th gen

felix920506:
As discussed on matrix earlier, Intel 10th gen and older are losing support for QSV.
Updates documentation to not recommend 10th gen and older for iGPU

### DIFF
--- a/docs/general/administration/hardware-acceleration/intel.md
+++ b/docs/general/administration/hardware-acceleration/intel.md
@@ -55,7 +55,7 @@ QSV can be used together with VA-API and DXVA/D3D11VA for a more flexible hybrid
 
 :::caution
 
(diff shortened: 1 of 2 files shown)


nyanmisaka:
TBH this will confuse users who own these hardware. Losing support completely due to code aging may be a few years away, at least it doesn't cause problems yet. Users with low budgets also do not need to upgrade to 11th Gen+ very urgently. Just mention that Intel has stopped developing on these hardware and let users make their own decisions.

felix920506:
@nyanmisaka I changed the wording a bit

nyanmisaka:
It would be good to attach these two links for reference.

- https://github.com/Intel-Media-SDK/MediaSDK
- https://github.com/intel/compute-runtime/blob/master/LEGACY_PLATFORMS.md

felix920506:
@nyanmisaka linked, please re-review

nyanmisaka (approved):

PR #1163, opened 2024-10-22
Update TrueNAS Scale instructions for upcoming 24.10 release

p0358:
Here's the situation:
- TrueCharts is no longer supporting TrueNAS SCALE at all, they removed their catalog, now it's 404 and the app installations are basically unsupported
- Jellyfin is now available in the official app catalog (wasn't at the time these instructions were created)
- (upcoming TrueNAS Scale 24.10 will replace Kubernetes-based apps with Docker-based apps, but as far as UI is concerned, pretty much nothing is changed when installing apps from official catalogue)

So this PR changes the instructions to simply use the official Jellyfin app that's now available.

The wording is adjusted for the interface of upcoming release of TrueNAS Scale 24.10 that's expected soon in late October (currently in RC), but the general instructions for installing Jellyfin from official app catalog are pretty much the exact same from user perspective in old versions now too. The only practical difference is some wording changes (such as "Available Applications" is now "Discover Apps" and "Open" button is now "Web UI", nothing that would stop people from going through).

So I think it can be merged as-is without waiting. Especially given TrueCharts instructions are now broken/outdated and the new ones will work more-less for old versions just as well.


JPVenson (approved):

docs/general/installation/container.md
```diff
@@ -311,28 +311,21 @@ SuccessExitStatus=0 143
...
+2. Click `Discover Apps` and search for `Jellyfin`
 
-2. Click `Available Applications` and search for `Jellyfin`
+3. Click `Install`, which will take you to the GUI installation wizard and you'll be able to fill out the necessary info
 
-3. Click `Install`, which will take you to the GUI Wizard and you'll be able to fill out the necessary info
+   - Published Server URL to publish in UDP Auto Discovery response
+   - User and Group Configuration, Network Configuration
+   - Storage Configuraton
+     - For Jellyfin Config Storage and Jellyfin Cache Storage, consider mapping Host Paths for ease of backup
```
felix920506, line 327: A few things.
1. Please specify that this SHOULD be on SSDs and not HDDs. Putting this on HDDs WILL lead to a poor experience 
2. It is generally not advised to store databases on ZFS or other CoW file systems but this is Truenas Scale so whatever

docs/general/installation/container.md
```diff
@@ -311,28 +311,21 @@ SuccessExitStatus=0 143
...
+     - For Jellyfin Transcode Storage, consider using `Temporary` or `tmpfs`
+     - For media, add your library directories as Additional Storage
```
felix920506, line 330: Can you have instructions on how to give Jellyfin access to GPUs? Also disclaim that users might need to check kernel version on newer Intel GPUs

felix920506 (changes requested):

docs/general/installation/container.md
p0358: Good point, completely forgot about the GPUs there. They actually have option at the very bottom Resources Configuration -> GPU Configuration -> Passthrough available (non-NVIDIA) GPUs. This leaves the open question of what about NVIDIA GPUs xD

p0358: > Also disclaim that users might need to check kernel version on newer Intel GPUs

Current and upcoming versions of TrueNAS (24.10 and 24.04) seem to use 6.6, while previous old 23.10 used 6.1. I read that Intel Arcs were introduced in 6.2. So in a week it'll be last two stable versions already supporting them, with which I think it might be fine to shorten the docs by not mentioning that maybe if it's new enough?

I mean TrueNAS has a read-only main system fs, so it would be somewhat challenging to upgrade the kernel to begin with too :/ so the mention should boil down to whether 6.6 is considered new enough to not cause any problems

p0358: Transcoding mention done. For the record, this is how the screen is supposed to appear there:


I didn't go into more details on how to configure it within Jellyfin, as the instructions for Docker etc above don't go either. I don't remember if the `/dev/dri/renderD128` is configured in Jellyfin by default, but in my case having VA-API and that path in the settings works then.

docs/general/installation/container.md
p0358: 1. Done
2. Welp, there's no escaping from ZFS on TrueNAS

docs/general/installation/container.md
felix920506: Some Intel GPUs need 6.7 so a disclaimer is still needed. A short disclaimer linking the Intel known issues and limitations page under hardware acceleration, and suggest them to use a VM with pcie pass through would be sufficient

docs/general/installation/container.md
```diff
@@ -311,28 +311,29 @@ SuccessExitStatus=0 143
...
 
-3. Click `Install`, which will take you to the GUI Wizard and you'll be able to fill out the necessary info
+   - Published Server URL to publish in UDP Auto Discovery response
+   - User and Group Configuration, Network Configuration
+   - Storage Configuraton
+     - For `Jellyfin Config Storage` and `Jellyfin Cache Storage`, consider mapping Host Paths for ease of backup.
+       - Store them on SSD storage if possible, as using HDD storage **will** lead to poor experience.
+     - For `Jellyfin Transcode Storage`, consider using `Temporary` or `tmpfs`
+       - `Temporary` places a Docker volume under your configured Apps dataset (`Apps` -> `Configuration` -> `Choose Pool`). Avoid if that's on HDD.
+       - `tmpfs` creates a temporary directory on the RAM (but swap usage is permitted). Ensure you set the limit to at least a few gigabytes.
```
felix920506, line 331:
```suggestion
       - `tmpfs` creates a temporary directory on the RAM (but swap usage is permitted). Ensure you set the limit to at least a few gigabytes. You may need to increase the limit if you have many large media files, or transcodes might fail.
```

docs/general/installation/container.md
```diff
@@ -311,28 +311,29 @@ SuccessExitStatus=0 143
...
-      Branch: main
-   5. Click Save and allow SCALE to refresh its catalog with TrueCharts (this may take a few minutes)
+2. Click `Discover Apps` and search for `Jellyfin`
 
-2. Click `Available Applications` and search for `Jellyfin`
+3. Click `Install`, which will take you to the GUI installation wizard and you'll be able to fill out the necessary info
 
-3. Click `Install`, which will take you to the GUI Wizard and you'll be able to fill out the necessary info
+   - Published Server URL to publish in UDP Auto Discovery response
+   - User and Group Configuration, Network Configuration
```
felix920506, line 325: Mention that 'host' mode will be required if DLNA is used

felix920506 (changes requested):

felix920506:
@nyanmisaka please review the HWA related items

docs/general/installation/container.md
p0358: Ohh, so that's why DLNA didn't seem to work for me xD

docs/general/installation/container.md
p0358: Okay I'll add that line. But now this makes me again wonder about the total limits, because I've read some forum posts online etc. and it might be that Docker app's total memory usage limit also applies to app+all tmpfs's combined! There isn't too much info about it online other than someone complaining that the behavior is not documented. If that's true then users might need to adjust the below too (but I'm really not sure if yes or no...):

obraz

Maybe I should add a new item there to say that if someone's boot disk is the only SSD in the system, to mount it to Host Path at `/var/tmp/jellyfin_transcodes`?

docs/general/installation/container.md
felix920506: it is mentioned in the docs https://jellyfin.org/docs/general/installation/container#docker

p0358: Yeah I found it there after you mentioned it. Must have missed it otherwise, and people using Docker Compose setups are already covered since it's put there in the example config file.

docs/general/installation/container.md
felix920506: You should maybe ask on the Truenas scale forums for clarification. If the memory limit excludes tmpfs then 4GB should be sufficient for most setups.

p0358: I think I'll add "Note that total memory usage may also be capped in the `Resources Configuration` section." for the time being, so that hopefully people notice that if they were to still run into out of memory problems due to that.

We could ask on the forums, but I didn't find anything mentioned in the docs and iirc I glanced over how they configure Docker and they literally just put in these values to respective configuration sections there, which means it should behave like vanilla Docker, so with that I bet it will exhibit this behavior unless they were to end up adding size of user tmpfs's to the configured total memory limit...

p0358:
I added forum post here with the question about tmpfs limits: https://forums.truenas.com/t/question-about-tmpfs-mount-for-apps-and-its-memory-limits-in-electric-eel/21473 (but it's a miniscule problem, so I wouldn't hold the whole PR just for this if we don't get a concrete answer)

Btw there's also more thorough instructions at TrueNAS docs themselves for Jellyfin: https://www.truenas.com/docs/truenasapps/communityapps/jellyfin/

We could consider linking them too somewhere at the end maybe? But it's worth noting that the screenshots are based on older TrueNAS release, so they're gonna be a bit outdated compared to ours now.

felix920506:
You don't need to force push with every change. We squash when merging.

p0358:
Fair. Just a habit for situations where there's multiple logically separated commits potentially, but that's not really in docs repo I suppose

p0358:
- fixed mention of swap, I learnt that since 24.04 swap is globally disabled by default
- changed wording to `You may need to increase the limit if you have many media files streamed with transcoding at once, or transcodes might fail.`, since when you think about it, the size of your library doesn't really matter at all, what matters is how many people try to transcode at the same time and what kind of bitrate they're using (the latter largely boiling down to whether they stream from LAN or WAN since autodetection)

But now that I think about it, large media files might result in larger default bitrate selected and thus also more storage needed by default. Hmm, maybe I should re-add the "large" word then. Let me know if I should please, because I'm actually not sure how much the original file size influences the automatic bitrate...

As for the Docker memory usage limit, I got a response on the forums post:

> We’ve had some internal discussion about it, but ultimately anything I say about the tmpfs behavior would be a guess. As you’ve seen, Docker docs aren’t very helpful on the subject and looking around online I’ve seen both things saying that the tmpfs mount is and is not limited by the overall container restrictions.
> 
> Since we’re passing through default docker behavior there, I think the best bet barring extensive testing is to reach out to the docker devs and try to get an answer from the source.

But Docker devs were kinda silent about it on the posts I found on their forums. So for now I'd just leave that cautious sentence there, and it could be revisited in case the problem comes up for someone and they'll find out empirically:

> Note that total memory usage may also be capped in the `Resources Configuration` section.

felix920506:
Single large files and many small files at the same time can both cause problems with transcode directory filling up. so the large files should be added back.

docs/general/installation/container.md
```diff
@@ -311,28 +311,35 @@ SuccessExitStatus=0 143
...
+       - If your only SSD storage is the boot drive, you may configure a Host Path mount to `/var/tmp/jellyfin_transcodes`.
+     - For media, add your library directories as Additional Storage.
+     - For hardware-accelerated transcoding, under the Resources Configuration:
+       - If using NVIDIA, tick `Use this GPU` under your listed GPU. If you don't see your GPU, go to `Apps` -> `Configuration` -> `Settings` -> Install NVIDIA Drivers.
+       - If using other vendors, tick `Passthrough available (non-NVIDIA) GPUs`.
+       - Then in Jellyfin UI, configure it under `Dashboard` -> `Playback` -> `Transcoding`.
+       - Note that some newer Intel GPUs might require a newer kernel version than provided. If that's the case for your GPU, you may consider using a VM with PCIe pass-through instead of app containers.
```
felix920506, line 341: Add some instructions on how to check kernel version on Truenas Scale

p0358: I mean just `uname -a` (System -> Shell or SSH), just standard stuff for Linux. It's not exposed anywhere in the UI specifically as far as I see. So I'm thinking that it might be bloat, since if someone doesn't know how to check it, they could google it easily imo...

And by extension I think that just getting the kernel version here won't actually answer the question they'll have. It'll be "okay I have 6.6 and now what? is it new enough or not?". I think if anything ideally we could link "might require a newer kernel version" to some place that documents what new hardware expects what new kernel. So just knowing the version won't be immediately too helpful :/

We could link to:
- https://jellyfin.org/docs/general/administration/hardware-selection
- https://jellyfin.org/docs/general/administration/hardware-acceleration/known-issues#intel-on-linux

I guess none of these currently very directly answer the question, unless we were to link the second one under "some newer Intel GPUs" fragment?

felix920506: Link to the known issues one. I'd say the command would still be useful to include since Truenas users are less likely to be familiar with actual Linux administration.

p0358: I think they'll be fiiiine, since if the issue of new hardware pertains to them, they'll have to do some separate detailed reading and dig into it anyways, so it won't help too much. And those who have decently old hardware can just ignore the warning then

nyanmisaka (approved):

felix920506 (approved):
Looks good enough for now

docs/general/installation/container.md
```diff
@@ -311,28 +311,35 @@ SuccessExitStatus=0 143
...
+   - Network Configuration
+     - Tick `Host Network` if DLNA is used, otherwise it's optional.
+     - You may optionally change the `WebUI Port` to the Jellyfin's default of `8096` if it's not occupied already by another app.
+   - Storage Configuraton
+     - For `Jellyfin Config Storage` and `Jellyfin Cache Storage`, consider mapping Host Paths for ease of backup.
+       - Store them on SSD storage if possible, as using HDD storage **will** lead to poor experience.
+     - For `Jellyfin Transcode Storage`, consider using `Temporary` or `tmpfs`
+       - `Temporary` places a Docker volume under your configured Apps dataset (`Apps` -> `Configuration` -> `Choose Pool`). Avoid if that's on HDD.
+       - `tmpfs` creates a temporary directory on the RAM. Ensure you set the limit to at least a few gigabytes and that you have considerate amount of spare memory, since TrueNAS disables swap by default. You may need to increase the limit if you have many large media files streamed with transcoding at once, or transcodes might fail. Note that total memory usage may also be capped in the `Resources Configuration` section.
```
Gauvino, line 334:
```suggestion
       - `tmpfs` creates a temporary directory on the RAM. Ensure you set the limit to at least a few gigabytes, and that you have considerate amount of spare memory, since TrueNAS disables swap by default. You may need to increase the limit if you have many large media files streamed with transcoding at once, or transcodes might fail. Note that total memory usage may also be capped in the `Resources Configuration` section.
```

Gauvino (changes requested):
Little typo over-all it's great

PR #1165, opened 2024-10-24
Change sudo nano to sudoedit

madscientist16:
`sudoedit` is safer than running `nano` with `sudo`. I changed the 3 entries I could find that had `sudo nano` to `SUDO_EDITOR=nano sudoedit`

### DIFF
--- a/docs/general/installation/linux.md
+++ b/docs/general/installation/linux.md
@@ -384,7 +384,7 @@ sudo apt install -f
 Due to the number of command line options that must be passed on to the Jellyfin binary, it is easiest to create a small script to run Jellyfin.
 
 ```sh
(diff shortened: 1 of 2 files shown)


joshuaboniface (changes requested):
I think this is a good idea. For our trivial uses, I doubt the safety is a big deal, but it would get it into people's consciousness.

That said, adding `SUDO_EDITOR=` to each command makes them unwieldy to type. I'd remove that and just change `sudo nano` to `sudoedit` generally and leave it at that. If users know what `$EDITOR` is and set it, good, and if not, it will use the default (which is probably `nano` anyways).

joshuaboniface (approved):

PR #1169, opened 2024-10-26
Update social.png in meta tags to fix cache on some services

nielsvanvelzen:
Some services are still caching our old social.png with the old logo. Updating the URL with a cache buster should solve this.

### DIFF
--- a/docusaurus.config.ts
+++ b/docusaurus.config.ts
@@ -19,7 +19,7 @@ const config: Config = {
   organizationName: 'jellyfin',
   projectName: 'jellyfin.org',
   themeConfig: {
-    image: 'images/social.png',
+    image: 'images/social.png?v2',
     metadata: [
       { name: 'og:type', content: 'website' },
       { name: 'twitter:card', content: 'summary_large_image' },


thornbill (commented):
Maybe use the date or year?

ferferga (changes requested):
If we had this image imported by the bundler (https://docusaurus.io/docs/static-assets#in-jsx), it would have cache busting by default, instead of relying in copying static assets at the output path after build, which is flakey as we're seeing.

Ser also "Import statement" section from: https://docusaurus.io/docs/markdown-features/assets#images

nielsvanvelzen:
This is part of the docusaurus config though, so no we can't bundle it without circumventing the config.

ferferga:
@nielsvanvelzen It shouldn't matter since the config is also JS. I you want I can do it, can I push to this PR?

nielsvanvelzen:
Yeah sure, if you can figure it out go ahead. But AFAIK the config is handled differently from all other sources.

ferferga:
@nielsvanvelzen You're right and the solution I could come up with is very convoluted and not worth it because it requires an additional webpack config for the docusaurius config file.

I think the best course of action long term is to follow Astro's approach: Config file is just for building, while all the page configuration lies in the page files themselves as JSX elements (see [Astro SEO](https://github.com/jonasmerlin/astro-seo?tab=readme-ov-file#how-to-use) docs).

Not sure though if that's suitable for this PR, what do you think?

nielsvanvelzen:
I think we should prefer the built-in configuration options of Docusaurus, otherwise we'll probably run into issues when upgrading versions.

ferferga (approved):

PR #1185, opened 2024-11-01
Inform about add-on mode subtitles issues

intereglementet:
Users that typically use subtitles should be made aware of the long standing synchronization issues that are attached to using the add-on instead of native.


thornbill:
Native mode is dead?

intereglementet:
I hope not. That would be a serious issue with international users of the Kodi plugin. I tried add-on mode myself with a release a few weeks back and I am sorry to say it is a bit of a trainwreck to try to get subs to work reliably. I switched back to native immediately.

The setting is not visible in GUI, though. But it still works, and it works really great.

thornbill:
It's right above the bit you added:

> Do not use Native mode! Use Addon-mode instead.
> 
> Jellyfin 10.10 no longer provides the shared network folder via the API, and as such native mode does not get the required information to function.

felix920506:
Native mode is dead.

PR #1192, opened 2024-11-06
ADD: Fail2ban guide for IP blocking on upstream server to Fail2ban guide (useful if jellyfin is hosted behind an upstream reverse proxy).

famesjranko:
Hey Jellyfin team,

I wanted to suggest adding a setup guide to the Jellyfin documentation that covers using Fail2Ban for users who are running Jellyfin behind a reverse proxy on a separate upstream server. In my current setup, I have Jellyfin running in a Docker container on a local server, but it's accessible to the internet through an upstream reverse proxy server. 

In my experience, trying to block IPs with Fail2Ban on the Jellyfin host didn’t work effectively because Docker’s internal networking can bypass the host’s `iptables` rules, allowing banned IPs to still reach Jellyfin.

So I set up Fail2Ban on the Jellyfin host to automatically ban IP addresses on the upstream server using `iptables` whenever unauthorised login attempts are detected.

I’ve put together a basic guide that walks through the setup: [Fail2Ban Upstream Proxy Chain Setup Guide](https://github.com/famesjranko/server-scripts/blob/main/fail2ban_upstream_proxy_chain_setup_guide.md).

Here’s an overview of my setup:
```
internet -> upstream-proxy <- (ban/unban IP commands) <- Fail2Ban (monitors logs)
```
```
internet -> upstream-proxy -> (allowed traffic) -> host(docker[jellyfin])
```

### What I Did:
- **Fail2Ban Configuration:** Fail2Ban is installed on the Jellyfin host and monitors the Jellyfin logs for failed login attempts or other triggers.
- **Upstream Proxy Banning:** Instead of banning IPs directly on the Jellyfin host, Fail2Ban executes `iptables` commands via SSH on the upstream reverse proxy server. This ensures that banned IPs are blocked before they reach the Jellyfin host, adding an extra layer of protection.
- **Flexible Approach:** I implemented two options for managing IP bans:
  1. **Dynamic Chains:** Fail2Ban creates and manages individual `iptables` chains for each jail on the upstream server, allowing for more granular control of IP blocking by service.
  2. **Communal Chain:** All Fail2Ban jails share a single `iptables` chain on the upstream proxy, simplifying the management of banned IPs.

### Why This Is Useful:
- **Enhanced Security:** This setup can be especially helpful for users running Jellyfin behind a reverse proxy, as it blocks malicious traffic at the proxy level before it reaches the local network.
- **Scalable:** This method works well for users with multiple services running behind the same reverse proxy, not just Jellyfin, as Fail2Ban can be configured to manage bans for all services centrally.
- **Adaptable:** The guide includes dynamic chain management, allowing advanced users to customise their ban rules based on individual services, or use a communal approach for simplicity.

I believe this setup could be valuable for other Jellyfin users with similar reverse proxy configurations and would be a useful addition to the documentation.

### DIFF
--- a/docs/general/networking/fail2ban.md
+++ b/docs/general/networking/fail2ban.md
@@ -227,7 +227,7 @@ Replace `<upstream-server-ip>` with the actual IP address of your upstream serve
 
 4. **Verify on Upstream Server**:
 
-   Check if the IP is banned in the corresponding jail's chain on the upstream server ('f2b-<jail-name>'):
+   Check if the IP is banned in the corresponding jail's chain on the upstream server ('f2b-jail-name'):
 
    ```bash
    ssh root@<upstream-server-ip> "iptables -L f2b-jellyfin"
@@ -243,7 +243,7 @@ Replace `<upstream-server-ip>` with the actual IP address of your upstream serve
 
 6. **Verify Unban**:
 
-    Verify that the IP is removed from the corresponding jail's chain ('f2b-<jail-name>'):
+    Verify that the IP is removed from the corresponding jail's chain ('f2b-jail-name'):
 
    ```bash
    ssh root@<upstream-server-ip> "iptables -L f2b-jellyfin"
--- a/docs/general/networking/fail2ban.md
+++ b/docs/general/networking/fail2ban.md
@@ -227,7 +227,7 @@ Replace `<upstream-server-ip>` with the actual IP address of your upstream serve
 
 4. **Verify on Upstream Server**:
 
-   Check if the IP is banned in the corresponding jail's chain on the upstream server (f2b-<jail-name>):
+   Check if the IP is banned in the corresponding jail's chain on the upstream server ('f2b-<jail-name>'):
 
    ```bash
    ssh root@<upstream-server-ip> "iptables -L f2b-jellyfin"
@@ -243,7 +243,7 @@ Replace `<upstream-server-ip>` with the actual IP address of your upstream serve
 
 6. **Verify Unban**:
 
-    Verify that the IP is removed from the corresponding jail's chain (f2b-<jail-name>):
+    Verify that the IP is removed from the corresponding jail's chain ('f2b-<jail-name>'):
 
    ```bash
    ssh root@<upstream-server-ip> "iptables -L f2b-jellyfin"
--- a/docs/general/networking/fail2ban.md
+++ b/docs/general/networking/fail2ban.md
@@ -86,3 +86,197 @@ Assuming you've at least one failed authentication attempt, you can test this ne
 ```bash
 sudo fail2ban-regex /path_to_logs/*.log /etc/fail2ban/filter.d/jellyfin.conf --print-all-matched
 ```
+
+---
+
+## Fail2Ban
+
+[Fail2ban](https://github.com/fail2ban/fail2ban) is an intrusion prevention software framework that protects computer servers from brute-force attacks.
+Fail2Ban operates by monitoring log files (e.g., `/var/log/auth.log`, `/var/log/apache/access.log`, etc.) for selected entries and running scripts based on their content.
+
+Jellyfin produces logs that can be monitored by Fail2Ban to prevent brute-force attacks on your machine.
+
+### Requirements
+
+- Jellyfin remotely accessible
+- Fail2Ban installed and running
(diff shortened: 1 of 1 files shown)


docs/general/networking/fail2ban.md
```diff
@@ -86,3 +86,197 @@ Assuming you've at least one failed authentication attempt, you can test this ne
...
+### Step Three: Test
+
+To test this setup with existing logs:
+
+```bash
+sudo fail2ban-regex /path_to_logs/*.log /etc/fail2ban/filter.d/jellyfin.conf --print-all-matched
+```
+
+---
+
```
felix920506, line 176: why is this section duplicated?

docs/general/networking/fail2ban.md
```diff
@@ -86,3 +86,197 @@ Assuming you've at least one failed authentication attempt, you can test this ne
...
+This guide offers two configurations for setting up Fail2Ban to manage IP bans on an upstream reverse proxy server:
+
+1. **Dynamic Chains**: Each Fail2Ban jail creates and manages its own `iptables` chain on the upstream server.
+2. **Communal Chain**: All Fail2Ban jails share a single `iptables` chain on the upstream server.
+
+### Prerequisites
+
+- **Fail2Ban** is installed on your local server (where services such as Jellyfin, Nginx, etc., are running).
+- **SSH key-based access** is set up from your Fail2Ban server to the upstream server.
```
felix920506, line 189: might want to link a guide

docs/general/networking/fail2ban.md
```diff
@@ -86,3 +86,197 @@ Assuming you've at least one failed authentication attempt, you can test this ne
...
+- **iptables** is configured on the upstream server.
```
felix920506, line 190: will systems that use firewalld work?

docs/general/networking/fail2ban.md
```diff
@@ -86,3 +86,197 @@ Assuming you've at least one failed authentication attempt, you can test this ne
...
+   ssh-copy-id -i /root/.ssh/id_rsa.pub root@<upstream-server-ip>
+   ```
+
+3. **Verify Access**:
+
+   ```bash
+   ssh -i /root/.ssh/id_rsa root@<upstream-server-ip>
+   ```
+
+---
```
felix920506, line 214: why is this both a prerequisite and part of the guide? If you would like to include this here, please remove this from the prerequisites

felix920506 (changes requested):

docs/general/networking/fail2ban.md
famesjranko: this was a mistake - will remove.

docs/general/networking/fail2ban.md
famesjranko: I think it is technically possible, but have not tested.

It should be as simple as changing the actionban and actionunban actions to use firewalld instead - something like:
```
actionban = ssh -i /root/.ssh/id_rsa -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null root@<upstream-server-ip> \
            'firewall-cmd --direct --add-rule ipv4 filter f2b-<name> 0 -s <ip> -j DROP && firewall-cmd --reload'
```

and 
```
actionunban = ssh -i /root/.ssh/id_rsa -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null root@<upstream-server-ip> \
               'firewall-cmd --direct --remove-rule ipv4 filter f2b-<name> 0 -s <ip> -j DROP && firewall-cmd --reload'
```

would need to confirm this though.

docs/general/networking/fail2ban.md
famesjranko: good point - will remove from prereqs.

docs/general/networking/fail2ban.md
famesjranko: link to what guide?  We can presume fail2ban is installed and SSH key generation will now be placed in guide and removed from prereqs. 

This might be resolved with the removal of ssh generation from prereqs?

docs/general/networking/fail2ban.md
felix920506: Can you confirm this?

joshuaboniface (approved):
I do something similar (only I ban on an upstream FreeBSD router rather than a reverse proxy) so I think this is good. Well-written and seems reasonable.

felix920506 (approved):

PR #1196, opened 2024-11-09
Add Fladder client

PartyDonut:


JPVenson (approved):

src/data/clients.ts
```diff
@@ -930,5 +930,28 @@ export const Clients: Array<Client> = [
...
     ]
+  },
+  {
+    id: "fladder",
+    name: "Fladder",
+    description: "A simple, cross-platform Jellyfin frontend built on top of Flutter.",
+    clientType: ClientType.ThirdParty,
+    deviceTypes: [DeviceType.Mobile, DeviceType.Desktop],
+    licenseType: LicenseType.OpenSource,
+    platforms: [Platform.Linux, Platform.Windows, Platform.MacOS, Platform.Android, Platform.IOS, Platform.Docker, Platform.Browser],
```
thornbill, line 941: I don't think Docker is used for clients.

thornbill (changes requested):
Fladder should have a distinct logo. Reusing the Jellyfin logo in such a blatant manner is something we discourage.

PartyDonut:
> Fladder should have a distinct logo. Reusing the Jellyfin logo in such a blatant manner is something we discourage.

The icon for Fladder has been changed

PartyDonut:
@thornbill 
I think all requested changes have been addressed?
If I missed something let me know.

thornbill:
Sorry for the delay and thanks for updating the logo. The only other feedback I have is maybe iOS shouldn't be listed since it isn't available in the App Store and alternate installation methods aren't achievable for the average user. 🤷‍♂️

PartyDonut:
iOS has been removed from the clients list.
Personally I would have preferred to keep it in there, but I also udnerst

PartyDonut:
I've removed iOS from the list 👍🏼 .

Would have preferred to keep it in there but I understand that the installation method doesn't make it feasible for most users.

nielsvanvelzen (approved):

thornbill (approved):

PR #1207, opened 2024-11-26
Update container.md

williamtrelawny:
Add note on `uid:id` argument. This value being a placeholder is not apparent to users.

ref: https://www.reddit.com/r/jellyfin/comments/ub6qwc/comment/i6281g1

### DIFF
--- a/docs/general/installation/container.md
+++ b/docs/general/installation/container.md
@@ -73,6 +73,12 @@ docker run -d \
  jellyfin/jellyfin
 ```
 
+:::note
+
+Replace `uid:gid` if you want to run jellyfin as a specific user/group (must exist in the host `/etc/passwd` file). Exclude the `--user` argument entirely if you want to use the default user.
+
+:::
+
 Bind Mounts are needed to pass folders from the host OS to the container OS whereas volumes are maintained by Docker and can be considered easier to backup and control by external programs.
 For a simple setup, it's considered easier to use Bind Mounts instead of volumes.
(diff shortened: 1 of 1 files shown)


docs/general/installation/container.md
```diff
@@ -73,6 +73,12 @@ docker run -d \
  jellyfin/jellyfin
 ```
 
+:::note
+
+Replace `uid:gid` if you want to run jellyfin as a specific user/group (must exist in the host `/etc/passwd` file). Exclude the `--user` argument entirely if you want to use the default user.
```
Shadowghost, line 78: I don't think `must exist in the host `/etc/passwd` file` is correct. You can input whatever you want.

felix920506: yeah, can be whatever you want

TimGels:
@williamtrelawny please have a look at the review comment.

docs/general/installation/container.md
BotBlake: @williamtrelawny How do you feel about this change?
That way this topic can actually be marked as resolved.

```suggestion
Replace `uid:gid` if you want to run jellyfin as a specific user/group. Exclude the `--user` argument entirely if you want to use the default user.
```

Kind regards.

felix920506:
This PR will be closed as stale by Oct 7 if there are no further replies from the original Author

BotBlake:
This PR is updated by #1572 which implements my suggested changes from here.
All discussion about the actual change will move there.

Thanks to everyone contributing here!

PR #1220, opened 2024-12-04
Update iis.md

kimboslice99:
Long running streams may freeze due to limits within IIS, I have added a command to prevent this issue.

Setting httpVersion wasn't necessary so that has been removed.

### DIFF
--- a/docs/general/networking/iis.md
+++ b/docs/general/networking/iis.md
@@ -20,13 +20,18 @@ IIS with default selections + Application Development->WebSocket Protocol (minim
 ```powershell
 Set-WebConfigurationProperty -pspath 'MACHINE/WEBROOT/APPHOST'  -filter "system.webServer/proxy" -name "enabled" -value "True"
 Set-WebConfigurationProperty -pspath 'MACHINE/WEBROOT/APPHOST'  -filter "system.webServer/proxy/cache" -name "enabled" -value "False"
-Set-WebConfigurationProperty -pspath 'MACHINE/WEBROOT/APPHOST'  -filter "system.webServer/proxy" -name "httpVersion" -value "Http11"
 Set-WebConfigurationProperty -pspath 'MACHINE/WEBROOT/APPHOST'  -filter "system.webServer/proxy" -name "preserveHostHeader" -value "True"
 Add-WebConfigurationProperty -pspath 'MACHINE/WEBROOT/APPHOST'  -filter "system.webServer/rewrite/allowedServerVariables" -name "." -value @{name='HTTP_X_FORWARDED_PROTOCOL'}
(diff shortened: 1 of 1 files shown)


Shadowghost (commented):
I think the HTTP version needs to be set for websockets to work properly. Can you please check?

kimboslice99:
While this may be true for other proxies it does not appear to be the case for ARR. The default is PassThrough which has no issues

Shadowghost (approved):
I don't know enough bout IIS to validate this properly but from a quick look at the docs it seems fine.

felix920506:
We decided to remove IIS from the docs entirely since none of the active team members / community members know how to use it.

kimboslice99:
I am the original writer of that doc and am well experienced with IIS. You can look in my repos for proof of this.

How could I become a community member? I could fulfill any need for IIS debugging or general documentation.

felix920506:
@kimboslice99 the best way would be to just hang out and help out on our forum or in our chat rooms and you will be recognized after a while of regular activity. https://jellyfin.org

kimboslice99:
I’ve been involved in the Jellyfin ecosystem for a couple of years now, with a handful of small commits across the project.

I am surprised that forum or chatroom activity is required to contribute, even for something like writing documentation. This feels like an unusual requirement—most projects I’ve seen focus primarily on the contributions themselves, not external factors like forum presence.

Why is this a part of the process? It feels restrictive and seems to go against GitHub’s open and collaborative nature.

JPVenson:
The forum and chat is by no means required to contribute, but its where all core members are usually present and discuss topics on a day to day bases. It would have made no difference if you are in chat or not, the whole IIS documentation was a very hot discussed topic in the first place if you remember.

We (the team) made the decision to put the docs up and see how it goes but we came to the conclusion that its not a platform we _want_ to support officially (hence the removal of the docs) because as felix said, no person assoicated with the project is able to support it, mainly in chat where the vast majority of support questions come up.

We have no way of reviewing this kind of change and for community members we also usually assume some sort of active discussion in chat for day-to-day topics but then you would need to be online most of the time if/when such questions would come up. 

In the end, if nobody really uses it we still need to support it and it adds no benefit and as soon as more people use it, we can still not support it because nobody from the team uses it. I hope you see our issue with it.

I know that this might not be the most satisfying answer i can give you, but we cannot do it all. Regardless i would be happy if you would join our matrix/discord server and engage with our community.

CaiB:
@kimboslice99 Your contributions and IIS expertise would be much appreciated in my (non-official) documentation for running Jellyfin on Windows Server/IIS: https://github.com/CaiB/Jellyfin-IIS-Snarky-Guide

PR #1222, opened 2024-12-09
Update nginx.md

jameskimmel:
small corrections

### DIFF
--- a/docs/general/networking/nginx.md
+++ b/docs/general/networking/nginx.md
@@ -40,8 +40,8 @@ server {
     ## The default `client_max_body_size` is 1M, this might not be enough for some posters, etc.
     client_max_body_size 20M;
 
-    # Uncomment next line to disable TLS 1.0 and 1.1 (Might break older devices)
-    #ssl_protocols TLSv1.3 TLSv1.2;
+    # Next line disables TLS 1.0 and 1.1 (Might break older devices)
+    ssl_protocols TLSv1.3 TLSv1.2;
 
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

felix920506 (changes requested):
The original config was correct. Change the comment to tell users to comment next line to enable TLS 1.0 and 1.1 as your devices will have to be truly ancient to not support TLS 1.2 or newer.

jameskimmel:
> The original config was correct. Change the comment to tell users to comment next line to enable TLS 1.0 and 1.1 as your devices will have to be truly ancient to not support TLS 1.2 or newer.

I don't thinks so.

`Version 1.23.4 and later: the default SSL protocols are TLSv1, TLSv1.1, TLSv1.2, and TLSv1.3`


Source: https://nginx.org/en/docs/http/configuring_https_servers.html

felix920506:
Just because it's the "default" doesn't mean it's better. TLS 1.0 and 1.1 have well documented vulnerabilities at this point so disabling them is more secure.

jameskimmel:
> Just because it's the "default" doesn't mean it's better. TLS 1.0 and 1.1 have well documented vulnerabilities at this point so disabling them is more secure.

I never said so. 
And I agree with you. 
But then we would need to change it the other way round. 
Make TLS 1.2 and 1.3 the default by not commenting them out and write in a comment that users with older devices can comment that line if they have older incompatible devices.

felix920506:
Yes that is what I meant in my review comment

felix920506 (approved):

PR #1239, opened 2024-12-31
Add LineageOS Twelve music client

joshuaboniface:
As announced [here](https://lineageos.org/Changelog-29/), the built-in LineageOS 22 Music client now supports Jellyfin servers as a source (amongst others), so include it in our apps list.

The formatting might be a bit weird because LineageOS doesn't _officially_ support installing the APK directly, but links are provided to various places to get this (the third is a known developer for the apps and is trustworthy).

### DIFF
--- a/src/data/clients.ts
+++ b/src/data/clients.ts
@@ -930,5 +930,37 @@ export const Clients: Array<Client> = [
         url: 'https://tilo.dev/manet/'
       }
     ]
-  }
+  },
+  {
+    id: 'lineageos-twelve',
+    name: 'Twelve (LineageOS)',
+    description: 'LineageOS 22 built-in music app ("Music") with Jellyfin support.',
+    clientType: ClientType.ThirdParty,
+    deviceTypes: [DeviceType.Mobile],
+    licenseType: LicenseType.OpenSource,
+    platforms: [Platform.Android],
+    primaryLinks: [
+      {
+        id: 'lineageos',
+        name: 'Main LineageOS Page',
+        url: 'https://lineageos.org/'
+      {
+        id: 'github',
+        name: 'GitHub (Official)',
+        url: 'https://github.com/lineageos/android_packages_apps_glimpse/actions/workflows/build.yml?query=branch%3alineage-21.0'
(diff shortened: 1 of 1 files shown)


felix920506:
I think we should at least give the LineageOS team a heads up before we merge this.

src/data/clients.ts
```diff
@@ -930,5 +930,37 @@ export const Clients: Array<Client> = [
...
+    clientType: ClientType.ThirdParty,
+    deviceTypes: [DeviceType.Mobile],
+    licenseType: LicenseType.OpenSource,
+    platforms: [Platform.Android],
+    primaryLinks: [
+      {
+        id: 'lineageos',
+        name: 'Main LineageOS Page',
+        url: 'https://lineageos.org/'
+      {
```
nielsvanvelzen, line 947:
```suggestion
      },
      {
```

nielsvanvelzen:
> I think we should at least give the LineageOS team a heads up before we merge this.

In fact, to list a client on our website we ask the author of the client to make the pull request or approve it (by replying here). Otherwise we won't merge it.

src/data/clients.ts
```diff
@@ -930,5 +930,37 @@ export const Clients: Array<Client> = [
...
+        id: 'github',
+        name: 'GitHub (Official)',
+        url: 'https://github.com/lineageos/android_packages_apps_glimpse/actions/workflows/build.yml?query=branch%3alineage-21.0'
```
nielsvanvelzen, line 950: This URL is not a user friendly download page so we shouldn't include it. Additionally the query in the URL refers to eleven, not twelve.

src/data/clients.ts
```diff
@@ -930,5 +930,37 @@ export const Clients: Array<Client> = [
...
+      },
+      {
+        id: 'sebaubuntu',
+        name: 'SebaUbuntu (Mirror)',
+        url: 'https://www.sebaubuntu.dev/lineageapps.html'
```
nielsvanvelzen, line 955: This one doesn't feel like any official source

PR #1245, opened 2025-01-09
Update truenas.md

IetIesAai:
the folder for transcodes by default in the docker app seems to be cache/transcodes, instead of config/transcodes as indiecated in this part of the docs 

ps: I hope I am doing this the right way- I am not experienced with this process....

### DIFF
--- a/docs/general/installation/truenas.md
+++ b/docs/general/installation/truenas.md
@@ -37,7 +37,7 @@ The official Jellyfin Docker image internally creates the following necessary di
 
 1. cache
 2. config
-3. config/transcodes
+3. cache/transcodes
 
 If you use the community app, you can allow SCALE to create datasets for the directories Jellyfin requires automatically during app installation.
 You can also choose to create a static **transcodes** dataset or use temporary storage on the disk or in memory (system RAM) for transcoding.


maru801:
It used to be in the past (v10.8.13) that transcodes would reside in `config/transcodes` by default.

I did some tests and indeed the current behavior does switch this to default to `cache/transcodes`.

I find it weird however that the Docker image still automatically creates a transcode folder in the config directory, but doesn't use it now.

I added `config/transcodes` into the docs because that's what was default on my server since it was created back then.
I'm going to wait a bit to see if I hear anything about this before reviewing and approving this PR.

maru801:
Messed around a bit more and found out the following:
1. The default of `config/transcodes` was changed to `cache/transcodes` starting at v10.9.0
2. Using the official Docker image of Jellyfin now, a transcodes folder will not be made in `/config` anymore (only in `/cache` now)
3. The Community App from Truenas Scale's repository for some reason creates a transcode folder in both `/cache` and `/config` despite `/config` no longer being the default
4. If you created your server during v10.8.13 or before and migrated the `/config` folder through each update, `/config/transcodes` will still be used (if you didn't change it before)

That should clear up why this confusion on my part happened.

maru801 (approved):
Looks good.

felix920506 (approved):

PR #1250, opened 2025-01-11
Update Manet to include the MacOS platform

bzf:
This adds the recently added macOS support for the music player Manet on the clients page.

The macOS client requires the in-app subscription, so not sure if you still want to have it on the website? I'll mention it here so we can discuss it at least so you don't get surprised by it 😊

### DIFF
--- a/src/data/clients.ts
+++ b/src/data/clients.ts
@@ -911,11 +911,11 @@ export const Clients: Array<Client> = [
   {
     id: 'manet',
     name: 'Manet',
-    description: 'A third-party music client for iOS',
+    description: 'A third-party music client for iOS and macOS',
     clientType: ClientType.ThirdParty,
-    deviceTypes: [DeviceType.Mobile],
+    deviceTypes: [DeviceType.Mobile, DeviceType.Desktop],
     licenseType: LicenseType.Proprietary,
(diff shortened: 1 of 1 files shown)


src/data/clients.ts
```diff
@@ -915,7 +915,7 @@ export const Clients: Array<Client> = [
     clientType: ClientType.ThirdParty,
     deviceTypes: [DeviceType.Mobile],
```
felix920506, line 916: You should update this as well

bzf: Great find, pushed an update! 🙌🏽

JPVenson (approved):

joshuaboniface (approved):

src/data/clients.ts
```diff
@@ -913,9 +913,9 @@ export const Clients: Array<Client> = [
     name: 'Manet',
     description: 'A third-party music client for iOS',
```
felix920506, line 914: You should put macOS here as well. I think a mention of the requirement of subscription would be nice but not strictly required per the guidelines.

felix920506 (approved):

nielsvanvelzen (approved):

PR #1251, opened 2025-01-11
Update client guidelines according to previous conversations in matrix

felix920506:
According to a previous conversation on matrix

Question:
The guideline says `In particular, the client must **NOT** engage in, encourage, or facilitate piracy.`
Does that include integration with the *arr stack, jellyseerr or other tools commonly used for media acquisiton?

### DIFF
--- a/docs/general/clients/index.mdx
+++ b/docs/general/clients/index.mdx
@@ -32,15 +32,16 @@ Please verify it meets the requirements below and [submit a pull request](https:
 
 ### Requirements for Inclusion in All Clients
 
-- Any client meeting the following requirements:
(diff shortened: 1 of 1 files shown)


joshuaboniface (approved):

joshuaboniface:
> Does that include integration with the *arr stack, jellyseerr or other tools commonly used for media acquisiton?

I think it's a fine line and requires a specific example of what "integration" means. Jellyseerr I think is on the alright side of the line, ARRs are on the not alright side of the line. Also if integration is just convenience links or something configurable, I think that's fine.

PR #1254, opened 2025-01-12
Warn about a .nfo genre tag and Albums/Artists

Jamy-L:
Add a note mentioning that album.nfo and artist.nfo are not compatible with the genre tag, as mentioned in jellyfin/jellyfin#13350.

### DIFF
--- a/docs/general/server/metadata/nfo.md
+++ b/docs/general/server/metadata/nfo.md
@@ -37,10 +37,6 @@ It's currently not possible to disable .nfo metadata. Local metadata will always
 User data importing is only possible for a single user. This user can be set in the .nfo settings.
 :::
 
-:::note
-Music Albums and Music Artists are currently not compatible with the genre tag, that will just be ignored.
-:::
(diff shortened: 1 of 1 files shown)


docs/general/server/metadata/nfo.md
```diff
@@ -37,6 +37,10 @@ It's currently not possible to disable .nfo metadata. Local metadata will always
 User data importing is only possible for a single user. This user can be set in the .nfo settings.
 :::
 
+:::note
+Music Albums and Music Artists are currently not compatible with the genre tag, that will just be ignored.
```
gnattu, line 41: I prefer to have this is the "note" section in the following table, or make a link in that table that points to this warning.

Jamy-L: You are right, it is cleaner this way. I have moved it in the table

gnattu (approved):

felix920506 (approved):

PR #1269, opened 2025-01-22
Add a warning to warn against the use of IIS

felix920506:
Basically no one in our community / on our team can troubleshoot it effectively, so add a warning to warn users against using it.


nielsvanvelzen:
Then just remove the page instead of those annoying admonitions everywhere

felix920506:
Fair point

PR #1277, opened 2025-01-25
Nginx Documentation Re-work

solidsnake1298:
- Simplifying example to minimum "secure" config.
- Adding ```font-src``` to CSP to resolve https://github.com/jellyfin/jellyfin.org/issues/1275
- Removing http examples.  Not appreciably different that https.
- Adding Nginx Proxy Manager sub-path example as provided by forum user.  https://forum.jellyfin.org/t-nginx-proxy-manager-config?pid=42446#pid42446

### DIFF
--- a/docs/general/networking/nginx.md
+++ b/docs/general/networking/nginx.md
@@ -94,16 +94,6 @@ server {
 }
 ```
 
-### HTTP config example
-
-:::tip
-
-If you are planning on exposing your server over the Internet, you should setup HTTPS. [Let's Encrypt](https://letsencrypt.org/getting-started/) can provide free TLS certificates which can be installed easily via [certbot](https://certbot.eff.org/). Using only HTTP will expose passwords and API keys.
-
-:::
-
-</details>
-
 ## Nginx with Subpath (example.org/jellyfin)
 
 When connecting to server from a client application, enter `http(s)://example.org/jellyfin` in the address field.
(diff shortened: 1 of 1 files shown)


felix920506 (approved):
Someone familiar with nginx should double check

joshuaboniface (approved):

nielsvanvelzen:
If we really want to simplify this page I'd suggest some additional removals:
- Remove SSL config (rely on default options)
- Remove X-Content-Type-Options header
- Remove permission policy header
- Remove CSP header
- Remove "nginx with subpath" part, just focus on a normal install with dedicated subdomain
- Remove image cache example, the server already caches images
- Remove download rate limiting!! This should've never been added in the first place as clients do NOT expect this

solidsnake1298:
> If we really want to simplify this page I'd suggest some additional removals:
> 
> * Remove SSL config (rely on default options)
> * Remove X-Content-Type-Options header
> * Remove permission policy header
> * Remove CSP header
> * Remove "nginx with subpath" part, just focus on a normal install with dedicated subdomain
> * Remove image cache example, the server already caches images
> * Remove download rate limiting!! This should've never been added in the first place as clients do NOT expect this

The goal isn't to simplify down to the absolute minimum for a working HTTPS proxy.   I want this to be the minimum secure config that doesn't break 99.999% of user clients.  The exception is super super old clients, which I mention for the ```ssl_protocols``` line, which would compromise security unnecessarily if that was not configured.

Subpaths are pretty common for users that don't want wildcard certs or multiple subdomain certs.

The image cache is useful for proxies not located on the same host as Jellyfin.  A common use case to provide an example for, but clearly labeled as optional.  Same with download limiting for users with limited upload speed.

However, my download limit testing was flawed and it does not work as documented.  Will remove that for now.

PR #1284, opened 2025-01-29
Update nginx.md

Shiro-2000:
Updating the nginx documentation around setting up caching for images and video in the reverse proxy configuration. 

I recently followed the documented steps to setup the reverse proxy and ran into issues. I am updating the guide so others will not have the issues.


Shiro-2000:
Video Caching is not recommended in the reverse proxy. Kestrel does video caching at jellyfin. 
Caching video with NGINX can cause a bunch of problems, specifically with bit rate control.

PR #1285, opened 2025-02-01
Rewrote changing protocols in apache, added a cache-control header directive

Andrew-Geo:
The previous directives for managing WebSocket connection upgrades were not working with syncplay. Fixed with two simple rewrite rules.

Added a strict cache-control header for asking clients not to keep cache. This fixes UI updating (e.g. when reloading libraries) and some syncplay issues.

## Summary by CodeRabbit

- **Documentation**
  - Updated Apache reverse proxy setup instructions to use dynamic mod_rewrite rules for improved WebSocket and HTTP handling.
  - Added guidance to set a `Cache-Control` header to prevent client-side caching and improve display and sync reliability.


docs/general/post-install/networking/8_reverse-proxy/apache.md
```diff
@@ -34,11 +34,17 @@ title: Apache
...
+    RewriteEngine On
+    RewriteCond %{HTTP:Upgrade} =websocket
+    RewriteRule /(.*) ws://SERVER_IP_ADDRESS:8096/socket/$1 [P,L]
+    RewriteCond %{HTTP:Upgrade} !=websocket
+    RewriteRule /(.*) http://SERVER_IP_ADDRESS:8096/$1 [P,L]
+
+    # Sometimes, Jellyfin requires clients to empty their cache to display and function correctly.
+    # This header tells clients not to keep any cache and is quite strict on that.
+    # This might also fix some syncplay issues (#5485 and #8140 @ https://github.com/jellyfin/jellyfin-web/issues/)
+    Header set Cache-Control "no-store, no-cache, must-revalidate, max-age=0"
```
BotBlake: I can see users copying this example over blindly.
Therefore I would say mark this additional setting as a comment, so it can be commented out by the users if they find they need it.

```suggestion
    # Header set Cache-Control "no-store, no-cache, must-revalidate, max-age=0"
```

felix920506:
@coderabbitai review

docs/general/post-install/networking/8_reverse-proxy/apache.md
Andrew-Geo: That's great thinking, probably makes the directive seem more like a suggestion or an opt-in fix than a required step. 
Thanks for the feedback/

docs/general/post-install/networking/8_reverse-proxy/apache.md
```diff
@@ -34,11 +34,17 @@ title: Apache
...
-    ProxyPassReverse "/socket" "ws://SERVER_IP_ADDRESS:8096/socket"
-
-    ProxyPass "/" "http://SERVER_IP_ADDRESS:8096/"
-    ProxyPassReverse "/" "http://SERVER_IP_ADDRESS:8096/"
+    # Apache should be able to know when to change protocols (between WebSocket and HTTP)
+    RewriteEngine On
+    RewriteCond %{HTTP:Upgrade} =websocket
+    RewriteRule /(.*) ws://SERVER_IP_ADDRESS:8096/socket/$1 [P,L]
+    RewriteCond %{HTTP:Upgrade} !=websocket
+    RewriteRule /(.*) http://SERVER_IP_ADDRESS:8096/$1 [P,L]
```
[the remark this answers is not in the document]
Andrew-Geo: Adding the no-case flag seems of no need.

I haven't seen any client doing upgrade headers for websocket with the websocket value being in anything but all lowercase.
Also all relevant docs seem to follow this. (e.g. [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Protocol_upgrade_mechanism#upgrading_to_a_websocket_connection))

Besides, although probably not that important, [NC] might also add overhead to rewrite rules which are already expensive, as far as directives are concerned.

I see the [NC] flag as a needless complexity, but I definitely need input on this from someone.

PR #1316, opened 2025-03-04
Add Finer into clients

wddwycc:
Hi team, thanks for the work on Jellyfin community.  

I've recently built a new music client for Jellyfin, the app was originally built for macOS, and expanded to iPadOS/iOS as requested by users, and it's free. 

The player is built with performance and minimum resource usage in mind, and has some very unique features like 3rd party lyrics syncing, lyrics translation and push notification integration.  

Check out https://monk-studio.com/finer. 

Screenshot 2025-03-04 at 23 32 30

------------------

This MR added Finer to client list, also fixed type-missing of IpadOS, and prettier violation of the ts file.

### DIFF
--- a/src/data/clients.ts
+++ b/src/data/clients.ts
@@ -882,7 +882,7 @@ export const Clients: Array<Client> = [
     clientType: ClientType.ThirdParty,
     deviceTypes: [DeviceType.Mobile],
     licenseType: LicenseType.OpenSource,
-    platforms: [Platform.Android, Platform.IOS, Platform.iPadOS],
+    platforms: [Platform.Android, Platform.IOS, Platform.IpadOS],
     primaryLinks: [
       {
         id: 'app-store',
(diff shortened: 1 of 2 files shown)


src/data/clients.ts
```diff
@@ -882,7 +882,7 @@ export const Clients: Array<Client> = [
     clientType: ClientType.ThirdParty,
     deviceTypes: [DeviceType.Mobile],
     licenseType: LicenseType.OpenSource,
-    platforms: [Platform.Android, Platform.IOS, Platform.iPadOS],
+    platforms: [Platform.Android, Platform.IOS, Platform.IpadOS],
```
felix920506, line 885: why are you touching unrelated clients?

src/data/platform.ts
```diff
@@ -7,6 +7,7 @@ enum Platform {
   Discord = 'Discord',
   FireOS = 'Fire TV',
   IOS = 'iOS',
+  IpadOS = 'iPadOS',
```
felix920506, line 10: we don't need an ipados entry when iOS and iPadOS are functionally the same.

wddwycc: Hi, I noticed this has been used by `streamyfin` but it's actually not exist  in this enum, so I added here as a fix.

wddwycc: Sure thing, updated. Now this MR only adds one entry for the clients json.

joshuaboniface (approved):

felix920506 (approved):

PR #1319, opened 2025-03-05
Move internet radio to Live TV category

felix920506:
Internet radio is added as a Live TV source so I think it would be more appropriate to be under the Live TV category compared to the Media category.


scripts/data/jellyfin-docs-urls.json
```diff
@@ -39,10 +39,10 @@
   "/docs/general/server/live-tv/index.html",
   "/docs/general/server/live-tv/post-process.html",
   "/docs/general/server/live-tv/setup-guide.html",
+  "/docs/general/server/live-tv/internet-radio.html",
   "/docs/general/server/media/books.html",
   "/docs/general/server/media/comics.html",
   "/docs/general/server/media/external-files.html",
-  "/docs/general/server/media/internet-radio.html",
```
thornbill, line 45: Keep this and add a redirect

docs/general/server/live-tv/setup-guide.md
```diff
@@ -1,6 +1,7 @@
 ---
 uid: server-live-tv-setup-guide
 title: Setup Guide
```
theguymadmax, line 3: title: Live TV Setup Guide

docs/general/server/live-tv/internet-radio.md
```diff
@@ -1,6 +1,7 @@
 ---
-uid: server-media-internet-radio
+uid: server-live-tv-internet-radio
 title: Internet radio
```
theguymadmax, line 3: Fix capitation to match other items in this sub-menu.  Or change the other item's titles to match.

theguymadmax:
The contents of Live TV and Internet Radio menu should be moved to a subpage and then Live TV and Internet Radio should just be a drown menu like the Media page.

jhparker88:
> The contents of Live TV and Internet Radio menu should be moved to a subpage and then Live TV and Internet Radio should just be a drown menu like the Media page.

I find it generally concerting to put Internet Radio and LiveTV under the page of Live TV because those are to distinct categories of streaming and it would make much more sense to separate them under:

```
Live streaming
 |- Live TV
 |- Internet radio
```

theguymadmax:
> find it generally concerting

Yes, this was what I was thinking too. I was also going to suggest renaming it to Live Streaming. The Live TV and Internet Radio page doesn't have anything referring to Internet Radio, that's why it should be a subpage of Live Streaming

felix920506:
I think calling it "Live Streaming" is very misleading given that what most people think of Live Streaming these days is something like Twitch

theguymadmax:
I guess that's a generational difference b/c Twitch never even crossed my mind. If it were to be renamed, another option to consider could be "Live Streaming Media".

jhparker88:
> I think calling it "Live Streaming" is very misleading given that what most people think of Live Streaming these days is something like Twitch

Well the only denominator between Live TV and Internet radio is the IP "stream" that is used as transfer medium, those two have nothing else in common (taken aside the technical implementation details of m3u etc.). Radio is for audio only while TV is a combined stream of video and audio.

I wouldn't associate "Live streaming" with anything like Twitch in the the first place and while one can it is generally to be know for [1].

> Livestreaming, live-streaming, or live streaming is the [streaming](https://en.wikipedia.org/wiki/Streaming_media) of [video](https://en.wikipedia.org/wiki/Video) or [audio](https://en.wikipedia.org/wiki/Digital_audio) in [real time](https://en.wikipedia.org/wiki/Real-time_communication) or [near real time](https://en.wikipedia.org/wiki/Near_real_time). While often referred to simply as streaming, the real-time nature of livestreaming differentiates it from other non-[live broadcast](https://en.wikipedia.org/wiki/Live_broadcast) forms of streamed media such as [video-on-demand](https://en.wikipedia.org/wiki/Video-on-demand), [vlogs](https://en.wikipedia.org/wiki/Vlogs) and [video-sharing platforms](https://en.wikipedia.org/wiki/Video-sharing_platform) such as [YouTube](https://en.wikipedia.org/wiki/YouTube) and [Twitch](https://en.wikipedia.org/wiki/Twitch_(service)). 


[1] https://en.wikipedia.org/wiki/Live_streaming

docs/general/server/live-tv/internet-radio.md
```diff
@@ -1,6 +1,7 @@
 ---
-uid: server-media-internet-radio
-title: Internet radio
+uid: server-live-tv-internet-radio
+title: Internet Radio
+sidebar_position: 3
 ---
 
 # Internet radio
```
theguymadmax, line 7: Casing of "Internet radio" doesn't match the title and other pages in this category.

joshuaboniface (approved):

PR #1325, opened 2025-03-08
Fix dead link to Upgrade and Downgrade documentation

timminator:
Found a dead link on the site [Testing Jellyfin Server](https://jellyfin.org/docs/general/testing/server/) for the Upgrade & Downgrade documentation.

I've fixed this issue in this commit.


nielsvanvelzen (approved):
Thanks!

PR #1332, opened 2025-03-11
Update quick-connect.md

dbareis:
Correctly describe the quick connect process as the existing text was not understandable (at least by me).

### DIFF
--- a/docs/general/server/quick-connect.md
+++ b/docs/general/server/quick-connect.md
@@ -15,7 +15,13 @@ Settings > Dashboard > General > Enable Quick Connect on this server
 
 ## Using Quick Connect
 
-To sign in to a supported client, you have to enter the Quick Connect code in your user settings.
+The sign in page of the client has a "**Use Quick Connect**" button which will generate a 6 digit code, take note of the code.
+
+![image](https://user-images.githubusercontent.com/12074633/115973542-c99ed480-a523-11eb-9d61-17ccd628e123.png)
+
+
+To sign in to a supported client, you have to enter the Quick Connect code in your user settings where you are already logged on.
+Click your profile picture on the top right and the quick connect dialog is the second item on the settings page.
 
 Settings > Quick Connect
 
@@ -23,6 +29,3 @@ Settings > Quick Connect
 
 If the code is validated successfully, your new device will be signed in without entering your Jellyfin username or password on the new device.
 
-The client will generate a 6 digit code, which you have to enter in the already signed in client in your user settings.
-
-![image](https://user-images.githubusercontent.com/12074633/115973542-c99ed480-a523-11eb-9d61-17ccd628e123.png)


dbareis:
It also wouldn't hurt to get rid of the "got it" button (which you have to press even if you don't got it) and to add some text to the dialog that describes the process and or links to this documentation.

docs/general/server/quick-connect.md
```diff
@@ -15,14 +15,17 @@ Settings > Dashboard > General > Enable Quick Connect on this server
 
 ## Using Quick Connect
 
-To sign in to a supported client, you have to enter the Quick Connect code in your user settings.
+The sign in page of the client has a "**Use Quick Connect**" button which will generate a 6 digit code, take note of the code.
```
felix920506, line 18: Please mention that not all clients support this

docs/general/server/quick-connect.md
```diff
@@ -15,14 +15,18 @@ Settings > Dashboard > General > Enable Quick Connect on this server
...
```
felix920506, line 18: Some clients don't have that button and instead show the code by default. Please mention that.

felix920506 (changes requested):

docs/general/server/quick-connect.md
dbareis: I give up, you can make whatever changes you like, what I provided is a million times better than what is there, I have no experience of knowledge of other clients.

docs/general/server/quick-connect.md
```diff
@@ -5,7 +5,7 @@ title: Quick Connect
 
 # Quick Connect
 
-Starting with Jellyfin server version 10.7.0 and supported clients, you can use Quick Connect to sign into your account without the need of a password. You need to previously be logged into a supported client, like the default Jellyfin Web Client.
+Starting with Jellyfin server version 10.7.0 and on **clients that support Quick Connect**, you can use it to sign into your account without the need of a password. You need to previously be logged into a supported client, like the default Jellyfin Web Client.
```
BotBlake, line 8: I dont like how the explaination of what QuickConnect is, starts with "QuickConnect".
In my personal oppinion the order of the sentence seems more 'with the flow' if the order is switched.

Also, in the guide is mentioned too often that you will need another supported client, therefore I suggest to remove it here.

Additionaly, Jellyfin 10.7 is quite old now. Therefore I would suggest to remove this line.

```suggestion
On **clients that support it**, you can use Quick Connect to sign into your account without the need of a password.
```

docs/general/server/quick-connect.md
BotBlake, line 18: How about generalizing this to something like this:
```suggestion
The sign in page of supported clients will give you the option to "**Use Quick Connect**" for login. This can be through a dedicated button, or by directly showing the 6 digit code.
```

docs/general/server/quick-connect.md
```diff
@@ -15,14 +15,18 @@ Settings > Dashboard > General > Enable Quick Connect on this server
 
 ## Using Quick Connect
 
-To sign in to a supported client, you have to enter the Quick Connect code in your user settings.
+The sign in page of the client has a "**Use Quick Connect**" button which will generate a 6 digit code, take note of the code.
+
+![image](https://user-images.githubusercontent.com/12074633/115973542-c99ed480-a523-11eb-9d61-17ccd628e123.png)
+
+
+To sign in to a supported client, you have to enter the Quick Connect code in your user settings where you are already logged on.
```
BotBlake, line 23: Since it should be clear by now, that Quick Connect is about signing into a supported Client, I would not repeat that here again. Also it cannot hurt to go slow and explain every step of the way individualy.

I would suggest something like:
```suggestion
To proceed, you have to enter the Quick Connect code in another supported Client where you are already logged in.
```

docs/general/server/quick-connect.md
```diff
@@ -15,14 +15,18 @@ Settings > Dashboard > General > Enable Quick Connect on this server
...
+Click your profile picture on the top right and the quick connect dialog is the second item on the settings page.
```
BotBlake, line 24: Lets also mention here, that this is the case on jellyfin-web (and clients based on it), but not neccecarily on others.
```suggestion
To enter a code in Jellyfin-Web, click your profile picture on the top right and the quick connect dialog will be the second entry on the page.
```

docs/general/server/quick-connect.md
```diff
@@ -15,14 +15,18 @@ Settings > Dashboard > General > Enable Quick Connect on this server
...
 Settings > Quick Connect
 
 ![Quick Connect sign in](https://user-images.githubusercontent.com/12074633/115973526-aecc6000-a523-11eb-9ed6-59bee41bac7b.png)
 
 If the code is validated successfully, your new device will be signed in without entering your Jellyfin username or password on the new device.
 
-The client will generate a 6 digit code, which you have to enter in the already signed in client in your user settings.
-
-![Quick Connect error](https://user-images.githubusercontent.com/12074633/115973542-c99ed480-a523-11eb-9d61-17ccd628e123.png)
+The client that support Quick Connect will generate a 6 digit code, which you have to enter in the already signed in client in your user settings.
```
BotBlake, line 32: In my oppinion this has all already been said above. It is not nessecary to write a summary here.
This can simply be removed.

BotBlake:
This has been superseded by #1430 
Kind regards!

Issue #1009, opened 2024-05-26
Enable the possibility of select light mode


jesusfj710:
Hello there!

I love what you guys are doing in Jellyfin. I have spent a lot of time the latest weeks on the web reading the docs etc and I'm having some difficulties to use the web.

Honestly, the dark mode makes it very difficult for some people to read for a long time, and this is my problem. I can understand that this is something that depends on the individual, preferences, or even design decisions with aligning Jellyfin with a particular visual identity. All of this should never prevent a user from optionally choosing the light theme, even if only in the documentation.

Can we work on implementing both modes together?

Thanks :)

felix920506:
you are free to work on it and submit a PR

christmaswreath:
Bump

PR #1347, opened 2025-03-30
Light mode theme & accessibility fixes

dausm:
Addresses [issue 1009](https://github.com/jellyfin/jellyfin.org/issues/1009)

Audited the entire site for keyboard, screen reader, and contrast accessibility. Additionally fixed accessibility issues related to custom Jellyfin pages and theme. With these changes the site is WCAG AA compliant as much as it can be. Other issues are related to the Docusaurus framework so I'll address those with that project.

Changes:
- Enabled light and dark themes. Ussage is based on the users system preferences
- Ensured contrast requrirements are met for WCAG AA compliance no matter which mode the user select (light/dark)
- Ensured html markup is symantic
- Removed swiper/carousel since the package used wasn't accessible for keyboard or screen reader users. Since it's best to not hide content from users and carousels have falled out of fashion I update the homepage to display the content optimzally on all screen sizes.

Documentation updates are in [PR #1348](https://github.com/jellyfin/jellyfin.org/pull/1348)


erikbucik:
The WCAG 2 algorithm doesn't take into account how humans perceive color, the contrast calculation is based on just luminance. 
APCA does, so please check the color contrast again. Especially the buttons' text color.

felix920506 (changes requested):
Please make changes according to the comment from @erikbucik

dausm:
@erikbucik & @felix920506 I am working on this it's just taking a while since it's requiring updating a lot of colors.

erikbucik:
@dausm take your time. 

From what I've seen quickly, leave the social buttons' colors as they were and change the text over the main gradient to always be white. 
The text of the "Open source" badge in Downloads should also be white. 
And filled buttons do not need an outline:
- "Download" on Home
- "Find a client" on Home
- "Jetbrains" on Contribute

P.S.

Since you're changing colors, you can help me out by changing the background colors to:
- #05070A (dark mode)
- #FCFCFD (light mode)
These are from the new design I'm working on.

dausm:
@erikbucik It would have been really helpful to have this feedback a week ago, when you first mentioned the APCA guidelines. If I’d known sooner, I could’ve incorporated those suggestions.

From what I understand, most of the social button colors weren’t going to meet the APCA threshold — except for the purples — which I’ve since adjusted again so they all look more cohesive. That was only one of a dozen separate issues I had to address to meet APCA thresholds.

Regarding the button outlines, I included them to ensure compliance with WCAG standards. Since WCAG is currently the official web accessibility guideline and goes further than APCA in terms of usability and interactivity, I’m still prioritizing that as the baseline. I understand APCA is promising, but it’s still in the proposal phase.

After dedicating time to meeting APCA-related requirements, it’s a bit discouraging to change course again — it feels like the target keeps moving. Since a redesign might be on the horizon, maybe it makes sense to hold off on further color changes for now to avoid duplicating effort. What do you think?

Please take a look at my recent changes and get back to me.

felix920506:
What's up with the sizes of the cards being all out of whack here?
截圖 2025-04-09 晚上8 17 15

erikbucik:
> Since a redesign might be on the horizon, maybe it makes sense to hold off on further color changes for now to avoid duplicating effort. What do you think?

I couldn't agree with you more.

dausm:
Good catch on those cards! I'll get those cleaned up tomorrow.

dausm:
> What's up with the sizes of the cards being all out of whack here? 

Got this cleaned up and ran through the whole site again to find any other oddities.

erikbucik:
What did you do to the hover state colors? Why is everything so high contrast?

dausm:
I increased them for accessibility.

erikbucik:
Great! Now we need a seizure warning as well 

https://github.com/user-attachments/assets/b0ac8406-494b-4b92-9cc9-abeaddfb43e4

dausm:
@erikbucik I'm just trying to help here. If you have some constructive feedback I'd love to hear it. If you don't like thes hover state, that's fine, tell me what you want and I'll update the styles.

dausm:
@erikbucik  I see where you’re coming from and agree some of those are issues. Accessibility guidelines are somewhat open to interpretation and it sounds like our experiences with implementing them have differed. 

At this point I’m gonna close this pull request cause you have something in mind and I feel like I’m wasting my time without a design. Since you’re working on a redesign hopefully the accessibility issues can be fixed with that design. I’ll take responsibility for running too far with this and that it should have been a larger discussion up front. 

Accessibility goes far beyond color contrast and there are a number of issues on the site. Some of those are:
- Content announced out of order. On this site this was from a `flex-direction: reverse;` Just markup the content in the order you want it displayed.
- Automatic animations on screen with no way to pause or disable. This is from the carousel/swiper on the homescreen.
- Incorrect heading levels
- Content changes not being announced for screen readers.
- Plus more

Please keep these in mind as part of your redesign.

PR #1350, opened 2025-04-02
add X-Frame-Options

jameskimmel:
Not sure if this breaks any clients

### DIFF
--- a/docs/general/networking/nginx.md
+++ b/docs/general/networking/nginx.md
@@ -54,6 +54,8 @@ server {
 
     # Security / XSS Mitigation Headers
     add_header X-Content-Type-Options "nosniff";
+    # this could break some clients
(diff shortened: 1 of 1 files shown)


thornbill:
Yes it will 😉

jameskimmel:
Cheers. So maybe would could add it as optional.

nielsvanvelzen (changes requested):
This was removed in #1277 and I don't think we intend to add it back.

jameskimmel:
> This was removed in #1277 and I don't think we intend to add it back.

Too bad, because I agree with  @solidsnake1298 
_The goal isn't to simplify down to the absolute minimum for a working HTTPS proxy. I want this to be the minimum secure config that doesn't break 99.999% of user clients._

PR #1351, opened 2025-04-02
HSTS and SSL settings

jameskimmel:


nielsvanvelzen (changes requested):
We're using a minimal configuration in our documentation that highlights options needed for Jellyfin. Additional security on-top of the default security practices from a webserver is out of scope for our docs.

jameskimmel:
> We're using a minimal configuration in our documentation that highlights options needed for Jellyfin. Additional security on-top of the default security practices from a webserver is out of scope for our docs.

I would argue that HSTS and ssl stapling are default security practices nowadays. 
IMHO more important than the Permissions policy we also have in the NGINX docs.

PR #1358, opened 2025-04-08
update opus and ogg support status

felix920506:
Apple added support for OGG Vorbis and OGG Opus in iOS 18.4 / macOS 15.4


Shadowghost (approved):
Please fix the merge conflict

PR #1360, opened 2025-04-11
use url of 10.10.z branch in web-config.md

Viperinius:
The link to the stable default web config file is probably meant to point to the latest release, correct?

If this was otherwise intentionally left at 10.8.z, please disregard this PR


joshuaboniface (approved):

joshuaboniface:
Definitely not intentional, thanks!

PR #1328, opened 2025-03-09
Create passthrough.md

LordAnchemis:
Added instructions on unpriviledged LXC passthrough

PR #1366, opened 2025-04-15
make install instructions nicer

felix920506:
Trying to implement some ideas from https://github.com/jellyfin/jellyfin-meta/discussions/92
Open to feedback and/or major changes


docs/general/installation/automatic-installer.md
```diff
@@ -0,0 +1,93 @@
...
+2. Stop the currently running server either via the dashboard or using the menu bar icon.
+3. Drag the new `.app` package into the Applications folder and click yes to replace the files.
+4. Start the application.
+
+### Uninstalling on macOS
+
+1. Stop the currently running server either via the dashboard or using the application icon.
+2. Move the `.app` package to the trash.
+
+### Deleting Configuration on macOS
```
joshuaboniface, line 56: Not sure if we really need this section? If we do, we should add a similar one for Windows.

docs/general/installation/community.md
```diff
@@ -0,0 +1,139 @@
+---
+uid: installation-community
+title: Community Maintained Packages
+description: Community maintained packages.
+sidebar_position: 5
+---
+
+# Community Maintained Packages
+
+These packages are maintained by community packagers independent of Jellyfin for distributions not officially supported by Jellyfin. For these distributions, [container images](/docs/general/installation/container) are recommended.
```
joshuaboniface, line 10: >For these distributions, [container images](/docs/general/installation/container) are recommended.

This is where we get into where our line should be. Personally I think it should be a little less forceful, something like.

>We can't always support problems with community-maintained packages; if you encounter trouble with them, please contact their maintainers first. If you're not sure, we recommend using [container images](/docs/general/installation/container) on these systems instead.

joshuaboniface (commented):
Two thoughts.

docs/general/installation/automatic-installer.md
felix920506: I don't really know, this was directly copied from the original macOS docs. I think it would be useful to have but not sure if installation would be the best place for it to go.

felix920506: Removing for now, we can make a proper uninstall doc later

jhparker88:
As a first time user (or even occasional installer) you would expect a list of OS's from where you start (as most software installation divide there instructions by OS) but now you have a single page dedicated to automatic installations mixing (which means scrolling until you find your OS) all OS information merged into one page (for users that might never us Linux or Windows or macOS). I would recommend to keep Linux, Windows, macOS separated by there dedicated pages to reduce the need for searching relevant information on a specific OS.

joshuaboniface (approved):

joshuaboniface:
I do agree wiht @jhparker88 that the flow/list should be a little nicer starting with broad OS categories. But also this derives from a discussion regarding what we want to support at 1st party vs. 3rd party so perhaps that's better left for another PR.

felix920506:
@joshuaboniface I separated it out into the broad categories for the automatic installers.

jhparker88:
I would suggest moving the TrueNAS SCALE into the "Advanced" section given that it is similar to Synology and avoids cluttering the installation section.

felix920506:
@jhparker88 the reason Truenas scale isn't in advanced but Synology is that Synology users are actively causing support problems for us especially for hardware acceleration related problems since they use a kernel version so old even Intel doesn't support drivers for anymore. On top of that because they are such potatoes in terms of processing power, they are just a pain to support. 

Truenas doesn't have this problem.

jhparker88:
> for us especially for hardware acceleration related problems since they use a kernel version so old even Intel doesn't support drivers for anymore. On top of that because they are such potatoes in terms of processing power, they are just a pain to support.

Well, I saw that more under the premise of "user" reach which means most users (not having any numbers here) will choose from the four main (Linux, Windows, macOS, or Container) and any specialized installations can be found in Advanced. I was also under the impression that advanced means doing more than just pulling a container or download a file and press enter. For example, I would love to see a Promox section next to the Synonlogy and TrueNAS to incorporate #1328 into it but I'm digressing here.

PR #1369, opened 2025-04-16
Update nginx.md

jameskimmel:
Move redirect to the bottom and remove the unnecessary comment about uncommenting.

We could additionally add a comment about what this server block is doing.


nielsvanvelzen (approved):

nielsvanvelzen:
> Why move it to the bottom?

This is a common practice with Nginx configs, the http->https block is small and won't see much change so the "important" one that would see more frequent updates is on top

PR #1372, opened 2025-04-17
update firefox hevc status

felix920506:
https://github.com/jellyfin/jellyfin-web/pull/5876


felix920506:
sorry didn't notice there is an existing PR for this

PR #1377, opened 2025-04-19
remove quick start

felix920506:
The quickstart page doesn't really contain much useful info
The windows guide under it I think is more suited for somewhere else.


JPVenson (approved):

thornbill (changes requested):
This should be replaced with _something_. This is linked to in the startup wizard.

Also you **CANNOT** remove entries from the url tests. This will break links that we know are actively used. They at least need replaced with a redirect.

felix920506:
Noted, I will write the replacement for this and then remove this in that PR.

PR #1381, opened 2025-04-21
Warn about unauthenticated endpoints.

Tremoneck:
A lot of new users seems to connect their Server directly to the internet without knowing the implications of the unauthenticated endpoints. This commit adds a warning to the networking section about it. Includes a link to https://github.com/jellyfin/jellyfin/issues/5415 for more information for the users.

### DIFF
--- a/docs/general/networking/index.md
+++ b/docs/general/networking/index.md
@@ -7,6 +7,12 @@ title: Networking
 
 This section describes how to get basic connectivity to a Jellyfin server, and also some more advanced networking scenarios.
 
+:::caution
+
+Jellyfin should not be directly exposed to the internet. Several Endpoints are not authenticated including unauthorized streams. If you need connectivity outside of your homenetwork consider using a vpn or authentification in your reverse proxy. For more information visit [\#5415](https://github.com/jellyfin/jellyfin/issues/5415)
+
+:::
+
 ## Connectivity
 
 Many clients will automatically discover servers running on the same LAN and display them on login. If you are outside the network when you connect you can type in the complete IP address or domain name in the server field with the correct port to continue to the login page. You can find the default ports below to access the web frontend.


docs/general/networking/index.md
```diff
@@ -7,6 +7,12 @@ title: Networking
 
 This section describes how to get basic connectivity to a Jellyfin server, and also some more advanced networking scenarios.
 
+:::caution
+
+Jellyfin should not be directly exposed to the internet. Several Endpoints are not authenticated including unauthorized streams. If you need connectivity outside of your homenetwork consider using a vpn or authentification in your reverse proxy. For more information visit [\#5415](https://github.com/jellyfin/jellyfin/issues/5415)
+
+:::
+
```
felix920506, line 15: Yes jellyfin should not be directly exposed to the internet, but not for the reasons listed here.

A reverse proxy will do ABSOLUTELY NOTHING to the unauthenticated endpoints problem.

The real reason is that the built in HTTPS implementation is crap, and you shouldn't open services to the wider internet without HTTPS.

JPVenson:
Authentication via reverse proxy is only possible by web browsers limiting the usefullness.
I dont think a simple short note on that does not suffice as the topic is very complex.
Generally, even the unauthenticated endpoints are not just simple to missuse and require a lot of work. There are far better options to secure your jellyfin application like using wildcard DNS, TLS in general, RateLimiter, GeoBlocking and alike.

docs/general/networking/index.md
Tremoneck: For Authentication in the reverse proxy I meant using functionality like `auth_request` for nginx. In combination with an SSO solution the login happens before any request ever gets to jellyfin. 

If you think that isn't clear enough I will be more specific.

felix920506: This will break every single client that is not a browser.

Tremoneck: It's not ideal, so that was the reason to also suggest a VPN. I've removed the suggested fixes and only left the warning and the link to the issue. That way everyone can decide for themself about how to proceed without surprises.

PR #1384, opened 2025-04-22
fix broken title for post-install guide

felix920506:
when the index of the post install guide was removed from the PR the title also got removed


felix920506:
git problems

PR #1386, opened 2025-04-22
Add missing hyphen

felix920506:
Here is another pull request because venson couldn't wait


JPVenson (approved):

JPVenson:
technically correct, the best form of correct

PR #1390, opened 2025-04-26
Update traefik.md to v3.x.x

cedricwalter:
work in progress, do not review nor merged til marked ready

### DIFF
--- a/docs/general/post-install/networking/9_advanced/traefik.md
+++ b/docs/general/post-install/networking/9_advanced/traefik.md
@@ -1,19 +1,50 @@
 ---
 uid: network-reverse-proxy-traefik
-title: Traefik v2.x
+title: Traefik
 ---
-
-## Traefik v2.x
-
-[Traefik](https://traefik.io/) is a modern HTTP reverse proxy and load balancer that makes deploying microservices easy. Traefik integrates with your existing infrastructure components (ie: Docker) and generally configures itself dynamically as services are added or removed.
-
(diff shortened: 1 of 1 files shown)


docs/general/post-install/networking/9_advanced/traefik.md
```diff
@@ -22,22 +22,215 @@ Your router must local forward to the host running Traefik on port 80 (only if T
...
+          User-Agent: keep # log user agent strings
+
+# Entrypoints (ports) Traefik should listen to; here we define two: "http"
+# for unencrypted traffic, and "https" for SSL-encrypted traffic (port 443).
+entryPoints:
+  http:
+    address: ":80"
+    forwardedHeaders:
+      trustedIPs: &trustedIps
+        # Start of Cloudlare's public IP list
```
JPVenson, line 64: Please consult our rules on breaking the TOS of other services.
Tunneling jellyfin via CF is not permitted and we dont aid in breaking TOS

JPVenson:
Generally if you are still working on a PR, please mark it as DRAFT.
Anyway we just recently scaled back our documentation including third party tooling to the minimum working example because its a giant pain to maintain several extensive documentations so that is what we currently aim for. Our current example is compatiable with V3 and V2

docs/general/post-install/networking/9_advanced/traefik.md
cedricwalter: I will document how to exclude in cloudflare the Edge caching, since 2 years I run it like this and it is not against the TOS if properly configured

cedricwalter:
> Generally if you are still working on a PR, please mark it as DRAFT. Anyway we just recently scaled back our documentation including third party tooling to the minimum working example because its a giant pain to maintain several extensive documentations so that is what we currently aim for. Our current example is compatiable with V3 and V2

Im ready to maintain that Traefik page, i went away from nginx proxy manager as it is memory leaking and way too slow. There are migration steps to go from traefik 2 to 3, so the config are certainly not working. The version 2 should be marked as outdated. Can you please mark the PR as draft? I don’t find the switch to do so

docs/general/post-install/networking/9_advanced/traefik.md
JPVenson: if __any__ of jellyfin video traffic touches thier servers, its against their TOS. We have had users that got banned running JF though CF, i am sory but this is not up for discussion. If you dont tunnel though their servers you dont need to trust their headers. I am telling you plainly, any PR aiding in that will never get merged.

PR #1392, opened 2025-04-26
Update container.md

johnwblue:
Added lines for HTTPS and autodiscovery ports.


docs/general/installation/container.md
```diff
@@ -173,6 +173,9 @@ Steps to run Jellyfin using Podman are similar to the Docker steps.
     --label "io.containers.autoupdate=registry" \
     --name myjellyfin \
     --publish 8096:8096/tcp \
+    --publish 8920:8920/tcp \
+    --publish 1900:1900/udp \
```
felix920506, line 177: for DLNA to work, host networking is required, which makes forwarding individual ports redundant

johnwblue: The instructions provide two different examples of how to start a container.  The first using a docker compose file does include the host option but the second one does not.  It specifically only maps 8096 which seems to block everything else.

johnwblue: As seen here:

serveradmin@jellyfin:~/podman/jellyfin$ podman run -d -p 8096:8096/tcp -u $(id -u):$(id -g) --label "io.containers.autoupdate=registry" --name jellyfin --rm --userns keep-id --volume /home/serveradmin/podman/jellyfin/cache:/cache:Z --volume /home/serveradmin/podman/jellyfin/config:/config:Z --mount type=bind,source=/home/serveradmin/podman/jellyfin/media,destination=/media,ro=false,relabel=private docker.io/jellyfin/jellyfin:latest
852e320cdf4e897c1072f0d29bb6b3b817bb6c92562bd7eb05be644097d3c79d

serveradmin@jellyfin:~/podman/jellyfin$ podman port jellyfin
8096/tcp -> 0.0.0.0:8096

felix920506: I would prefer to have host networking be the default, since some features require it.

felix920506: The docker section uses host mode in the listed example command for reference

johnwblue: The current "podman run" example does not use host mode thus the need to open the autodiscovery ports.

johnwblue: https://docs.podman.io/en/stable/markdown/podman-run.1.html
"host: Do not create a network namespace, the container uses the host’s network. Note: **The host mode gives the container full access to local system services** such as D-bus and **is therefore considered insecure**."

BotBlake: @johnwblue 
The new seperated [DLNA Plugin Documentation](https://jellyfin.org/docs/general/post-install/networking/dlna#general) includes a warning that Docker will have to run in host mode for it to work,

No other services - appart from UPnP Port forwarding - make use of port 1900.
Maybe ad a smoll note that users who want to use the DLNA Plugin will have to use host-networking.

That way this hole topic can be solved easily.

Forwarding port 1900 to the docker-container is defenetly pointless though, as the required multicast messages will not be abled to reach the docker subnet regardless. So that should be removed regardless.

Kind regards 👍

felix920506: I don't think UPnP port forwarding needs that port and it got removed from the server anyways

BotBlake: @felix920506 If UPnP port forwarding uses any sort of discovery (which I would assume, so that it finds the Router and can communicate about what Ports to open. I could be wrong though.

Anyways, even better. And confirms my original point then 👌

docs/general/installation/container.md
```diff
@@ -173,6 +173,9 @@ Steps to run Jellyfin using Podman are similar to the Docker steps.
     --label "io.containers.autoupdate=registry" \
     --name myjellyfin \
     --publish 8096:8096/tcp \
+    --publish 8920:8920/tcp \
```
felix920506: Using jellyfin's internal HTTPS handling is not recommended and therefore should not be part of the example config.

felix920506:
This PR will be closed as stale by Oct. 7 if there is no further activity from the original author

BotBlake:
@felix920506 @johnwblue 
I switched the port forward to simply allow podman to use `host networking` too!
You where right! Podman was not using that before!

If we should no longer use host-networking mode at all should not be discussed here as well!

Thanks for the contribution!

nielsvanvelzen (approved):

PR #1398, opened 2025-05-06
TV Shows/Movies: Document theme media behaviour

mihawk90:
This is a followup to jellyfin/jellyfin-web#5714 to document those changes.

* User setting was undocumented
* Proper example boxes like in other sections
* New random media playback
* Theme Videos preferred over Theme Music
* Recommendations for theme media video format
* Per Season theme media was undocumented
* Movie theme media was entirely undocumented

### DIFF
--- a/docs/general/server/media/movies.md
+++ b/docs/general/server/media/movies.md
@@ -168,7 +168,7 @@ Supported filenames are:
 
 - `trailer`
 - `sample`
-- `theme` - Audio file of the theme song
+- `theme` - see [Theme Media](#theme-media) below
 
 ```txt
 Movies
@@ -302,3 +302,63 @@ Movie (2010)/extrafanart/fanart1.jpg, Movie (2010)/extrafanart/fanart2.jpg, _etc
 Example:
 
 Movie (2010)/logo.png
+
+## Other
+
+### Theme Media
+
+Theme media gives browsing your library a more audio-visual touch by playing theme songs and/or theme videos in the background while you're looking at your media listings.
+
+:::tip
+
+For these to play, users need to enable the option in their clients. In the WebUI and WebUI-based clients this is found in the User settings under Display > Library > Theme Songs, and Theme Videos, respectively.
+
+:::
+
+:::note
+
+Any type of media and codec that Jellyfin supports is allowed. Like with regular playback, the server will transcode if the client does not support the media.
+
+:::
+
+:::note
+
+In case both Theme Videos and Theme Music are found, Theme Videos will be preferred when enabled and Theme Music will not play.
+
+In the WebUI and WebUI-based clients, if there are multiple theme media found, they will be shuffled when opening the listing. [This cannot be changed](https://github.com/jellyfin/jellyfin-web/pull/5714).
+
+:::
+
+#### Videos
+
+- backdrops/\*
+
+Example:
+
+```txt
+Movies
+└── Best_Movie_Ever (2019)
+    └── backdrops
+        └── bluray-menu.ext
+```
+
+:::note
+
+Since transcoding can result in delay of theme media playback, Web-standard formats like WebM (VP9/Opus) are recommended for a smooth experience, since they generally direct play.
+
+:::
+
+#### Music
+
+- theme.ext
+- theme-music/\*
+
+Examples:
+
+```txt
+Movies
+└── Best_Movie_Ever (2019)
+    ├── theme.ext
+    └── theme-music
+        └── awesome-soundtrack-song.ext
+```
--- a/docs/general/server/media/shows.md
+++ b/docs/general/server/media/shows.md
@@ -262,21 +262,73 @@ Series Name (2010)/logo.png
 
 ## Other
 
-### Theme Videos
+### Theme Media
+
+Theme media gives browsing your library a more audio-visual touch by playing theme songs and/or theme videos in the background while you're looking at your media listings.
+
+:::tip
+
+For these to play, users need to enable the option in their clients. In the WebUI and WebUI-based clients this is found in the User settings under Display > Library > Theme Songs, and Theme Videos, respectively.
+
+:::
+
+:::note
+
+Any type of media and codec that Jellyfin supports is allowed. Like with regular playback, the server will transcode if the client does not support the media.
+
+:::
+
+:::note
+
+In case both Theme Videos and Theme Music are found, Theme Videos will be preferred when enabled and Theme Music will not play.
+
+In the WebUI and WebUI-based clients, if there are multiple theme media found, they will be shuffled when opening the listing. [This cannot be changed](https://github.com/jellyfin/jellyfin-web/pull/5714).
+
+:::
+
+#### Videos
 
 - backdrops/\*
 
 Example:
 
-Series Name (2010)/backdrops/S1Intro.ext
+```txt
+Shows
+└── Series Name (2010)
+    ├── backdrops
+    │   ├── S1Intro.ext
+    │   └── S2Intro.ext
+    ├── Season 01
+    │   └── backdrops
+    │       └── S1Intro.ext
+    └── Season 02
+        └── backdrops
+            └── S2Intro.ext
+```
 
-### Theme Music
+:::note
+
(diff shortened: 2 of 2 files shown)


mihawk90:
I meant to mark this as a draft because the web PR isn't released yet anyway 😅

mihawk90:
Rebased and Preview looks fine at first glance, unless you want the example boxes in the include file as well.

felix920506:
Please put the examples into _video-theme-media.md and only give one example with multiple types of theme media per library type

mihawk90:
Like this? :thinking: 

Or did you mean combining the `Videos` and `Music` header as well?

felix920506:
Now that I look at it closer, these should all be under the extras section. The examples part what I mean is to have a single example folder structure with both theme videos and music at the same time.

mihawk90:
OK, this should be it then.

I moved it under `Extras`, but I left it as a separate file, just moved the include.

Also fixed the linting error.

docs/general/server/media/_video-theme-media.md
```diff
@@ -0,0 +1,73 @@
...
+
+Any type of media and codec that Jellyfin supports is allowed. Like with regular playback, the server will transcode if the client does not support the media.
+
+:::
+
+:::note
+
+In case both Theme Videos and Theme Music are found, Theme Videos will be preferred when enabled and Theme Music will not play.
+
+In the WebUI and WebUI-based clients, if there are multiple theme media found, they will be shuffled when opening the listing. [This cannot be changed](https://github.com/jellyfin/jellyfin-web/pull/5714).
```
BotBlake: Personaly, I am not a fan of the link here.
```suggestion
In the WebUI and WebUI-based clients, if there are multiple theme media found, they will be shuffled when opening the listing. This cannot be changed.
```

docs/general/server/media/_video-theme-media.md
```diff
@@ -0,0 +1,73 @@
...
+
+:::
+
+#### Videos
+
+- backdrops/\*
```
BotBlake, line 24: These rules are super basic, I would suggest to not have individual chapters for Videos, Movies and Example - Instead this could easily fit into the text above.

docs/general/server/media/_video-theme-media.md
```diff
@@ -0,0 +1,73 @@
+<!-- markdownlint-disable MD041 -->
+
+### Theme Media
+
+Theme media gives browsing your library a more audio-visual touch by playing theme songs and/or theme videos in the background while you're looking at your media listings.
+
+:::tip
```
BotBlake, line 7: Try to keep the admonition count as low as possible. Having to many in the documentation will make them lose their effect on raising attention!
If something is not IMPORTANT, we dont realy want admonition blocks to be used.

BotBlake (changes requested):
Hi there!
Is this still in Draft? The WebPR has been merged.

Let me ad some feedback as well, while I am at it:
We generally try to keep the amount of admonitions as low as possible.
grafik
This is defenetly a few to many. If possible, please remove a few of those (actually, I think all of them can be removed here)

Additionaly, I think this is way too long to be an reusable element! If possible, try to squeeze down the chapter count. I added some feedback regarding that in the comments.

Thanks for the contribution!
Kind regards

mihawk90:
> Is this still in Draft? The WebPR has been merged.

Merged, but not released. This is in draft because it does not apply to the current released version of JF.

> This is defenetly a few to many. If possible, please remove a few of those (actually, I think all of them can be removed here)

I agree with that and that bothered me too while I wrote it, but I felt like all of these are important.

I guess the second one could be removed because the transcode is handled automatically and the transcoding thing under videos implies this anyway.

But the others I feel like will come up  with users because theme media is not enabled by default and the question of "how does multiple different theme media work" is valid too. IDK, maybe have that one just in the text?

I would certainly not remove the `Tip` boxes tho.

> Additionaly, I think this is way too long to be an reusable element! If possible, try to squeeze down the chapter count. I added some feedback regarding that in the comments.

I doubt this is going to be reused anywhere else, but that aside how would you even split these to still make sense when and if they are reused elsewhere?
I guess technically splitting by Music and Video would be an option, but why? Theme media only applies to video libraries anyway (as far as I know), so it's included in the Movies and TV Show docs, both of which allow both, so splitting them up into separate includes would just require to include all of them anyway.
The example box has already been unified as requested above so it would also have to be split up again or the separation wouldn't make sense.

I think *if* this is ever going to be included anywhere else (which I highly doubt), then it can still be split up at that time. But as of right now I don't think it makes sense to do so.

docs/general/server/media/_video-theme-media.md
mihawk90: I just followed the rest of the page to have it consistently layed out. I don't think having it in the text has any benefit, on the contrary I feel like it would get lost and/or overread. When I am looking for just the paths used I look for the list, not somewhere buried in the text.

docs/general/server/media/_video-theme-media.md
mihawk90: Fair enough, removed

docs/general/server/media/_video-theme-media.md
mihawk90: (re)moved some stuff, as per comment below

mihawk90:
Made some changes as per comments above:

* first "Note" removed
* changed second "Note" to a regular paragraph
* PR link removed

Also, two small things in addition:
* changed "Theme Music" to "Theme songs", as per the client setting
* swapped Music/Songs and Video chapters to space out the boxes further and for consistency within the text

felix920506:
@mihawk90 we track things that don't apply to the current version using the future-release tag. If the content is good please mark this ready to review.
Please also resolve the merge conflicts.

mihawk90:
I don't understand how there are merge conflicts now because there weren't when I last pushed, but the change that's conflicting is 267743c6ac53357e35e5b137f3c26de36a9afa7c from July... wadahel.

Anyway, rebased to master with some small changes:

* put the examples in a tabber similar to the metadata images section
* some wording adjustment in the WebM tip box to make it clear that any codec/container is supported
* removed the `theme` listing from `_video-external-extras.md` because they're in this dedicated section now and aren't extras in the sense the other 2 examples are anyway (i.e. not selectable in the library)

mihawk90:
I just stumbled on this while I was looking at the network requests for something unrelated...

```json
{
    "ThemeVideosResult": {
        "OwnerId": "f27caa37e5142225cceded48f6553502",
        "Items": [],
        "TotalRecordCount": 0,
        "StartIndex": 0
    },
    "ThemeSongsResult": {
        "OwnerId": "f27caa37e5142225cceded48f6553502",
        "Items": [],
        "TotalRecordCount": 0,
        "StartIndex": 0
    },
    "SoundtrackSongsResult": {
        "OwnerId": "00000000000000000000000000000000",
        "Items": [],
        "TotalRecordCount": 0,
        "StartIndex": 0
    }
}
```

`ThemeVideosResult` and `ThemeSongsResult` is obvious, but what exactly is `SoundtrackSongsResult`? What is the difference to the other two and why does it have a dummy(?) `OwnerId`?

Is this an emby leftover that noone knows about or something new that's undocumented as well?

It's mentioned in the result in the [API docs](https://api.jellyfin.org/#tag/Library/operation/GetThemeMedia) as well (although with a proper `OwnerId`), but no further details on it.

If anyone can enlighten me, I can add it to the PR as well.

PS:
Since 10.11 is out now the `future-release` label can be removed?

---

Also rebased to current master.

BotBlake:
Merged master to re-trigger CI.

docs/general/server/media/_video-external-extras.md
```diff
@@ -58,19 +58,17 @@ Supported filenames are:
 
 - `trailer`
 - `sample`
-- `theme` - Audio file of the theme song
 
 ```txt
 Best_Movie_Ever (2019)
 ├── Best_Movie_Ever (2019) - 1080P.mp4
-└── theme.mp3
+└── trailer.mp3
```
BotBlake:
```suggestion
└── trailer.mp4
```
I dont think movie Trailers make sense in an audio format

PR #1417, opened 2025-06-04
Updated HWA docs to match latest hardware and kernel changes

nyanmisaka:
- Updated HWA docs to match latest hardware and kernel changes
  - AMD RX 9000 encoding quality improvement
  - NVIDIA Blackwell speed and quality
  - Rockchip RK3576 speed and quality
  - Rockchip VPU load/utilization in BSP kernel 6.1.84+

### DIFF
--- a/docs/general/post-install/transcoding/hardware-acceleration/index.md
+++ b/docs/general/post-install/transcoding/hardware-acceleration/index.md
@@ -148,7 +148,7 @@ Dolby Vision (P5 & P8) to SDR tone-mapping is supported in Jellyfin 10.8 and req
 
 - VPP is prefered if both tone-mapping options are enabled.
 
-- Rockchip RKMPP currently only support HDR10 and HLG tone-mapping.
+- Rockchip RKMPP only support HDR10 and HLG tone-mapping in Jellyfin 10.9 and 10.10.
 
 :::
 
--- a/docs/general/post-install/transcoding/hardware-acceleration/index.md
+++ b/docs/general/post-install/transcoding/hardware-acceleration/index.md
@@ -148,7 +148,7 @@ Dolby Vision (P5 & P8) to SDR tone-mapping is supported in Jellyfin 10.8 and req
 
 - VPP is prefered if both tone-mapping options are enabled.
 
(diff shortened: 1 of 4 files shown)


docs/general/post-install/transcoding/hardware-acceleration/index.md
```diff
@@ -148,7 +148,7 @@ Dolby Vision (P5 & P8) to SDR tone-mapping is supported in Jellyfin 10.8 and req
 
 - VPP is prefered if both tone-mapping options are enabled.
 
-- Rockchip RKMPP currently only support HDR10 and HLG tone-mapping.
+- Rockchip RKMPP only support HDR10 and HLG tone-mapping in Jellyfin 10.9 and 10.10.
```
JPVenson:
```suggestion
- Rockchip RKMPP only support HDR10 and HLG tone-mapping in Jellyfin 10.9 or newer versions
```
or do we plan on removing them soon?

JPVenson (approved):

docs/general/post-install/transcoding/hardware-acceleration/index.md
nyanmisaka: 10.9 and 10.10 only have HDR10 and HLG tonemap. 10.11 adds DoVi tonemap.

So this is a comment on a limitation in older versions.

JPVenson: but the edit would still be correct then. Just that 10.11 has additional features.
Otherwise it reads that 10.11 would not support it.

nyanmisaka: Well indeed. How about this?
```
Rockchip RKMPP does not support Dolby Vision tone-mapping in Jellyfin 10.9 and 10.10.
```

This way there is no need to modify the docs when 10.11 is released.

JPVenson:
```suggestion
- Rockchip RKMPP does support Dolby Vision tone-mapping Jellyfin 10.11 and later versions.
```
But yours would also totally be fine.

PR #1428, opened 2025-06-24
Update nginx.md

hcebke:
Add note for WebOS app users experiencing error -27

I was hunting this error for hours before coming across https://github.com/jellyfin/jellyfin-webos/issues/258 where I found the solution: get rid of `frame-ancestors 'self';` in the CSP header.

### DIFF
--- a/docs/general/post-install/networking/8_reverse-proxy/nginx.md
+++ b/docs/general/post-install/networking/8_reverse-proxy/nginx.md
@@ -90,6 +90,11 @@ server {
 }
 ```
 
+## WebOS Jellyfin App Error -27
+
+For the Jellyfin app on WebOS (running on LG TVs) the `Content-Security-Policy` header needs to be modified slightly: Remove `frame-ancestors 'self';`.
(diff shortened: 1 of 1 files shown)


felix920506 (commented):
Should also remove it from the sample Nginx config

felix920506:
This PR will be closed as stale if there are no further comments from original Author by Oct 7

hcebke:
I didn't realize you were waiting for a comment. I also wasn't mad aware of any issue with my PR. What would you like me to comment on?

felix920506:
@hcebke if it is known to cause issues it should also be changed in the sample Nginx config.

hcebke:
Happy to do that, but a tiny caveat: I can only say for sure that the statement causes issues with the type of LG TV that I own. - I am assuming someone put this statement in there for a reason and I have no way of testing how its removal is going to affect Jellyfin running on other devices.

hcebke:
I've updated the nginx example config and removed the earlier paragraph about WebOS since it does no longer apply.

PR #1461, opened 2025-07-20
Add HDR section to the clients codec support page

felix920506:


docs/general/clients/codec-support.md
```diff
@@ -65,6 +65,16 @@ The goal is to Direct Play all media. This means the container, video, audio and
...
+### HDR Support
+
+HDR is only supported on a very limited range of devices:
+
+- iOS Devices with an HDR capable internal display can display HDR content when using a browser or the Jellyfin iOS App (not Swiftfin) as long as the codec is supported by the client. All of them support HEVC and VP9, while AV1 is supported on newer devices. HDR10, Dolby Vision and HLG are supported.
+
+- A Mac running macOS that supports HDR can play HDR content when using Safari on an HDR capable display as long as the source codec is supported. HDR10, Dolby Vision and HLG are supported. Please refer to [the list of Macs that support HDR](https://support.apple.com/en-us/102205) to check if your device is supported.
+
+- An Android TV device running the official Android TV client on a supported device. HDR format support depends on the specific device you are running.
+
```
JPVenson, line 87: maybe also add an "catch-all" phrase here regarding other devices

docs/general/clients/codec-support.md
```diff
@@ -65,6 +65,18 @@ The goal is to Direct Play all media. This means the container, video, audio and
...
+Devices not listed above do not support HDR on any official client. Any HDR content will have to be converted to SDR by the server for proper playback on these clients. For more info, please refer to our [tone-mapping documentation](/docs/general/post-install/transcoding/#hdr-to-sdr-tone-mapping)
```
BotBlake, line 86: Note that you are not actually listing DEVICES but groups of devices.
Lets say I would use the AndroidTV App on an Android Phone that supports HDR.
I dont know how to form that sentence better, I just wanted to add my 2 cents here.

docs/general/clients/codec-support.md
```diff
@@ -65,6 +65,18 @@ The goal is to Direct Play all media. This means the container, video, audio and
...
 
+### HDR Support
+
+HDR is only supported on a very limited range of devices:
+
+- iOS Devices with an HDR capable internal display can display HDR content when using a browser or the Jellyfin iOS App (not Swiftfin) as long as the codec is supported by the client. All of them support HEVC and VP9, while AV1 is supported on newer devices. HDR10, Dolby Vision and HLG are supported.
+
+- A Mac running macOS that supports HDR can play HDR content when using Safari on an HDR capable display as long as the source codec is supported. HDR10, Dolby Vision and HLG are supported. Please refer to [the list of Macs that support HDR](https://support.apple.com/en-us/102205) to check if your device is supported.
+
+- An Android TV device running the official Android TV client on a supported device. HDR format support depends on the specific device you are running.
```
BotBlake: > A DEVICE running the app on a DEVICE

This is dual usage of the word "device".
Also additionally, what does "supported device" mean in this instance? Does it need to support HDR, or the AndroidTV App itself?

docs/general/clients/codec-support.md
```diff
@@ -65,6 +65,18 @@ The goal is to Direct Play all media. This means the container, video, audio and
 <br />
 <sup>2</sup><a href="https://en.wikipedia.org/wiki/MPEG-4_Part_17">MPEG-4 Part 17: MP4TT Subtitles</a>
 
+### HDR Support
+
+HDR is only supported on a very limited range of devices:
+
+- iOS Devices with an HDR capable internal display can display HDR content when using a browser or the Jellyfin iOS App (not Swiftfin) as long as the codec is supported by the client. All of them support HEVC and VP9, while AV1 is supported on newer devices. HDR10, Dolby Vision and HLG are supported.
+
+- A Mac running macOS that supports HDR can play HDR content when using Safari on an HDR capable display as long as the source codec is supported. HDR10, Dolby Vision and HLG are supported. Please refer to [the list of Macs that support HDR](https://support.apple.com/en-us/102205) to check if your device is supported.
```
BotBlake: If we need to specify that it is a device running macOS, does it realy have to be a Mac?
I get that people COULD POTENTIALY run Windows or Linux on their macs, but I dont think that is worth a mention in the docs here. If it realy has to, why cant it simply be ANY device running macOS?

Additionally, while I do like the Idea of listing the macOS supported Codecs here again, I think it gives a level of redundancy to the codec support table above. Just keep in mind that they will have to be updated here too when the supported codecs change over time.

docs/general/clients/codec-support.md
```diff
@@ -65,6 +65,18 @@ The goal is to Direct Play all media. This means the container, video, audio and
 <br />
 <sup>2</sup><a href="https://en.wikipedia.org/wiki/MPEG-4_Part_17">MPEG-4 Part 17: MP4TT Subtitles</a>
 
+### HDR Support
+
+HDR is only supported on a very limited range of devices:
+
+- iOS Devices with an HDR capable internal display can display HDR content when using a browser or the Jellyfin iOS App (not Swiftfin) as long as the codec is supported by the client. All of them support HEVC and VP9, while AV1 is supported on newer devices. HDR10, Dolby Vision and HLG are supported.
```
BotBlake, line 72: Do the Devices Running iOS have to be iPhones?
Since we specifically mention it below for the Macs running macOS, I think we should find a scheme and stick to it :sweat_smile:

BotBlake (commented):
I understand this as if there is no way to get HDR to function with official clients on Windows/ Linux Desktop clients.
Is that actually the case?

Additionally I added a bunch of nitpicky comments.

docs/general/clients/codec-support.md
felix920506: How do you even run iOS on non Apple hardware to begin with?

docs/general/clients/codec-support.md
felix920506: It has to be macOS and using Safari. Also, running macOS on non Apple hardware is very much against its license.

docs/general/clients/codec-support.md
felix920506: If you are using Android TV app on Android that's obviously not intended usage so do we really need to care?

docs/general/clients/codec-support.md
BotBlake: I would suggest getting rid of the first "mac" then 🗿 
"A Mac running macOS" sounds weird imo

docs/general/clients/codec-support.md
BotBlake: Great, get rid of the Mac above then 🗿

felix920506:
@BotBlake There is no way to get HDR on any Jellyfin desktop client (except maybe Kodi I'm not sure) on any platform. To get HDR on macOS you have to use Safari

docs/general/clients/codec-support.md
BotBlake: Its fine, I just wanted to mention that "devices not listed above" indicates everything that is not a Mac, AndroidTV or an iPhone 😅

BotBlake:
My Samsung S25 also seems to be capable to play back HDR Content just fine.
grafik

This is direct playing without visual issues.
I tested with both WebPlayer (default) and the integrated player (ExoPlayer).

Therefore I think it is save to say that the Android-Mobile-App also supports HDR

docs/general/clients/codec-support.md
```diff
@@ -65,6 +65,24 @@ The goal is to Direct Play all media. This means the container, video, audio and
...
+
+- Android devices running either the official Android App or using Chrome can display HDR content if the device supports HDR and the source format. Format support depends on device.
+
+- Macs running macOS that supports HDR can play HDR content when using Safari, Chrome or Firefox on an HDR capable display as long as the source codec is supported. Safari has the most complete support, including HDR10, Dolby Vision (including P5) and HLG. Chrome supports HDR10 and Dolby Vision (except P5), while Firefox only supports HDR10 on VP9 and AV1 content. Please refer to [the list of Macs that support HDR](https://support.apple.com/en-us/102205) to check if your device is supported.
```
gnattu: No chrome has no Dolby Vision. It is just playing the fallback layer.

docs/general/clients/codec-support.md
```diff
@@ -65,6 +65,24 @@ The goal is to Direct Play all media. This means the container, video, audio and
...
+
+- Windows PCs with an HDR display with HDR enabled in the settings using Edge or Chrome. Dolby Vision requires the Dolby Vision addon from the Microsoft Store to be installed.
```
gnattu: DV extension only works in edge and edge has generally better HDR support on windows, like having a better client-side tone mapping implementation. A lot of users have reported that edge produces much better tone mapped playback than chrome.

docs/general/clients/codec-support.md
```diff
@@ -65,6 +65,24 @@ The goal is to Direct Play all media. This means the container, video, audio and
...
+
+- Windows PCs with an HDR display with HDR enabled in the settings using Edge or Chrome. Dolby Vision only works in Edge and requires the Dolby Vision addon from the Microsoft Store to be installed. Edge generally produces better colors than Chrome when playing HDR content.
+
+- Android TV devices with HDR support running the official Android TV App. HDR format support depends on the specific device you are running.
+
+- WebOS devices (Newer LG Smart TVs) running the official WebOS client supports HDR. However, [mkv containers are known to cause issues](https://github.com/jellyfin/jellyfin-web/issues/4678). Please enable the force remux mkv to mp4 option in the client settings as a workaround.
+
+Devices not listed above do not support HDR on any official client. Any HDR content will have to be converted to SDR by the server for proper playback on these clients. For more info, please refer to our [tone-mapping documentation](/docs/general/post-install/transcoding/#hdr-to-sdr-tone-mapping)
```
gnattu, line 86: I *think* at least some of the Chromium based browsers on Linux will do client side tone mapping, although sometimes the tone mapping looks very bad. If the browser or the client is doing client side tone mapping, there is no server-side transcoding involved.

felix920506: A lot of them are going to look bad anyway so I don't think we should include that just yet.

docs/general/clients/codec-support.md
```diff
@@ -65,6 +65,26 @@ The goal is to Direct Play all media. This means the container, video, audio and
...
 
+### HDR Support
+
+HDR is only supported on a very limited range of devices:
+
+- iOS Devices with an HDR capable internal display can display HDR content when using a browser or the Jellyfin iOS App (not Swiftfin) as long as the codec is supported by the client. All of them support HEVC and VP9, while AV1 is supported on newer devices. HDR10, Dolby Vision and HLG are supported.
+
+- Android devices running either the official Android App or using Chrome can display HDR content if the device supports HDR and the source format. Format support depends on device.
+
+- Macs running macOS that supports HDR can play HDR content when using Safari, Chrome or Firefox on an HDR capable display as long as the source codec is supported. Safari has the most complete support, including HDR10, Dolby Vision (including P5) and HLG. Chrome and Firefox only support HDR10. Chrome supports HDR10 on HEVC, VP9 and AV1. Firefox only supports HDR on VP9 and AV1 content. Please refer to [the list of Macs that support HDR](https://support.apple.com/en-us/102205) to check if your device is supported. On non-Apple displays, other HDR formats will be tone-mapped to HDR10 by the client device.
```
BotBlake:
```suggestion
- Macs running macOS that support HDR can play HDR content when using Safari, Chrome or Firefox on an HDR capable display as long as the source codec is supported. Safari has the most complete support, including HDR10, Dolby Vision (including P5) and HLG. Chrome and Firefox only support HDR10. Chrome supports HDR10 on HEVC, VP9 and AV1. Firefox only supports HDR on VP9 and AV1 content. Please refer to [the list of Macs that support HDR](https://support.apple.com/en-us/102205) to check if your device is supported. On non-Apple displays, other HDR formats will be tone-mapped to HDR10 by the client device.
```

PR #1468, opened 2025-07-25
Fix quick start guide link redirect for iOS app

thornbill:
Fixes the 404 in https://github.com/jellyfin/jellyfin-expo/issues/658

We still need a better replacement with similar content to the original quick start guide, but at least this will resolve the 404 for now. The app will also need updated to directly link to this new page when it is available.

### DIFF
--- a/redirects.ts
+++ b/redirects.ts
@@ -114,8 +114,13 @@ const redirects: ClientRedirects.Options['redirects'] = [
   },
   // quick start replaced with post install
   {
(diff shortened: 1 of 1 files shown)


nielsvanvelzen (approved):

nielsvanvelzen:
@thornbill build is failing 🙃

thornbill:
Stupid trailing slash 😅

PR #1477, opened 2025-07-30
Use tabs for container installation instructions

felix920506:


nielsvanvelzen (changes requested):
It would be better to move the actual tab contents to their own components (files) to avoid mixing React code & actual documentation content.

PR #1481, opened 2025-08-03
add jellyfin-documentation to contact page

BotBlake:


src/components/contact/MatrixCard.tsx
```diff
@@ -68,6 +68,9 @@ const MatrixCard = () => {
               <li>
                 <a href='https://matrix.to/#/#jellyfin-hwa-tool:matrix.org'>#jellyfin-hwa-tool</a>
               </li>
+              <li>
+                <a href='https://matrix.to/#/#jellyfin-documentation:matrix.org'>#jellyfin-documentation</a>
+              </li>
```
nielsvanvelzen: Should be on top as the documentation is more accessible for users compared to other development, we also put in on top in the discord list

PR #1484, opened 2025-08-03
add Jellyfin xbox

BotBlake:
Since `Jellyfin-uwp` is now officially `Jellyfin for Xbox` this needs to be updated in the documentation.
Ironically I was not abled to find jellyfin-uwp being mentioned there anyways :sweat_smile: 

I believe Jellyfin for Xbox officially deserves to be on the Clients list too.

PS: Yes I know I did two things in one PR, please don't roast me for that. Imo they belong together.


JPVenson (approved):

src/components/common/PlatformIcon.tsx
```diff
@@ -117,6 +117,9 @@ const PlatformIcon = ({
           <path d='M0 93.7l183.6-25.3v177.4H0V93.7zm0 324.6l183.6 25.3V268.4H0v149.9zm203.8 28L448 480V268.4H203.8v177.9zm0-380.6v180.1H448V32L203.8 65.7z' />
         </svg>
       );
+    case Platform.Xbox:
+      // Not available in simple-icons because it was removed due to Microsoft (https://github.com
```
TimGels: Is this comment cut off?

BotBlake: fixed :moyai:

src/components/common/PlatformIcon.tsx
```diff
@@ -118,6 +118,10 @@ const PlatformIcon = ({
         </svg>
       );
 
+    case Platform.Xbox:
+      // Not available in simple-icons because it was removed due to Microsoft (https://github.com/simple-icons/simple-icons/issues/11236)
+      return <Icon path={mdiMonitor} size={`${size}px`} className={className} />;
```
nielsvanvelzen: Should use an actual Xbox icon, e.g. https://fontawesome.com/icons/xbox (see windows icon as example)

src/components/common/PlatformIcon.tsx
```diff
@@ -118,6 +118,10 @@ const PlatformIcon = ({
         </svg>
       );
 
+    case Platform.Xbox:
+      // Not available in simple-icons because it was removed due to Microsoft (https://github.com/simple-icons/simple-icons/issues/11236)
```
nielsvanvelzen: Irrelevant comment, we never used that icon to begin with.

src/data/clients.ts
```diff
@@ -1000,5 +1000,34 @@ export const Clients: Array<Client> = [
...
+      {
+        id: 'microsoft-store',
+        name: 'Microsoft Store',
+        url: 'https://apps.microsoft.com/detail/9P2DRTG62QF8'
+      },
+      {
+        id: 'gh-downloads',
+        name: 'GitHub Downloads',
+        url: 'https://github.com/jellyfin/jellyfin-xbox/releases'
+      }
```
nielsvanvelzen:
```suggestion
      }
```

We don't list the Github page when a client can be downloaded using official store(s), the GitHub url is alaos linked on the left already.

src/data/clients.ts
```diff
@@ -1000,5 +1000,34 @@ export const Clients: Array<Client> = [
         url: 'https://monk-studio.com/finer'
       }
     ]
+  },
+  {
```
nielsvanvelzen, line 1004: Official clients should be ordered above third-party clients

src/data/clients.ts
```diff
@@ -1000,5 +1000,34 @@ export const Clients: Array<Client> = [
...
+      }
+    ],
+    secondaryLinks: [
+      {
+        id: 'github',
+        name: 'GitHub',
+        url: 'https://github.com/jellyfin/jellyfin-xbox'
+      }
+    ],
+    recommended: true
```
nielsvanvelzen: I'm not sure if we should recommend it yet or if we should wait for the next release first. As to my knowledge the current version of the app is barely usable.

nielsvanvelzen (changes requested):

src/data/clients.ts
TimGels: We plan to release a new Xbox for Jellyfin version sometime after the Jellyfin server 10.11.0 release. In that version controller navigation has been overhauled and many major crashes have been resolved. I think it is fine if the "recommendation" sticker is only added after that version is released.

TimGels: @JPVenson what is your view on this?

JPVenson: Both valid points.
I am thinking of temporarly adding cursor mode back in and releasing a version that also has a banner to upgrade to 10.11

TimGels: @BotBlake could you remove the recommendation for now.

JPVenson: fine by me

src/components/common/PlatformIcon.tsx
BotBlake: The comment is there to explain on why we need to do additional shenanigans to get the xbox logo instead of using `simple-icons` for it.

src/components/common/PlatformIcon.tsx
BotBlake: How I see it, `simple-icons` was essentially forced by microsoft to remove the logos.
I assume there is a greater reason behind this - the Microsoft Branding Guidelines also dont allow use of their Logos.

Are you sure we are allowed to use this Logo?

src/data/clients.ts
BotBlake: I cannot do this without restructuring the entire `clients.ts` file.
`mopidy`, `jellyfin-androidtv`, `jellyfin-roku`, `jellyfin-webos`, ...

All are mixed in between `ThirdParty` Clients.

src/components/common/PlatformIcon.tsx
BotBlake: @coderabbitai can you elaborate?

src/components/common/PlatformIcon.tsx
nielsvanvelzen: At the time of adding this platform there is no xbox icon at all in the icon library we use so it's unnecessary and useless information. Similar comments are added on other platforms because we changed them when they were removed from the library we used so it explains why we took the icon from their repo and added it directly.

src/components/common/PlatformIcon.tsx
BotBlake: @nielsvanvelzen Just say yes and I dont mind🗿 
I just want somone saying "yes we can use it" and we good 🗿

nielsvanvelzen: Well I asked for it so yes?

src/components/common/PlatformIcon.tsx
BotBlake: Ultimately I don't mind, so will remove.
Allthough imho it should also be removed on the windows icon then.

Cause if I want to know why something was replaced, I will check the commit/ PR 🗿

BotBlake:
To quote the other PR which modifies the `clients.ts` file:

> Done.
> 
> @nielsvanvelzen ?

nielsvanvelzen (approved):

PR #1486, opened 2025-08-05
Clarify bash script setup process

linkandzelda91:
this should make it more clear that the given bash script will work out of box without modification, and also supplies the user and group name for the chown command. 

the $USER variable should already exist, but in trying to figure out a clear way to have the id command just spit out the username I found someone complaining that their distro doesn't set the env variable, so I figure it's best to just set it anyway.

### DIFF
--- a/docs/general/installation/advanced/manual.md
+++ b/docs/general/installation/advanced/manual.md
@@ -162,13 +162,19 @@ sudo apt install -f
 
 ### Running Jellyfin
 
-Due to the number of command line options that must be passed on to the Jellyfin binary, it is easiest to create a small script to run Jellyfin.
+Due to the number of [command line options](https://jellyfin.org/docs/general/administration/configuration/#command-line-options) that must be passed on to the Jellyfin binary, it is easiest to create a small script to run Jellyfin.
 
 ```sh
 sudoedit jellyfin.sh
 ```
 
-Then paste the following commands and modify as needed.
+Then paste the following commands.
+
+:::note
+
+Optionally, to make changes to the directories used or specify extra [command line options](https://jellyfin.org/docs/general/administration/configuration/#command-line-options), add or modify the script below as needed.
+
+:::
 
 ```sh
 #!/bin/bash
--- a/docs/general/installation/advanced/manual.md
+++ b/docs/general/installation/advanced/manual.md
@@ -168,13 +168,7 @@ Due to the number of [command line options](https://jellyfin.org/docs/general/ad
 sudoedit jellyfin.sh
 ```
 
-Then paste the following commands.
-
-:::note
-
-Optionally, to make changes to the directories used or specify extra [command line options](https://jellyfin.org/docs/general/administration/configuration/#command-line-options), add or modify the script below as needed.
-
-:::
+Then paste the following commands, optionally changing paths and options as needed.
 
 ```sh
 #!/bin/bash
--- a/docs/general/installation/advanced/manual.md
+++ b/docs/general/installation/advanced/manual.md
@@ -193,7 +193,9 @@ Assuming you desire Jellyfin to run as a non-root user, `chmod` all files and di
 Also make the startup script above executable.
 
 ```sh
-sudo chown -R user:group *
+USER=$(id --name --user)
+GROUP=$(id --name --group)
+sudo chown -R $USER:$GROUP *
 sudo chmod u+x jellyfin.sh
 ```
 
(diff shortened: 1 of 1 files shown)


linkandzelda91:
did a quick test:
```
gdewitt@jftest:/opt/jellyfin$ ls -lh * | tail
-rwxr--r-- 1 root root 2.8M Sep  6  2024 libse.dll
-rwxr--r-- 1 root root  59K Feb 12 04:44 mscorlib.dll
-rwxr--r-- 1 root root  99K Feb 12 04:44 netstandard.dll
-rwxr--r-- 1 root root  93K Sep 27  2024 prometheus-net.DotNetRuntime.dll
drwxr-xr-x 3 root root    3 Apr  5 19:25 wwwroot
-rwxr--r-- 1 root root  55K Aug 23  2024 zlib.net.dll

log:
total 9.0K
-rw-r--r-- 1 root root 23K Aug  5 03:13 log_20250805.log
gdewitt@jftest:/opt/jellyfin$ USER=$(id --name --user)
GROUP=$(id --name --group)
sudo chown -R $USER:$GROUP *
sudo chmod u+x jellyfin.sh
gdewitt@jftest:/opt/jellyfin$ ls -lh * | tail
-rwxr--r-- 1 gdewitt gdewitt 2.8M Sep  6  2024 libse.dll
-rwxr--r-- 1 gdewitt gdewitt  59K Feb 12 04:44 mscorlib.dll
-rwxr--r-- 1 gdewitt gdewitt  99K Feb 12 04:44 netstandard.dll
-rwxr--r-- 1 gdewitt gdewitt  93K Sep 27  2024 prometheus-net.DotNetRuntime.dll
drwxr-xr-x 3 gdewitt gdewitt    3 Apr  5 19:25 wwwroot
-rwxr--r-- 1 gdewitt gdewitt  55K Aug 23  2024 zlib.net.dll

log:
total 9.0K
-rw-r--r-- 1 gdewitt gdewitt 23K Aug  5 03:13 log_20250805.log
gdewitt@jftest:/opt/jellyfin$ echo $USER $GROUP
gdewitt gdewitt
gdewitt@jftest:/opt/jellyfin$ 
```

docs/general/installation/advanced/manual.md
```diff
@@ -162,13 +162,19 @@ sudo apt install -f
...
+Due to the number of [command line options](https://jellyfin.org/docs/general/administration/configuration/#command-line-options) that must be passed on to the Jellyfin binary, it is easiest to create a small script to run Jellyfin.
 
 ```sh
 sudoedit jellyfin.sh
 ```
 
-Then paste the following commands and modify as needed.
+Then paste the following commands.
+
+:::note
```
felix920506: Do we really need admonition?

BotBlake: I agree with felix on this.
We try to keep the "note"/"warning" blocks as low as possible, so that they stand out more IF there actually is something important to note.

Also imo there is no reason to link to the same page twice within 10 lines.

linkandzelda91: updated, is better?

docs/general/installation/advanced/manual.md
```diff
@@ -187,7 +187,9 @@ Assuming you desire Jellyfin to run as a non-root user, `chmod` all files and di
 Also make the startup script above executable.
 
 ```sh
-sudo chown -R user:group *
+USER=$(id --name --user)
```
BotBlake, line 190: I dont REALY see the point in using variables here.
Most users will probably be fine copy-pasting this as individual commands onto their terminal.

However I am not strictly against it either.
Please elaborate on that. Why do you think are the variables a useful addition?

linkandzelda91: My intent is it only to make it more blind copy and paste friendly: set the vars to the current user/primary group, then run the command with them added. Currently they would need to substitute manually, of course.

I'm not super insistent that it's better to change it either, I do think it's good for someone following any setup guide to have to stop and think occasionally instead of copy and paste everything, plus this is under the "manual installation" section for a reason. Then again, accessibility is a good thing so 🤷 

These two changes originally spawned from someone in the troubleshooting channel who was unsure:
1. They thought the wording of the bash script setup paragraph meant they were supposed to change contents, not that it was optional and would run fine unmodified.
2. They asked what user they should be setting the permissions to.

BotBlake (approved):
Why not.

PR #1489, opened 2025-08-05
Use tabs for HWA setup per GPU vendor

linkandzelda91:
trying to follow the example set here:
https://github.com/jellyfin/jellyfin.org/pull/1477


linkandzelda91:
yay building now, but very broken looking :(

linkandzelda91:
table method is better, closing.

PR #1491, opened 2025-08-07
sort clients.ts

BotBlake:
1. splitting into Official and third Party (separate lists)
2. enforce alphabetical sorting in the list
3. removed non-existant `iPadOS` Platform from `streamyfin`

Proposal for future changes: 
Cleanup of the clients file!
Some of them appear to be unmaintained and/or no longer compatible with the latest Version of Jellyfin.
If that is the case, these clients should be removed.

### DIFF
--- a/src/data/clients.ts
+++ b/src/data/clients.ts
@@ -728,22 +728,6 @@ export const thirdPartyClients: Array<Client> = [
     ]
   },
   {
-    id: 'jellyfin-plugin-for-volumio',
-    name: 'Jellyfin Plugin for Volumio',
-    description: 'A Volumio plugin for playing audio from one or more Jellyfin servers.',
-    clientType: ClientType.ThirdParty,
-    deviceTypes: [],
-    licenseType: LicenseType.OpenSource,
-    platforms: [],
-    primaryLinks: [
-      {
-        id: 'github',
-        name: 'GitHub',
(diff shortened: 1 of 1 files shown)


BotBlake:
I also found that some third party client dont "id" their website as `website`.

Instead, some use `browser`, others literally {client-name} for the `id` of the website!
If not wanted in a seperate PR, I would like to change the id for client websites to be `website` in all of them.

BotBlake:
I have to disagree.
If we want to have a specific order for the clients, then we should have a value to represent that.

I do agree that alphabetical ordering in the `Recommended` view is not ideal.
In my opinion, the recommended clients should be sorted by their “recommendation worthiness.”

If I enter the "All Clients" view, I don’t really expect any particular ordering.
In that view, some sort of ordering still makes sense.

I also think we should allow more filtering options (although I’m not sure how much Docusaurus can handle in that regard).

I suggest adding a sorting_id to the clients and ordering them accordingly in the `Recommended` view.

anultravioletaurora:
My two cents!

> In my opinion, the recommended clients should be sorted by their “recommendation worthiness.”

Totally agree. I think from an outsider looking in, it would be nice to have the first party apps highlighted at the top in some sort of predictable order

BotBlake:
Closed due to differing opinions regarding the sorting methodology and overall maintenance approach for the clients file.

PR #1494, opened 2025-08-08
Update intel.md

bpoulliot:
Fix relative links to known-issues page


BotBlake:
Hi there @bpoulliot 
The documentation SHOULD not require avsolute paths.

Allthough, I made sure to test this.
For me the changed URLs work just fine on the current jellyfin.org page.
Please try again or elaborate on the exact issue you are having with them.

Kind regards and sorry for the late reply.

felix920506 (approved):

felix920506:
There have been some relative links on the HWA pages that sometimes work and sometimes don't for a while now, no one has been able to reliably catch when they do/don't work so it would be best to make the problematic ones permanent for now.

PR #1374, opened 2025-04-18
rework video content docs

felix920506:
Combine documentation for shows, movies and music videos.

If they were to be different pages and all entirely complete, there would be about 80% of the content duplicated between pages

## Summary by CodeRabbit

- **Documentation**
  - Added new detailed documentation on external images, 3D videos, extras, external streams, multipart and multiversion videos, and metadata providers.
  - Restructured and modularized documentation for movies, TV shows, and music videos to improve clarity and maintainability.
  - Deprecated the "Mixed Movies and Shows" library type with a recommendation against its use.
  - Updated subtitle documentation redirect rules to reflect new documentation locations.

PR #1495, opened 2025-08-09
use tabs for metadata docs that covers multiple library types

felix920506:
move to using tabs instead of having everything listed together on the same page.
The default tab will be selected based on the current page that is being viewed


linkandzelda91:
should movies and shows be unified into a "dedicated movies or shows" page and go all in on the tabs for each section? feels weird that they dup on content for eachother.

image

felix920506:
@linkandzelda91 see discussion here for why it's separate pages. https://github.com/jellyfin/jellyfin.org/pull/1374#issuecomment-2833654929

nielsvanvelzen (dismissed):
Make sure  to use the query string feature of the tabs component so documentation stays easily sharable.

BotBlake (changes requested):
As previously discussed in the documentation room, it should be more clear what is part of the Tab.

I have created a mockup by modifying the css in the browser direcly.
grafik


I dont KNOW if you can modify the styling directly, but if you can, here is the css snippet I used to make the borders there.
```css
.tabs-container.tabList__CuJ {
  border: 2px solid rgba(255, 255, 255, 0.2);
  border-radius: 8px;
  padding: 16px;
  box-shadow: 0 0 8px rgba(255, 255, 255, 0.2); /* fancy glowy shizz that I realy like */
}
```

felix920506:
blocked by #1564

BotBlake (commented):
I think this has become unblocked by https://github.com/jellyfin/jellyfin.org/pull/1598
I know the better CSS solution is still in the cooking https://github.com/jellyfin/jellyfin.org/pull/1564 - but I dont see a reason not to get this merged already.

@felix920506 What do you think?

felix920506:
@BotBlake if taking the position of "don't let perfect be the enemy of good" I think this should be good to merge after a few minor touchups required by changes elsewhere on the repo.

I have resolved the merge conflicts to start

felix920506:
I think this should be good enough for now

BotBlake:
Merge to re-trigger CI jobs

BotBlake (approved):
This makes sense and indeed is good to merge.
Sorry for the horrible review time, I simply forgot about this!

PR #1496, opened 2025-08-10
Add WizdomSubs Downloader to Plugins

DeDuplicate:
add WizdomSubs Downloader Addon


JPVenson (changes requested):
Please use the existing Format for 3rd Party plugins

nielsvanvelzen (approved):

PR #1501, opened 2025-08-12
Auto detect os for server downloads

timreichen:


package-lock.json
```diff
@@ -4104,6 +4105,71 @@
         "@jridgewell/sourcemap-codec": "^1.4.14"
       }
     },
+    "node_modules/@jsr/std__bytes": {
```
JPVenson: isn't there a useragent parser without all those dependencies?

.npmrc
```diff
@@ -1,2 +1,3 @@
 lockfile-version=3
 save-exact=true
+@jsr:registry=https://npm.jsr.io
```
nielsvanvelzen: We're not adding third-party repositories for our packages

nielsvanvelzen (dismissed):

package-lock.json
timreichen: Yes there is `ua-parser-js`. updated.

.npmrc
timreichen: Fixed.

BotBlake:
Hi @timreichen 
From my 5 minutes of testing the preview, I don't think this is working 😓 
Is it working in the preview for you?

Thanks already and kind regards 🫡

timreichen:
> Hi @timreichen From my 5 minutes of testing the preview, I don't think this is working 😓 Is it working in the preview for you?
> 
> Thanks already and kind regards 🫡

It works on macos. What is your user agent?

joshuaboniface:
So, I'm kinda iffy on this. Not for code quality reasons or anything like that - that part is fine and I like this implementation.

This came up long ago, and we were generally against it for a basic reason: most of the time what people run on their desktops/browser computers, is not what they want or nay *should* run on their servers. Case in point probably 90+% of users are on Windows, and with this functionality it will show them the Windows server downloads first. Which is fine and all, but we generally would rather people run on Linux, and defaulting to the Windows installer doesn't make that obvious.

Now that said, at our point of maturity I think the problem is fairly inconsequential now; people are gonna click the Windows tab if they want Windows and set it up, so we're probably OK with this being set up. But I want to at least raise the concern for posterity. I'd rather a simpler menu system though, which is part of what our [packaging tiers discussion](https://github.com/jellyfin/jellyfin-meta/discussions/92) was about but that initiative kinda died off. But for now this is OK.

timreichen:
> s fairly inconsequential now; people are gonna click the Windows tab if they want Windows and set it up, so we're probably OK with this being set up. But I want to at least raise the concern for posterity. I'd rather a simpler menu system though, which is part of what our [packaging tiers discussion](https://github.com/jellyfin/jellyfin-meta/discussions/92) was about but that initiative kinda died off. But for now this is OK.

Thanks for the review and raising your concerns.
I think pointing people to linux would need to be handled before they even land on the download page.
Though not directly related to this PR I'll have a look at the [discussion](https://github.com/jellyfin/jellyfin-meta/discussions/92).

BotBlake (approved):
Ready for merge?

timreichen:
@nielsvanvelzen I don't know why the `package-lock` file `peer` properties change all over the place when installing `ua-parser-js`.

PR #1502, opened 2025-08-12
clean clients.ts

BotBlake:
This PR contains multiple changes regarding the Clients list on the Downloads page.
1. Ordering official clients first
2. Splitting official and unofficial clients into separate lists. 
3. fixing some website-id's not being called "website"
4. minor fixes, see commits

Removal of Clients has been reverted.

### DIFF
--- a/src/data/clients.ts
+++ b/src/data/clients.ts
@@ -410,6 +410,57 @@ export const thirdPartyClients: Array<Client> = [
     ]
   },
   {
+    id: 'preserve',
+    name: 'Preserve',
+    description: 'A music client inspired by players such as foobar2000 or Clementine.',
+    clientType: ClientType.ThirdParty,
+    deviceTypes: [DeviceType.Desktop],
+    licenseType: LicenseType.OpenSource,
+    platforms: [Platform.Browser, Platform.Desktop],
+    primaryLinks: [
+      {
+        id: 'website',
+        name: 'Website',
+        url: 'https://preserveplayer.com/'
+      },
+      {
+        id: 'gl-downloads',
+        name: 'GitLab Downloads',
+        url: 'https://gitlab.com/tonyfinn/preserve/-/releases'
+      }
+    ],
+    secondaryLinks: [
+      {
+        id: 'gitlab',
+        name: 'GitLab',
+        url: 'https://gitlab.com/tonyfinn/preserve'
+      }
+    ]
+  },
+  {
+    id: 'sonixd',
+    name: 'Sonixd',
+    description: 'A full-featured Subsonic/Jellyfin compatible desktop music player.',
+    clientType: ClientType.ThirdParty,
+    deviceTypes: [DeviceType.Desktop],
+    licenseType: LicenseType.OpenSource,
+    platforms: [Platform.Desktop],
+    primaryLinks: [
+      {
+        id: 'install',
+        name: 'Installation Guide',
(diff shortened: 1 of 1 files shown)


felix920506:
I think this should be broken up into multiple PRs and we shouldn't be doing multiple relatively unrelated things in the same PR.

1 and 2 can be one PR
3 another one
And removing clients one PR for each removal. 

This will make it easier to discuss changes without having one big chaotic thread.

BotBlake:
@felix920506 I have now reverted the client deletions in this PR.
However:

> This will make it easier to discuss changes without having one big chaotic thread.

What changes are there to discuss?
The ordering was already discussed in the documentation room and was agreed uppon.
The website-id's are a minor fix, an `id` should be something to identify the field on, not a random string.
The rename of Jellyfin-iOS was asked for specifically.

I can understand that you want to discuss the client deletion seperately.
However I dont see a point for creating 5 PRs for a few lines of actual changes.

I suggest we discuss the client deletions in the documentation room and I kick up a "remove stale clients" PR, in which all of the ones we decided uppon will get removed at once :+1:

0xceed:
Hey, I wanted to submit a PR including my Discrete music client into the list ([commit](https://github.com/jellyfin/jellyfin.org/commit/1d170098cc73d050f47ede38d8ccadb1859b9339)), but stumped upon this pull request and now wondering, what my steps should be? Wait until this merged? Made changes from this branch? If so, what about sorting?

Or better yet, maybe you instead will include it as well in this PR?

src/data/clients.ts
```diff
@@ -37,7 +37,7 @@ export type Client = {
   recommended?: boolean;
 };
 
-export const Clients: Array<Client> = [
+export const officialClients: Array<Client> = [
```
nielsvanvelzen: No need to export it

src/data/clients.ts
```diff
@@ -147,267 +124,423 @@ export const Clients: Array<Client> = [
...
+      {
+        id: 'github',
+        name: 'GitHub',
+        url: 'https://github.com/jellyfin/jellyfin-xbox'
+      }
+    ]
+  }
+];
+
+export const thirdPartyClients: Array<Client> = [
```
nielsvanvelzen: Also no need to export

src/data/clients.ts
```diff
@@ -128,8 +105,8 @@ export const Clients: Array<Client> = [
     platforms: [Platform.Browser],
     primaryLinks: [
       {
-        id: 'browser',
-        name: 'Open in Browser',
+        id: 'website',
+        name: 'Website',
```
nielsvanvelzen: We differentiate between "hosted" clients and the actual websites advertising the client.

you've now renamed all of those to "website" but this is a hosted instance, I don't think that's a good change. (same for preserve/feishin)

nielsvanvelzen (changes requested):

nielsvanvelzen:
> Hey, I wanted to submit a PR including my Discrete music client into the list ([commit](https://github.com/jellyfin/jellyfin.org/commit/1d170098cc73d050f47ede38d8ccadb1859b9339)), but stumped upon this pull request and now wondering, what my steps should be? Wait until this merged? Made changes from this branch? If so, what about sorting?
> 
> Or better yet, maybe you instead will include it as well in this PR?

Hi @0xceed with this PR merged now you can open yours.

PR #1504, opened 2025-08-13
tasks, storage: note playlist cleanup behavior

altotin:
related issue:  https://github.com/jellyfin/jellyfin/issues/12734#issuecomment-2673088266


BotBlake (commented):
Hi there!
We generally try to keep the amount of "Warnings" in the documentation as low as possible.

Are you sure that both of those notices require the "caution" block?
Otherwise I think its a good addition.

Kind regards

altotin:
hello,    I can understand not wanting to overload with too many warnings

caution felt suitable as data is being removed and the default behavior was a surprise to a number of us in the issue comments.   with storage appearing earlier in the docs, let's keep that as caution;  change tasks to note?

BotBlake:
I agree with you here.
I think having the same note twice is a bit overkill though.
Maybe remove the tasks one to just normal text beginning with the word "Note:"?

We REALY try to keep the "Note"/"Warnings"/... Block-Count as low as possible. so that IF there is a "Note" somewhere it is easy to catch.

Thanks and kind regards!

altotin:
Tasks changed to a plain text "Note: ..."

BotBlake:
Pretty sure I forgot to press the merge button 😅 
Sorry! Thanks for the PR again!

PR #1513, opened 2025-08-14
Updated `storage.md` with information about ZFS

HammyHavoc:
Thought info on ZFS was sorely lacking and is something I've been experimenting with due to there not being much information available on the topic. :- )

### DIFF
--- a/docs/general/administration/storage.md
+++ b/docs/general/administration/storage.md
@@ -1,13 +1,11 @@
 ---
 uid: server-storage
 title: Storage
----
-
-## Storage
+-
 
 Jellyfin is designed to directly read media from the filesystem. A network storage device that is using samba or NFS must be directly mounted to the OS. The Jellyfin database should also be stored locally and not on a network storage device.
 
-### NFS
+## NFS
 
 In case you encounter performance issues where files take a long time to start playing while using NFSv3, you might be running in an issue with .NET locking without NFSv3 having locking enabled.
 
@@ -18,18 +16,36 @@ To solve this, you have the following options:
 - Enable the lock service.
 - Use NFSv4 which has built-in lock support.
 
-## Docker or VMs
+## Docker and VMs
 
-For storage, a moderate size library database can grow anywhere from 10 to 100 GB. The [transcoding](/docs/general/post-install/transcoding) folder needs roughly the same size as the original media if it's being transcoded at the same bitrate. A single 50GB Blu-Ray Remux by itself can take up to approximately 60GB or as little as 15GB, depending on the quality selected. If the transcoding folder is held on the same storage as the database, this must be taken into consideration.
+A database for a moderate-sized library can grow anywhere from 10 to 100 GB.
+
+The [transcoding](/docs/general/post-install/transcoding) folder needs roughly the same size as the original media when transcoded at the same bitrate. A single 50 GB Blu-Ray remux by itself can take up to ~60 GB or as little as ~15 GB, depending on chosen quality. If the transcoding folder is held on the same storage as the database, this must be taken into consideration.
 
 ## Cloud
 
 A popular choice for cloud storage has been the program [rclone](https://rclone.org/downloads/). It is supported on most Operating Systems. To facilitate combining local and cloud filesystems, rclone can be paired with another program such as [mergerfs](https://github.com/trapexit/mergerfs). For cloud storage, it is recommended to disable image extraction as this requires downloading the entire file to perform this task.
 
-### MergerFS
+## MergerFS
 
 MergerFS isn't meant for everything, [see here](https://github.com/trapexit/mergerfs#what-should-mergerfs-not-be-used-for) for more.
 
 - rclone recommended [config](https://forum.rclone.org/t/my-best-rclone-config-mount-for-plex/7441).
 
 To modify and examine your mergerfs mount, here's a quick [guide](https://zackreed.me/mergerfs-neat-tricks).
+
+## Filesystem Considerations
+
+With certain filesystems, optimizations are highly recommended for acceptable performance.
+
+### ZFS
+
+Whilst development is being done on further database providers, in current implementation of Jellyfin Server, the database uses SQLite. ZFS uses a default record size of `128 K`. This is sub-optimal for the SQLite database.
+
+Ideally, you want `4 K` or `8 K` for the dataset that contains your database. This is easily configured when running Jellyfin Server within a Docker container as you are able to easily change bind mounts and can set various datasets for each path as appropriate.
+
+The record size for your media file dataset(s) must not be using `4 K` or `8 K`, otherwise you will likely encounter performance issues as your database scales.
+
+For ZFS datasets containing large media files (i.e., not the dataset containing the Jellyfin Server SQLite database), a record size of `1 M` is likely appropriate for optimal performance.
+
+N.b. Changing the record size on an existing ZFS dataset will not change the existing data within it, meaning performance will not be any different for anything but newly-written changes into the dataset. As such, it is recommended to rewrite your data into the dataset to take advantage of the change in record size, otherwise the configuration change will not perform as expected.


docs/general/administration/storage.md
```diff
@@ -18,18 +16,38 @@ To solve this, you have the following options:
 - Enable the lock service.
 - Use NFSv4 which has built-in lock support.
 
-## Docker or VMs
+## Docker and VMs
+
+A database for a moderate-sized library can grow anywhere from 10 to 100 GB.
 
-For storage, a moderate size library database can grow anywhere from 10 to 100 GB. The [transcoding](/docs/general/post-install/transcoding) folder needs roughly the same size as the original media if it's being transcoded at the same bitrate. A single 50GB Blu-Ray Remux by itself can take up to approximately 60GB or as little as 15GB, depending on the quality selected. If the transcoding folder is held on the same storage as the database, this must be taken into consideration.
+The [transcoding](/docs/general/post-install/transcoding) folder needs roughly the same size as the original media when transcoded at the same bitrate. A single 50 GB Blu-Ray remux by itself can take up to ~60 GB or as little as ~15 GB, depending on chosen quality. If the transcoding folder is held on the same storage as the database, this must be taken into consideration.
```
felix920506: A remux will be the same size as the source. You only have quality settings when you are transcoding.

HammyHavoc: Derp. Thanks! :- ) https://github.com/jellyfin/jellyfin.org/pull/1513/commits/1851521ef736e1a9659f66129b81701ed699ebe8

docs/general/administration/storage.md
```diff
@@ -1,13 +1,11 @@
 ---
 uid: server-storage
 title: Storage
----
-
-## Storage
+-
 
-Jellyfin is designed to directly read media from the filesystem. A network storage device that is using samba or NFS must be directly mounted to the OS. The Jellyfin database should also be stored locally and not on a network storage device.
+Jellyfin is designed to directly read media from a filesystem. A network storage device using Samba or NFS must be directly mounted to the OS. The Jellyfin Server database should also be stored locally and not on a network storage device for acceptable performance.
```
BotBlake: I dont think it is a good idea to name the specific Software. Instead the Protocol should be names (just like with NFS)
```suggestion
Jellyfin is designed to directly read media from a filesystem. A network storage device using SMB or NFS must be directly mounted to the OS. The Jellyfin Server database should also be stored locally and not on a network storage device for acceptable performance.
```

docs/general/administration/storage.md
```diff
@@ -18,18 +16,40 @@ To solve this, you have the following options:
...
 - Use NFSv4 which has built-in lock support.
 
-## Docker or VMs
+## Docker and VMs
+
+A database for a moderate-sized library can grow anywhere from 10 to 100 GB.
+
+The [transcoding](/docs/general/post-install/transcoding) folder needs roughly the same size as the original media when transcoded at the same bitrate. A single 50 GB Blu-Ray remux by itself can take up to ~60 GB or as little as ~15 GB. If the transcoding folder is held on the same storage as the database, this must be taken into consideration to prevent running out of storage and thus corrupting your database.
 
-For storage, a moderate size library database can grow anywhere from 10 to 100 GB. The [transcoding](/docs/general/post-install/transcoding) folder needs roughly the same size as the original media if it's being transcoded at the same bitrate. A single 50GB Blu-Ray Remux by itself can take up to approximately 60GB or as little as 15GB, depending on the quality selected. If the transcoding folder is held on the same storage as the database, this must be taken into consideration.
```
BotBlake, line 27: Why did you remove this?
I still think it is valuable information to take into consideration before migrating data to "the cloud".
Please elaborate.

docs/general/administration/storage.md
```diff
@@ -18,18 +16,40 @@ To solve this, you have the following options:
...
+## Cloud Storage Providers
 
-## Cloud
+[rclone](https://rclone.org/downloads/) is a popular choice for integrating cloud storage with a Jellyfin Server. rclone is supported on most operating systems. To combine local and cloud filesystems, rclone can be paired with another program such as [mergerfs](https://github.com/trapexit/mergerfs).
 
-A popular choice for cloud storage has been the program [rclone](https://rclone.org/downloads/). It is supported on most Operating Systems. To facilitate combining local and cloud filesystems, rclone can be paired with another program such as [mergerfs](https://github.com/trapexit/mergerfs). For cloud storage, it is recommended to disable image extraction as this requires downloading the entire file to perform this task.
+When using cloud storage, it is recommended to disable image extraction as the task requires downloading the entire file.
```
BotBlake: Optional change, I dont realy care about but I think saying "the task" is more confusing then just saying "it" for "this feature".

```suggestion
When using cloud storage, it is recommended to disable image extraction as it requires downloading the entire file.
```

docs/general/administration/storage.md
```diff
@@ -18,18 +16,40 @@ To solve this, you have the following options:
...
+
+Ideally, you should use a record size of `4 K` or `8 K` on the dataset that contains your Jellyfin Server SQLite database. This is easily configured when running Jellyfin Server within a Docker container as you are able to easily change bind mounts and can set various datasets for each path as appropriate.
+
+The record size for your media file dataset(s) must not be using `4 K` or `8 K`, otherwise you will likely encounter performance issues as your database scales.
+
+For ZFS datasets containing large media files (i.e., not the dataset containing the Jellyfin Server SQLite database), a record size of `1 M` is likely appropriate for optimal performance.
+
+N.b.: Changing the record size on an existing ZFS dataset will not change the existing data within it, meaning performance will not be any different for anything but newly-written changes into the dataset. As such, it is recommended to rewrite your data into the dataset to take advantage of the change in record size, otherwise the configuration change will not yield the expected change in performance.
```
BotBlake: As a non native English speaker, I never heard of `N.b.` before (I know its latin but I think you get what I mean 😅 )
I believe I will not be the only one that will read it and does not know what that is supposed to mean.
Therefore I suggest changing it to a simpler, more common word:

```suggestion
Note that changing the record size on an existing ZFS dataset will not change the existing data within it, meaning performance will not be any different for anything but newly-written changes into the dataset. As such, it is recommended to rewrite your data into the dataset to take advantage of the change in record size; otherwise, the configuration change will not yield the expected change in performance.
```

docs/general/administration/storage.md
```diff
@@ -18,18 +16,40 @@ To solve this, you have the following options:
...
+
+As ZFS snapshots can use a lot of storage over time without a sensible `destroy` schedule, there may be a temptation to keep your data on a mechanical drive instead of an SSD. Do NOT use ZFS-formatted mechanical drives to store your Jellyfin Server data (everything except your media files), or you will have terrible performance. SSD is absolutely necessary.
```
BotBlake:
```suggestion
As ZFS snapshots can use a lot of storage over time without a sensible `destroy` schedule, there may be a temptation to keep your data on a mechanical drive instead of an SSD. Do not use ZFS-formatted mechanical drives to store your Jellyfin Server data (everything except your media files), as this will result in poor performance. An SSD is strongly recommended.
```

docs/general/administration/storage.md
```diff
@@ -1,13 +1,11 @@
 ---
 uid: server-storage
 title: Storage
----
-
-## Storage
+-
```
BotBlake: "I" found the problem with the CI 🗿 
Thanks @nielsvanvelzen 🫶 
```suggestion
---
```

docs/general/administration/storage.md
felix920506: Why does the same section have info about both SSD vs HDD and snapshots?

HammyHavoc: Because it is general insight about using ZFS with Jellyfin to prevent people going astray. Suggestions welcome.

docs/general/administration/storage.md
JPVenson: this is not just about performance but generally as sqlite (our main DB engine) is still a database engine, it requires very precise direct access to the underlying storage system to perform a bunch of tasks and also expects to have block level access. So there can also be things going wrong with locking and race conditions but also in general running a sqlite db over a network share, risks data corruption.

docs/general/administration/storage.md
JPVenson: Yes in general we should aim to avoid abbriviations (barring names that _are_ such)

docs/general/administration/storage.md
HammyHavoc: Because it's relative. How big is "moderate"? Why does it differ between 10GB and 100GB? It's a very poor guesstimate without any actual logic or examples cited. Happy to add it back in if someone has stats. Otherwise, it is like saying "people have between 5GB and 80GB of photographs of their grandmother on their phone".

JPVenson: thats a bit of the point. We cannot tell you. We usually tell people (and i think its also somewhere else in the docs) as a rule of thumb:
`(Number of users) x (Biggest File) x 1.2` 
As the size for the transcode folder.

HammyHavoc: Ah, I thought you were remarking about the removal of `For storage, a moderate size library database can grow anywhere from 10 to 100 GB`.

HammyHavoc: I think we both have our wires crossed. I'm somewhat distracted at this present moment; final radiotherapy session for in-law.

Please do feel free to make suggestions and I'll be happy to wave them through for the sake of getting the documentation updated.

docs/general/administration/storage.md
felix920506: At least split the don't use HDDs and snapshots take up a lot of space into 2 sections.

felix920506: Also, we already recommend against using HDDs for Jellyfin elsewhere in the docs.

HammyHavoc: Actionable suggestions welcome.

docs/general/administration/storage.md
```diff
@@ -18,18 +16,40 @@ To solve this, you have the following options:
 - Enable the lock service.
 - Use NFSv4 which has built-in lock support.
 
-## Docker or VMs
+## Docker and VMs
+
+A database for a moderate-sized library can grow anywhere from 10 to 100 GB.
+
+The [transcoding](/docs/general/post-install/transcoding) folder needs roughly the same size as the original media when transcoded at the same bitrate. A single 50 GB Blu-Ray remux by itself can take up to ~60 GB or as little as ~15 GB. If the transcoding folder is held on the same storage as the database, this must be taken into consideration to prevent running out of storage and thus corrupting your database.
```
BotBlake: Woah. Its new info to me that a full transcoding folder would cause db corruptions 😓 
```suggestion
The [transcoding](/docs/general/post-install/transcoding) folder typically requires about the same amount of space as the original media when transcoded at an equivalent bitrate. For example, a single 50 GB Blu-ray remux might consume as much as ~60 GB or as little as ~15 GB after transcoding. If the transcoding folder shares the same storage as the database, this should be accounted for to avoid any problems.
```

BotBlake (approved):
Hi there!
Only one suggestion for change.
Otherwise this looks good to me!

Thanks for the changes and kind regards.

BotBlake (changes requested):
Pressed the wrong button on the previous review 😆

docs/general/administration/storage.md
HammyHavoc: Only a risk if it shares the same storage medium with your database and you're not using a scratch disk or something like a Fusion-io flash array (although that's semi-oldskool these days I must say).

Hope you're well! :- )

BotBlake: I have had issues with that before, because my transcoding cache and my database are on the same ZFS Pool.
All it caused was Jellyfin not starting up properly. 
No data lost though.

I'm doing good yea, thanks for asking.
How r u?
Btw, you can also join our documentation chat on matrix or discord if you want 👍 
https://jellyfin.org/contact/

BotBlake: Hi there @HammyHavoc 
Could you please elaborate?
Thank you and kind regards.

BotBlake:
@HammyHavoc could you please resolve the merge conflicts (+ have a look at the review comments?
I'd love to merge this!

HammyHavoc:
> @HammyHavoc could you please resolve the merge conflicts (+ have a look at the review comments? I'd love to merge this!

Resolved and merged, though as per a comment I made last year on one of my PRs, I will not be contributing further towards the Jellyfin project as it stands after another contributor told me I had used AI simply because I casually use Latin and semi-colons in my everyday speech (autistic, British, and physicists and chemists in the family who used Latin acronyms constantly, even spoken aloud; you can probably guess how pissed off that made me as how I am is not something I can help).

Also not happy with the 'othering' and general attitudes of certain other contributors when it comes to statistical bias with "millions of users" not having a problem with Jellyfin, which apparently thus invalidates the apparent handful that do and must mean with absolute certainty that it's absolutely user error and not something else, i.e., the capabilities of the CPU they're using, and can't possibly be a bug or quirk stemming from EFCore either. With so many variables in envs, it honestly boggles the brain to see people get shouted down.

Furthermore, I strongly dislike the angle that users are somehow 'whining', 'entitled', and not contributing anything when they report a bug that's arisen post-EFCore migration just because the user or even contributors can't figure out how to replicate the bug, nor what's causing it. The abandonment of scientific method in favor of confidence over facts is depressing to say the least.

If these things were easy, they would already be done. If the information to inform a solution was available then it would already be shared. There's also a very strong disconnect between hardcore devs versus casual users who don't know their arse from their elbow in terms of 'best practices' for bug reports, feature requests et al, and those certain contributors cannot fathom that some people can only use the thing and have no information, time, or skills to contribute. Cannot expect those end users to contribute anything meaningful other than a '+1' to the problem in saying that they're having the problem and isn't just the small handful it was otherwise purported to be. Likewise, I think a large amount of the less appealing side to this project stems from GitHub not really having a user-friendly means of saying `Me too` to an Issue without posting a comment. It's geared towards devs for sure, not non-technical end users. Sort of a perfect storm for less-than-great behavior.

Perhaps the shortfall of contributors is indicative of problems like those above, in my case, that's definitely why I'd rather put my time into more constructive projects that don't have a confrontational, obstinate and unconstructive approach to troubleshooting and cultivating community surrounding a project. I'm passionate about the goal, I do believe that potential is there with Jellyfin for a plethora of different use cases (otherwise, why would I contribute my time and energy?), and Jellyfin was certainly heading in the right direction for a long while, but there's problems for sure with how certain others speak and interact with others.

Consider the above a gesture of good will and not having my time and energy be completely wasted, likewise with any finalizing thereof with other outstanding PRs for Jellyfin I may have.

Perhaps something will change in future, but until such a time, I'll remain a user, albeit a very frustrated one (for many reasons, mostly ones related to the "how" rather than the result). It's nothing that you yourself have personally done, this is just a general statement. If anything, you're one of the few people involved who I've come across who hasn't totally turned me off the project as a whole.

BotBlake:
Hi @HammyHavoc 
I’m sorry to hear about the negative experiences and frustration you’ve had while contributing to Jellyfin.
Especially knowing that some of that stemmed from my own past comments.

Thank you for taking the time to explain your perspective and for seeing this PR through in good faith. Your contributions and the effort you’ve put into this work are appreciated.

Please know that you’re always welcome to contribute again in the future or simply to stop by and chat in the Jellyfin spaces if you ever wish to.

Wishing you all the best moving forward, and thank you again for your work on the project.

Issue #1519, opened 2025-08-16
Confusing client downloads filter


timreichen:
The client downloads filter on https://jellyfin.org/downloads can be confusing and its usage can be improved. This is especially important for first time users, that have no overview yet about all these clients. The questions I was asking myself when using that page for the first time was:
- "Why are there so many official clients (coming from plex) and why are only some recommended?"
- "Where is the official client one for macOS?" before realizing the platform was `Desktop`, and nowhere was the `macOS` icon seen like in the server section.

So I propose the following changes:
- `All` and `Recommended` should be inside the filter section, not separate tabs, as they also filter out clients.
- In addition to `All` and `Recommended` I think there should also be `Official` that only shows official clients.
- The pills toggle cumulatively, I think it would make more sense to have them like radio buttons in that only one Device type is selectable. Is there any reason why one would  for example filter by `desktop` and `mobile` at the same time?
- Client type and platform both have `Desktop` as an option. I think instead of the `Desktop` platform, they should add the OSes they support, for example `Jellyfin Media Player`  should have `Linux`,`macOS` and `Windows` as platforms.
The clients that concerns are:
  - Jellyfin Media Player
  - Jellyfin MPV Shim
  - Jellyamp
  - Preserve
  - Sonixd
  - Feishin
  - Tauon Music Box

PR #1520, opened 2025-08-16
Improve client download filter

timreichen:
ref: https://github.com/jellyfin/jellyfin.org/issues/1519

**Note**: The desktop clients have now the actual platforms listed instad of just `Desktop`. I listed the ones I think they support but they should be checked and verified before merging.


BotBlake (changes requested):
I really like the idea of adding additional filters to this page!

That said, I do have some concerns with the current implementation:

1. **Default view change** – The default view has been changed from Recommended to All Clients. I don’t think this should be the case, as it alters the intended first impression of the page.
2. **Hidden Recommended/All toggle** – The toggle between Recommended and All is now almost entirely hidden, which makes it unclear to users. This functionality should remain visible and obvious.
3. **Too many platforms in the filter menu** – On the All Clients view, the sheer number of platforms makes the filter menu overwhelming, especially on smaller screens or mobile devices where it looks quite cluttered. A submenu or some form of grouping would likely work better here.
4. **Filtering by platform** – Since the Recommended view only lists one client per platform, filtering by platform seems a bit redundant in practice.
5. **Filtering by device type** – Similarly, filtering by Device Type doesn’t add much value, since platforms generally already imply the device type. For example, Android suggests mobile, while Android TV clearly indicates a TV device.

I’d love for us to refine this feature together and agree on a meaningful set of filter combinations. The idea is great, but allowing filtering by everything feels unnecessary and potentially confusing in its current form.

timreichen:
> I really like the idea of adding additional filters to this page!
> 
> That said, I do have some concerns with the current implementation:
> 
> 1. **Default view change** – The default view has been changed from Recommended to All Clients. I don’t think this should be the case, as it alters the intended first impression of the page.
> 2. **Hidden Recommended/All toggle** – The toggle between Recommended and All is now almost entirely hidden, which makes it unclear to users. This functionality should remain visible and obvious.
> 3. **Too many platforms in the filter menu** – On the All Clients view, the sheer number of platforms makes the filter menu overwhelming, especially on smaller screens or mobile devices where it looks quite cluttered. A submenu or some form of grouping would likely work better here.
> 4. **Filtering by platform** – Since the Recommended view only lists one client per platform, filtering by platform seems a bit redundant in practice.
> 5. **Filtering by device type** – Similarly, filtering by Device Type doesn’t add much value, since platforms generally already imply the device type. For example, Android suggests mobile, while Android TV clearly indicates a TV device.
> 
> I’d love for us to refine this feature together and agree on a meaningful set of filter combinations. The idea is great, but allowing filtering by everything feels unnecessary and potentially confusing in its current form.

Thanks for your review.

1. I agree. Additionally, instead of `All Clients` I wonder if it would make sense to have the options be `Recommended (default)`, `Official` and `Third Party`. I think `All` isn't really helpful as it filters nothing.
2. Maybe we should get rid of the filter toggle altogether and just display it by default. On lager displays the filter section wouldn't take up too much space and if the filter is well designed, it wouldn't either on smaller displays.
3. True, on smaller displays this looks very messy.

4./5. How about use a selects instead of listing all options as pills? This would clean up the filter section significantly. If we'd do the same with the `recommended/official/third party`, There would only be one row of filters.

BotBlake:
Ignoring the UI for a moment, I think filtering by Platform makes the most sense and should definitely remain. I can easily see people searching for _“a Jellyfin client that works on my Android device”_ rather than _“a Jellyfin client that works on phones.”_

While filtering by Device Type is a nice idea in theory, I personally find it unnecessary. As again, the platform already implies the device type.

At the moment, I envision just two meaningful filters:
1. View Mode: Recommended (default) / Official / All
2. Platform(s): A dropdown with multi-select capability

What do you think?

timreichen:
> Ignoring the UI for a moment, I think filtering by Platform makes the most sense and should definitely remain. I can easily see people searching for _“a Jellyfin client that works on my Android device”_ rather than _“a Jellyfin client that works on phones.”_
> 
> While filtering by Device Type is a nice idea in theory, I personally find it unnecessary. As again, the platform already implies the device type.
> 
> At the moment, I envision just two meaningful filters:
> 
> 1. View Mode: Recommended (default) / Official / All
> 2. Platform(s): A dropdown with multi-select capability
> 
> What do you think?

1. Yes, with an additional `Third Party` option for completeness:
Bildschirmfoto 2025-08-18 um 00 19 17


2. Why multi-select capability? Would anyone search for multiple platforms at the same time? I think we could make it just a select but grouping it by device types:
Bildschirmfoto 2025-08-18 um 00 15 39

WDYT?

BotBlake:
I think your latest idea is great—grouping the filter options by device type makes a lot of sense!

As for multi-select: I believe it’s still valuable. Some users may own several different devices and want to see clients that work on all of them at once. Plus, it would allow someone to filter just for “device type: TV” if they really wanted to do so.

timreichen:
> I think your latest idea is great—grouping the filter options by device type makes a lot of sense!
> 
> 
> 
> As for multi-select: I believe it’s still valuable. Some users may own several different devices and want to see clients that work on all of them at once. Plus, it would allow someone to filter just for “device type: TV” if they really wanted to do so.

The current filter is inclusive though, so it will just show all clients for all selected platforms.

timreichen:
In general, I think the goal should be to find the client one needs as fast as possible for a certain device. The client card also shows what other clients are supported.

BotBlake:
Sorry it took me so long to respond.
Just note that I’m not fixed on the "Multi-Selection" approach.

Overall, I think we’ve already agreed on some good filtering categories.
I’d leave the rest open for you to decide for now.

Don’t worry—I’ll show up again with some critique if I don’t like it later 😆

timreichen:
> Sorry it took me so long to respond. Just note that I’m not fixed on the "Multi-Selection" approach.
> 
> Overall, I think we’ve already agreed on some good filtering categories. I’d leave the rest open for you to decide for now.
> 
> Don’t worry—I’ll show up again with some critique if I don’t like it later 😆

Alright, I pushed an update. Could you try it and give some feedback about the implementation?

BotBlake (approved):
Good work.
I personaly am a fan!
Awaiting core team review.

nielsvanvelzen (changes requested):
Overall this is a downgrade of the current download page. In the future, I'd recommend discussing your proposed changes with the team first.

- Using a full-width div to then show 2 selects in a column is a waste of space, it also introduces an additional click to change a filter
- We do not differentiate between desktop platforms in our official offerings, as such there is no need to split desktop into linux/macos/windows
- The styling of the select boxes is just plainly horrible and text inside of it is cutoff
- The styling of the server and client download pages is no longer consistent
- The urls changed which is breaking all links in the wild to our download pages

timreichen:
> - Using a full-width div to then show 2 selects in a column is a waste of space, it also introduces an additional click to change a filter
>

I guess you are talking about waste of space on wide screens?
That can be fixed by wrapping only on smaller screens👍 On mobile this looks fine to me.
The additional click is to open the select. However, the select combines the two current filters into one by grouping, potentially getting rid of one click. The user also had to click once to get the filters to show up in the first place, so I think this may actually be less clicks.

> - We do not differentiate between desktop platforms in our official offerings, as such there is no need to split desktop into linux/macos/windows

You are technically right, however a new user doesn't know the filter is structured that way.
The goal here is to clarify the difference of device types and software and help users find the client they are looking for fast.
The current setup is confusing because both filters have a `desktop` option.
Seeing the os icon on the client is better than the need to read what is supported imo.

> - The styling of the select boxes is just plainly horrible and text inside of it is cutoff

What platform are you on? There is no cutoff on any of my devices I tested this.

> - The styling of the server and client download pages is no longer consistent

I am working on an update on the server page as well: https://github.com/jellyfin/jellyfin.org/pull/1521 makes the server page single page like the client and I will open another PR that puts the `os` and `stable/unstable` under the `client/server` pills, as they kinda work like server filters.
So both will work the same:
- client/server selection on top
- filters underneath 
- list of entries

> - The urls changed which is breaking all links in the wild to our download pages

Can we add redirects to fix this?

BotBlake:
Perhaps it makes sense to have our design team (@erikbucik) take a look at the design changes.
I remember discussing this topic with him recently.

Note:
GitHub seems to be having issues running Actions at the moment, so I’ll re-trigger them later.
The preview deployment should be available afterwards.

timreichen:
> Perhaps it makes sense to have our design team (@erikbucik) take a look at the design changes.
> 
> I remember discussing this topic with him recently.
> 
> 
> 
> Note:
> 
> GitHub seems to be having issues running Actions at the moment, so I’ll re-trigger them later.
> 
> The preview deployment should be available afterwards.

That would be great!

erikbucik:
I have been summoned. 🤚

First of all, I don't like our current Downloads page - it's too confusing for ~~first time~~ visitors, and it makes it hard to download the appropriate files.

#1519 asks two great questions!
_But you didn't actually solve any of them._
<sup>except for the addition of platform icons - I like that one</sup>
Really, the filtering options have just been moved around a bit - moved into a dropdown instead of an inline container.

> **based on**	https://5b35e0a9.jellyfin-org.pages.dev

image

I wrote a quick proposal in #1878

erikbucik:
I just noticed what it's inspired by 😅
_A single-cohice list **is** cleaner than additive filters._

image

timreichen:
Converting this to a draft until https://github.com/jellyfin/jellyfin.org/issues/1878, https://github.com/jellyfin/jellyfin.org/pull/1885, https://github.com/jellyfin/jellyfin.org/pull/1886 and https://github.com/jellyfin/jellyfin.org/pull/1888 are resolved as this PR makes lots of changes at once of which some are  now addressed by these smaller separate PRs.

PR #1521, opened 2025-08-16
Make server downloads single page

timreichen:
This PR makes the server downloads a single page.
It uses an `os` search param to switch to and store the selected os, like the client downloads page does.

### DIFF
--- a/src/data/downloads.tsx
+++ b/src/data/downloads.tsx
@@ -13,11 +13,11 @@ export enum Feature {
 }
 
 export enum OsType {
-  Docker,
-  Linux,
-  MacOS,
-  Windows,
-  DotNet,
+  Docker = "Docker",
+  Linux = "Linux",
+  MacOS = "MacOS",
+  Windows = "Windows",
+  DotNet = "DotNet",
 }
 
 export type Button = {
--- a/src/pages/downloads/docker.tsx
+++ b/src/pages/downloads/docker.tsx
@@ -1,8 +1,0 @@
-import React from 'react';
-
-import DownloadsPage from './server';
-import { OsType } from '../../data/downloads';
-
-export default function DockerDownloads() {
-  return <DownloadsPage osType={OsType.Docker} />;
-}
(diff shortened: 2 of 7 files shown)


BotBlake (approved):
LGTM - allthough I dont think I have much say on the Downloads page 😆

nielsvanvelzen (changes requested):
We should keep the clean URL's we already have instead of using query parameters.

timreichen:
> We should keep the clean URL's we already have instead of using query parameters.

The point of this PR is to do exactly the opposite, so I don't quite understand what you mean by requesting changes. Could you elaborate?
There isn't any advantage serving a separate page for each download item as all items are served to and filtered on the client side anyway. The only difference is the url itself which is covered by redirect entries.

nielsvanvelzen:
Well if the entire point of the PR is to use query parameters then I don't see any chances of it getting merged.

timreichen:
> Well if the entire point of the PR is to use query parameters then I don't see any chances of it getting merged.

query parameters are used to filter clients, so this seems to improve consistency imo.

nielsvanvelzen:
If we're going for consistency then I'd say we need to change the client pages instead.

timreichen:
> If we're going for consistency then I'd say we need to change the client pages instead.

That wouldn't work because there are multiple parameters for filtering, the pages needed would be exponential.

At the moment they are a weird mix of path segment and query parameters.

- ua analysis preselects the item shown at `https://jellyfin.org/downloads/server`, so this url is not unique
- `os` is currently a path segment but filtering happens on client
- `stable/unstable` is not part of the url and state gets lost
- `help` is not part of the url and state gets lost

Same for clients:
- recommendations `all` is a path segment, `recommended` is not, but filtering happens on client
- filter open/close state is not part of the url and state gets lost
- all other filters are stored in query parameters

All these states could (imo should) be stored as query parameters so the same page can be reproduced again.

timreichen:
@nielsvanvelzen Should I update this PR to include the mentioned states as query parameters, so no states are lost?

nielsvanvelzen:
All current URLs need to keep working. Moving the stable/unstable buttons to query params is fine but thee linux/docker/windows/etc. urls should be kept. Changing the UA detection to a redirect is also fine.

PR #1537, opened 2025-08-31
Update series info to use coherent naming scheme

JPVenson:


felix920506 (approved):

JPVenson:
1. Yes some time ago all our examples had the series name in the filename
2. this is my 3rd PR that realignes the series names into a coherent format
3. I dont really care if its "series" or "shows" as long as its coherently used

Shadowghost (approved):

PR #1538, opened 2025-08-31
Docs: Update web-config.md and fix web config volume

BenRoe:
This PR fix the docker volume example for web config file.

`--volume /path/to/config/web-config.json:/jellyfin/jellyfin-web/config.json` did not work for me, but i use the linuxserver image.
I'm not sure if that was the problem.


JPVenson:
Yes it is. Our docs only handle the official docker container not the liso one

PR #1547, opened 2025-09-10
docs: remove legacy caution warning

toast-gear:
The linked issue in the caution says it was fixed by https://github.com/jellyfin/jellyfin/pull/13906 and so it is no longer true presumably


felix920506:
Duplicate of #1411

PR #1562, opened 2025-09-21
Update traefik.md from v2 to v3

paddy74:
**Changes**

This PR updates [`traefik.md`](docs/general/post-install/networking/reverse-proxy/traefik) to the current major version of Traefik, **v3**. It includes only minimal changes and does not intend to do _anything_ outside this scope.

- Create a copy of `traefik.md` named `traefik2.md` for continued reference by Traefik v2 users.
  - `traefik2.md` is **unlisted** and therefore will not show up in the navbar and (I assume) will not be indexed by search.
- Replace any `2.x` text in `traefik.md` with `3.x`.
- Remove deprecated headers in the `docker-compose`.yml code block.
  - As appropriate, replace with similar functionality.

**Issues**

- This PR is related to #1390, but with a significantly narrower scope.

### DIFF
--- a/docs/general/post-install/networking/8_reverse-proxy/traefik.md
+++ b/docs/general/post-install/networking/8_reverse-proxy/traefik.md
@@ -85,14 +85,10 @@ services:
       - 'traefik.http.routers.jellyfin.middlewares=jellyfin-mw'
       #### The customResponseHeaders option lists the Header names and values to apply to the response.
       - 'traefik.http.middlewares.jellyfin-mw.headers.customResponseHeaders.X-Robots-Tag=noindex,nofollow,nosnippet,noarchive,notranslate,noimageindex'
-      #### The sslRedirect is set to true, then only allow https requests.
-      - 'traefik.http.middlewares.jellyfin-mw.headers.SSLRedirect=true'
-      #### The sslHost option is the host name that is used to redirect http requests to https.
-      #### This is the exact URL that will be redirected to, so you can remove the :9999 port if using default SSL port
-      - 'traefik.http.middlewares.jellyfin-mw.headers.SSLHost=HOST_NAME.DOMAIN_NAME:9999'
-      #### Set sslForceHost to true and set SSLHost to forced requests to use SSLHost even the ones that are already using SSL.
-      #### Note that this uses SSLHost verbatim, so add the port to SSLHost if you are using an alternate port.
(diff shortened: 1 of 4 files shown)


felix920506:
our current example is already compatible with both traefik 2.x and 3.x

paddy74:
> our current example is already compatible with both traefik 2.x and 3.x

Respectfully, @felix920506, that is not immediately clear to readers. The document exclusively references v2.x and makes no references to v3, so a reader will initially assume the example is "out-of-date" and has not been updated to account for v3. Further reading and expertise is required for a reader to conclude that it is compatible with v3.

Additionally, there are minor areas that are not compatible with v3. Specifically, the referenced Docker image and use of deprecated headers.

Therefore, this PR clarifies that the document is compatible with v3 and makes the subsequent minor changes. It would be valid to remove my added duplicate page `traefik2.md`, but I considered it appropriate due to the reasons stated inline.

felix920506:
@paddy74 The more appropriate way to do 2 versions with minor differences would be something like tabs: https://docusaurus.io/docs/markdown-features/tabs
Especially when the rest of the page would be essentially the same. We want to avoid having the same information in multiple different files for maintainability reasons.
If you are not sure about your changes you are welcome at our documentation channel on matrix/discord: https://jellyfin.org/contact

paddy74:
That makes sense. I will take a look - thank you for the pointers!

BotBlake:
@felix920506 @paddy74 
I think an alternative option would be to just remove the `2.x` in the name.

Since the example supports both `3.x` and `2.x` and hopefully nobody using `1.x` (or lower) anymore, simply renaming to `Traefik` should do the trick.

What do you guys think?

felix920506:
The only thing I'm against is uselessly duplicating stuff. Everything else as long as it's an improvement then whatever

paddy74:
Easy enough. I'll submit a new PR without the duplication.

BotBlake:
@paddy74 @felix920506 
Ultimately, I would not put ANY traefik 2.x specifics in the documentation.
Since traefik 2.11 is basically EOL (https://endoflife.date/traefik) I dont see any reason to keep documentation for it.

I would therefore say to simply update the page to latest traefik 3.x standards and fully disregard traefik 2.x.

Also, since we will not support multiple versions of traefik, I see no point in mentioning the version. Imo, we should always go with "mainstream" version.

PR #1565, opened 2025-09-21
Add Authentication Developer Documentation

anultravioletaurora:
Adds a write up on the authentication methods supported by Jellyfin

Looking for any and all feedback while I'm drafting this 👍

### DIFF
--- a/content/developers/docs/api/authentication.md
+++ b/content/developers/docs/api/authentication.md
@@ -1,0 +1,16 @@
+---
+title: Authentication
+sidebar_position: 1
+---
+
+# Authentication
+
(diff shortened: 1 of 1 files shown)


nielsvanvelzen (changes requested):
Without reading in-depth these points came to mind;

- We should change the page ordering so Authorization comes before Authentication. This is because we still need to authorize the client/device (without token) when using the authencation APIs.
- We should at least show the API operation names with a link to api.jellyfin.org when explaining authentication processes
- We should mention that authorization is required for authentication API's, with a link to the appropriate page. Add a remark that the token is (obviously) not needed in this case.

PR #1572, opened 2025-09-27
add uid:gid note

BotBlake:
Updated version of #1207 

Old PR has been closed due to being unmaintained and now entirely replaced by this PR.
All discussion about the change should move to this - actively maintained - PR.

Adding a note to the docker installation guide on how to replace placeholders!

### DIFF
--- a/docs/general/installation/container.md
+++ b/docs/general/installation/container.md
@@ -23,11 +23,7 @@ Additionally, there are several third parties providing unofficial container ima
 
 ## Installation Instructions
 
-:::note
-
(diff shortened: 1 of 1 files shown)


JPVenson:
better to have that as a comment in the compose file

Shadowghost (approved):

BotBlake:
@JPVenson I disagree.
Because we have the "uid:guid" line in the compose file, the docker command AND the podman example configs.
This note applies to all of them, not just the compose file.

Therefore I added it before the Tabs section.

PR #1580, opened 2025-10-01
remove preserve-player

BotBlake:
Removes `preserve-player` from the Client list.

Preserve is a Client with its last release being from 2021.
https://gitlab.com/preserve/preserve/-/releases

Additionally it offers minimal Support of Jellyfins features.
It is a music Only client that does not support audio transcoding at all.

In todays times I do not think this is a Jellyfin Client that we should "recommend" to anyone!
There are a lot of alternatives in the "basic music player" category, and I think this one does not stand out.

We had a discussion about this in the documentation chat, which resulted in me creating my own Jellyfin Client specifically to showcase the lack of features Preserve comes with.

### DIFF
--- a/src/data/clients.ts
+++ b/src/data/clients.ts
@@ -411,34 +411,6 @@ const thirdPartyClients: Array<Client> = [
     ]
   },
   {
-    id: 'preserve',
-    name: 'Preserve',
-    description: 'A music client inspired by players such as foobar2000 or Clementine.',
-    clientType: ClientType.ThirdParty,
-    deviceTypes: [DeviceType.Desktop],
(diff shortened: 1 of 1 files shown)


felix920506:
> Additionally it offers minimal Support of Jellyfins features.
> It is a music Only client that does not support audio transcoding at all.

I am pretty sure this one is not the only music client that doesn't support transcoding at all so I don't think this point would be a good reason.

The other points are reasonable

nielsvanvelzen (approved):

BotBlake:
wait now I broke everything :moyai: 
Stupid web UI :moyai:

nielsvanvelzen (approved):

PR #1581, opened 2025-10-01
Prevent deployment CI from running on forks

BotBlake:
Thanks to @nielsvanvelzen for the help 👍 

Adds a check to some of the build workflows to make sure they only run on the `jellyfin.org` repo and don't run on forks.
This way the CF Deployment will no longer try to run on forks (where cf pages obviously is not set up!)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,9 @@ jobs:
         with:
           cache: npm
           node-version: 20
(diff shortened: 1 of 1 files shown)


felix920506:
more appropriate description would be to stop it from trying to run on forks

nielsvanvelzen (approved):

PR #1583, opened 2025-10-02
remove nginx proxy manager docs

felix920506:


BotBlake (changes requested):
I hate it

BotBlake (approved):
Removing Nginx Proxy Manager from Jellyfin.org entirely :+1:

PR #1587, opened 2025-10-04
Add Joshua's socials to authors.yml

joshuaboniface:


crobibero (approved):

blog/authors.yml
```diff
@@ -33,6 +33,10 @@ joshuaboniface:
   title: Project Leader
   url: https://github.com/joshuaboniface
   image_url: https://avatars.githubusercontent.com/u/4031396?v=4
+  socials:
+    github: joshuaboniface
+    mastodon: https://social.bonifacelabs.ca/@joshuaboniface
+    reddit: https://reddit.com/u/djbon2112
```
nielsvanvelzen, line 39: This is not a valid social media platform in our Docusaurus theme, also doesn't make any sense when we don't use Reddit as Jellyfin.

PR #1593, opened 2025-10-05
Add known issue for Intel OpenCL runtime in newer Linux Distros

SPK44:
Debian Trixie doesn't currently contain the `intel-opencl-icd` package that the guide mentions. A note could save others from a lot of searching.

Additionally it doesn't seem that Jellyfin [works on Debian Trixie anyway](https://forum.jellyfin.org/t-install-server-on-debian-testing-triexie), so providing this known issue could direct users to older stable distros.

### DIFF
--- a/docs/general/post-install/transcoding/hardware-acceleration/known-issues.md
+++ b/docs/general/post-install/transcoding/hardware-acceleration/known-issues.md
@@ -88,7 +88,7 @@ This page lists all known issues and limitations of hardware acceleration with J
 
 14. Resizable-BAR is mandatory for hardware acceleration on BMG / ARC B-series cards, or the [media driver will crash the transcoder](https://github.com/intel/media-driver/issues/1893).
 
-15. Intel Compute-Runtime [currently uses](https://github.com/intel/intel-graphics-compiler/issues/289) LLVM 14 for compilation, making it unavailable in some distibutions like Debian Trixie [as seen here](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1072376). Intel provided [releases](https://github.com/intel/compute-runtime/releases) which only require matching a matching libc runtime may be used instead.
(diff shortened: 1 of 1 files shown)


nyanmisaka:
https://github.com/intel/compute-runtime/releases

Users can actually install them from the Intel built debs. Those packages only have glibc version requirements, as long as the glibc version is the same as that used by Ubuntu 24.04 or newer, they are compatible. Maybe instruct users to also install manually from the Intel repo when the package is not available on Trixie.

docs/general/post-install/transcoding/hardware-acceleration/known-issues.md
```diff
@@ -88,6 +88,8 @@ This page lists all known issues and limitations of hardware acceleration with J
 
 14. Resizable-BAR is mandatory for hardware acceleration on BMG / ARC B-series cards, or the [media driver will crash the transcoder](https://github.com/intel/media-driver/issues/1893).
 
+15. Intel Compute-Runtime [currently uses](https://github.com/intel/intel-graphics-compiler/issues/289) LLVM 14 for compilation, making it unavailable in some distibutions like Debian Trixie [as seen here](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1072376).
```
BotBlake: Please avoid using phrases like “as seen here” in URLs. People who use screen readers need clear, descriptive text to understand the context.
Instead, use something like “as seen on the Debian bug report page” or a similarly descriptive phrase.
The same guideline applies to phrases such as “currently uses.”

SPK44: Got it, I hadn't considered that before. Thanks for pointing it out.

SPK44:
> https://github.com/intel/compute-runtime/releases
> 
> Users can actually install them from the Intel built debs. Those packages only have glibc version requirements, as long as the glibc version is the same as that used by Ubuntu 24.04 or newer, they are compatible. Maybe instruct users to also install manually from the Intel repo when the package is not available on Trixie.

I didn't realize that. I've updated the changeset to include those instructions.

BotBlake (approved):
LGTM, thanks for the quick changes!

PR #1596, opened 2025-10-05
Add generate-versions script and workflow

nielsvanvelzen:
This was requested so we can use the versions of various repositories in the docs.

The workflow is currently untested, data is not used (yet).

Once merged a webhook will be configured to trigger the workflow when a release is tagged. This will then regenerate the versions file and commit it which in turn deploys it to the website.


BotBlake:
Thank you so much <3

src/data/versions.json
```diff
@@ -0,0 +1,250 @@
+{
```
nielsvanvelzen, line 1: Might want to mark this file as generated in the gitattributes

scripts/generate-versions.mjs
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    hasNextPage = pageInfo.hasNextPage;
+    cursor = pageInfo.endCursor;
+  }
+
+  return repos;
+}
+
+try {
```
nielsvanvelzen: This try-catch is somewhat useless as the error will be send to stderr anyway

.github/workflows/generate.yml
```diff
@@ -0,0 +1,34 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
+        with:
+          cache: npm
+          node-version: 20
```
nielsvanvelzen, line 17: sidenote: We should update all our workflows to use Node.JS 22 (LTS)

BotBlake:
@nielsvanvelzen Do you want me to do the changes that you put in your self review? Or was that just a note to yourself? 😅

nielsvanvelzen:
I just reviewed it myself but can't fix it until later, feel free to fix the first two. The Node.Js update should be done separately (after this one is merged).

BotBlake:
> until later

We will see who's "later" comes first 😆 
I would also look at it "later" 🗿

BotBlake (approved):
I donno 🗿 
Looks good tho

nielsvanvelzen:
Meh this needs more work, you cannot use this data yet from docs and I found out you can't trigger workflows via webhook...

BotBlake:
F....
I want this really bad tho 🥲

PR #1604, opened 2025-10-09
docs: update media-segments availability

duhow:
It might be updated for other clients as well, but these are the ones I could find in the Blog announcement.

### DIFF
--- a/docs/general/server/metadata/media-segments.md
+++ b/docs/general/server/metadata/media-segments.md
@@ -42,7 +42,7 @@ Media segments are provided by plugins. Learn more in the [plugin support sectio
 
 ## Clients
 
-Media segments are supported by the web client, Android TV, Roku.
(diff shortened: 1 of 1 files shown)


docs/general/server/metadata/media-segments.md
```diff
@@ -42,7 +42,7 @@ Media segments are provided by plugins. Learn more in the [plugin support sectio
 
 ## Clients
 
-Media segments are currently only supported by the web client.
+Media segments are supported by the web client, Android TV, Roku.
```
felix920506:
```suggestion
Media segments are supported by the web client, Android TV and Roku.
```

nielsvanvelzen:
Would it make sense to simply remove this sentence so we don't need to update it each time a client adds support? We aim to support it in all official clients anyway.

PR #1606, opened 2025-10-10
Update nvidia.md ldconfig command

karoltheguy:
Since the container is running as 1000:1000, it cannot update ldconfig as it's a root owned file. 
I think the command should be docker exec -it --user root jellyfin ldconfig


jpaskaruk:
Can someone please review and get this into the docs? I spent hours figuring this out, and I could have just read it in the docs if someone had gotten this in. Please don't make anyone else go through that.

(Yes, it's true that I now understand a new thing about docker, but that is besides the point...)

BotBlake (approved):
This is correct... Sorry for the HORRIBLE delay.

PR #1628, opened 2025-10-23
add troubleshooting info about db locked on lxc

felix920506:
Discussed on matrix with @JPVenson 

Summary:
The common denominator with database locked issues now seems to be LXC, [even on EXT4](https://github.com/jellyfin/jellyfin/issues/15101#issuecomment-3435004493). Probably safe to assume something is wrong with LXC even though we don't know what is actually wrong. Only thing we can do is to tell people to not use LXC if they face this issue.


JPVenson (approved):

BotBlake:
@JPVenson Is this still the latest and greatest we know about this issue?
Then I will merge :+1:

PR #1630, opened 2025-10-24
use hackmd instead of framapad

felix920506:
framapad sucks for markdown
suggest something designed for markdown instead
HackMD is basically just closed source and cloud hosted version of Hedgedoc

### DIFF
--- a/docs/general/contributing/documentation.md
+++ b/docs/general/contributing/documentation.md
@@ -11,11 +11,11 @@ Documentation is fast-moving and ever-changing. Please think carefully about wha
 
 ## Please Self-Review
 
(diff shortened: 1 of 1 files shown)


anultravioletaurora (approved):

felix920506:
@BotBlake There have already been change proposals to the docs using framapad and it is unreadable for markdown content. That is why I opened the PR.
 https://discord.com/channels/772232779534172171/1400196377740312657/1428194542607405139

> we’ve been trying to stick to open-source tools

There is not a good hosted markdown editor that is open source to my knowledge.

BotBlake:
All I am saying is, that I do not like HackMD as an alternative to Framapad.
I am not trying to block things here. Just pointing out why I dont like it.

Since we have Core approval, I will merge it though.
However I am not a fan!

PR #1638, opened 2025-10-27
Incomplete instructions for stripping nginx's logs of api keys.

### LINKED ISSUE #1637: nginx's "Censor sensitive information in logs" instruction is incomplete
https://jellyfin.org/docs/general/post-install/networking/reverse-proxy/nginx/#censor-sensitive-information-in-logs

I noticed that there are requests using `ApiKey` instead of the mentioned `api_key` URL parameter. This leads to the keys not being stripped in the nginx logs. According to the [source code](https://github.com/jellyfin/jellyfin/blob/efd659412f96ed120866dab350705e1735c085ae/Jellyfin.Server.Implementations/Security/AuthorizationContext.cs#L93-L111) there are even 4 different possible token types that might be used:

- X-Emby-Token
- X-MediaBrowser-Token
- ApiKey
- api_key

I will be preparing a pull request for a proposed better config. If anybody knows if all those tokens are still actively used, please let me know.

wnhre2ur8cxx8:
**Changes**

This adds the more up to date `ApiKey` URL parameter to the instructions on how to set up stripping sensitive informatione from nginx's logs. I found two more possible apikey parameters in the code, but I am unsure if they need to be added as well. Maybe someone with a bit more insight can help here.  

**Copyediting**

**Issues**

closes #1637


BotBlake:
At the moment, there should only be two ways of providing API keys to Jellyfin.

Either through the `Authorization`-Header, or through the `ApiKey` url parameter.
`api_key`, `X-Emby-Token`, `X-MediaBrowser-Token` and `X-Emby-Authorization` are deprecated and should not be used.

I would suggest removing the `api_key` filter as well. What is your oppinion on that?

We are currently working on a developer documentation, so that information like that is more easily available.
At the moment you can check that [here](https://gist.github.com/nielsvanvelzen/ea047d9028f676185832e51ffaf12a6f)

Kind regards!

wnhre2ur8cxx8:
That documentation is very helpful, thank you a lot!

> Note: Not all official Jellyfin clients are currently compatible with legacy authorization disabled.

Looking through my logs there are indeed still some `api_key` calls. It looks like all `/socket` requests use this type of authorization still. This happens on the latest "Jellyfin Media Player" by the way. So no niche client at all.

I would suggest keeping both filters for now.

PR #1650, opened 2025-11-10
Update documentation around .ignore files

### LINKED ISSUE #1649: .ignore files/documentation mismatch
### Description of the bug

After upgrading to 10.11.2 (from 10.10.3, I believe), I noticed that previously hidden media within folders with ".ignore" files were now scanned and available in the library.
That is a jarring change, especially for sensitive subjects. 
I've had my .ignore files annotated with the content of "exclude from JellyFin", so that I know what these files are fore specifically in case I forget, and I believe this (different behavior based on the contents of .ignore files) is being referred to [here](https://github.com/jellyfin/jellyfin/issues/14502#issuecomment-3121586901).
I tried finding official documentation for this, but [the docs](https://jellyfin.org/docs/general/server/media/excluding-directory) simply say
> [excluding a folder] can be achieved by placing a .ignore file in the directory you want to exclude

If the behavior is different based on the contents, it should be mentioned and explained on the above page.

### Reproduction steps

1. on 10.10.3, place an ".ignore" file in a subfolder (containing scannable media) of a monitored folder
2. before it is scanned, enter generic text (such as "exclude from JellyFin") into this file and save
3. scan the library/folder and note that the subfolder is ignored
4. update JellyFin to 10.11.2
5. after a re-scan, note that the previously ignored media is now scanned and added to the library

### What is the current _bug_ behavior?

Media that should be ignored according to the documentation is not ignored

### What is the expected _correct_ behavior?

The documentation reflects ways in which the ignore behavior changes, based on the contents of .ignore files

### Jellyfin Server version

10.11.0+

### Specify commit id

_No response_

### Specify unstable release number

_No response_

### Specify version number

_No response_

### Specify the build version

10.11.2

### Environment

```markdown
- OS: Windows 10
- Linux Kernel: none
- Virtualization: none
- Clients: Browser, any (irrelevant)
- Browser: Chromium, any (irrelevant)
- FFmpeg Version: unsure, I thought the JellyFin version I DL'd came with its own but it seems this is not mentioned in the dashboard. My global version is n6.0-35-g067ed535f4-20231008, any (irrelevant)
...this is getting ridiculous...
- Plugins: AudioDB 10.11.2.0, MusicBrainz 10.11.2.0, OMDb 10.11.2.0, Open Subtitles 23.0.0.0, Playback Reporting 17.0.0.0, Reports 18.0.0.0, Studio Images 10.11.2.0, TMDb 10.11.2.0, Webhook 18.0.0.0
- Jellyfin Data Storage: Local SATA SSD
- Media Storage: Local HDD
```

### Jellyfin logs

```shell
See attached 🙄
```

### FFmpeg logs

```shell

```

### Client / Browser logs

_No response_

### Relevant screenshots or videos

_No response_

### Additional information

Logs attached.
The actual field "Jellyfin Logs" did not allow for attachments.

[log_20251109.log](https://github.com/user-attachments/files/23442881/log_20251109.log)

joshuaboniface:
**Changes**

Updates the `.ignore` file documentation to reflect the changes in jellyfin/jellyfin#13906.

**Issues**

Fixes #1649

### DIFF
--- a/docs/general/server/media/excluding-directory.md
+++ b/docs/general/server/media/excluding-directory.md
@@ -36,7 +36,7 @@ Shows
 
 ## Excluding specific files or subdirectories
 
-When a `.ignore` file is **not empty**, it is treated as a list of file/directory matches using the [same format as a Git `.gitignore` file](https://git-scm.com/docs/gitignore#_pattern_format).
+When a `.ignore` file is **not empty**, it is treated as a list of items to ignore using the [same format as a Git `.gitignore` file](https://git-scm.com/docs/gitignore#_pattern_format).
 
 :::caution
 


docs/general/server/media/excluding-directory.md
```diff
@@ -29,3 +33,27 @@ Shows
...
         └── ...
 ```
+
+## Excluding specific files or subdirectories
+
+When a `.ignore` file is **not empty**, it is treated as a list of items to ignore using the [same format as a Git `.gitignore` file](https://git-scm.com/docs/gitignore#_pattern_format).
+
+:::caution
+
+This is a new feature in Jellyfin 10.11.x and later; older versions did not care about the contents of `.ignore` files. Legacy non-empty `.ignore` files must be manually corrected or they will stop functioning as expected.
```
BotBlake, line 43: Since our documentation consistently reflects only the latest Jellyfin server release, and older versions are unsupported and undocumented, this note may not be necessary.
I’d suggest removing it.

BotBlake (commented):
LGTM. One suggestion: the note about older versions isn’t aligned with our documentation approach, and including it would imply we need to add similar notes for every new feature. It may be better to remove it.

docs/general/server/media/excluding-directory.md
felix920506: This warning should be included as it is a breaking behavior change from old versions in a very significant way.

PR #1654, opened 2025-11-13
Fix monospace ligature bug on macOS

Iksas:
**Changes**

This PR fixes a bug on macOS, where an "fi" ligature is rendered in monospaced text:

Before:
before

After:
after

The bug is fixed by removing all custom CSS that deals with ligatures. 

After testing on multiple operating systems (Windows and macOS), and multiple browsers (Firefox, Safari and Edge), I could not find a single place on the website where ligatures are rendered, except the buggy "fi" ligature on macOS. 

This is why I'd recommend to remove the custom ligature CSS altogether.

**Copyediting**

**Issues**

### DIFF
--- a/src/css/custom.scss
+++ b/src/css/custom.scss
@@ -5,6 +5,15 @@
  * work well for content-centric websites.
  */
 
+html {
+  /* Note: "normal" activates all common ligatures */
+  font-variant-ligatures: normal;
+}
+
+code {
+  font-variant-ligatures: none;
+}
+
 :root {
   /* base font */
   --ifm-font-family-base: 'Noto Sans', sans-serif;


erikbucik (approved):
That's a weird bug. Monospace text should never display ligatures (unless the font itself is designed to support them - and I'm not aware of any that are), because then it's not 'monospace' anymore.

thornbill (changes requested):
In my opinion it would be best to only disable ligatures for elements that use monospace fonts.

Iksas:
I investigated further, and apparently `font-feature-settings` is a low-level feature that is [no longer recommended](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font-feature-settings#:~:text=Whenever%20possible,enable%20small%20caps). Instead, [`font-variant-ligatures`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font-variant-ligatures#:~:text=available%20across%20browsers%20since%20%E2%81%A8January%202020%E2%81%A9) should be used to control ligatures. 

Would it be acceptable to modernize the CSS like this?

    html {
      /* Note: "normal" activates all common ligatures */
      font-variant-ligatures: normal;
    }
    
    code {
      font-variant-ligatures: none;
    }

If not, I'm also open to using the old `font-feature-settings` property:

    html {
      -moz-font-feature-settings: 'liga' on;
      font-feature-settings: 'liga';
    }
    
    code {
      -moz-font-feature-settings: 'liga' off;
      font-feature-settings: 'liga' off;
    }

thornbill:
The proposed modernized css seems like it would be fine 👍

PR #1655, opened 2025-11-16
Note firefox MKV support

felix920506:
https://github.com/jellyfin/jellyfin-web/pull/7335

### DIFF
--- a/docs/general/clients/codec-support.md
+++ b/docs/general/clients/codec-support.md
@@ -217,4 +217,4 @@ If the container is unsupported, this will result in remuxing. The video and aud
 <br />
 <sup>7</sup>OGG on Safari requires at least iOS 18.4 / macOS 15.4
 <br />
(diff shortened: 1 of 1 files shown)


thornbill (approved):

felix920506:
> as of Firefox 145

Unnecessary for our docs, as unless they magically somehow fix all those issues overnight we will have to keep updating this every new Firefox version, so just not including a version number would be better 

> buggy and causes playback issues 

Basically the same meaning as broken and unusable?

> whole mkv file before playback starts, which is extremely undesirable for streaming.

This is not the only issue with mkv files on Firefox. There are other issues for example https://github.com/jellyfin/jellyfin-web/issues/7336

UserX404:
Maybe the firefox ticket should also be mentioned here
https://bugzilla.mozilla.org/show_bug.cgi?id=2000420

BotBlake (approved):
I like that wording a lot better.
Thanks for adressing my review :+1:

PR #1561, opened 2025-09-20
Readd series notes

JPVenson:
Readds the notes for series naming scheme

PR #1656, opened 2025-11-16
Readd series notes

BotBlake:
We have received complaints about removing this note in https://github.com/jellyfin/jellyfin.org/pull/1374

This PR is a rework of https://github.com/jellyfin/jellyfin.org/pull/1561 as the original Author is ignoring our PR reviews.
All further discussions regarding this should be on THIS PR instead of the abandoned one.


BotBlake:
@felix920506

felix920506 (approved):

PR #1662, opened 2025-11-19
Fix markdown lint

felix920506:
I don't know why CI is suddenly failing on markdownlint but here is a fix


.markdownlintignore
```diff
@@ -0,0 +1,6 @@
+blog/2019/*
+blog/2020/*
+blog/2021/*
+blog/2022/*
+blog/2023/*
+blog/2024/*
```
nielsvanvelzen: Opt-out should be per file. We DO NOT use a global ignore file for that.

.markdownlint.yaml
```diff
@@ -17,3 +17,6 @@ MD026:
 
 # MD028/no-blanks-blockquote
 MD028: false
+
+# MD060/table-column-style
+MD060: false
```
nielsvanvelzen, line 22: This is a good rule to have. We should fix the broken tables instead, or change the rule to a different style that fits our markdown. I'm against disabling it entirely.

nielsvanvelzen (changes requested):

.markdownlintignore
felix920506: Changed pushed to address this comment

.markdownlint.yaml
felix920506: I will have to look into this one more deeply. Right now we essentially have markdownlint and prettier fighting each other on how unicode emojis should be formatted when using aligned mode

PR #1663, opened 2025-11-19
Detailed specifics of multiversioned media folder structure

etki:
**Changes**
Just a simple explanation that parent folder is also important for multiversioned media.

**Copyediting**

- [x] <del>I have received [out-of-band peer copyediting](https://jellyfin.org/docs/general/contributing/documentation#peer-copyediting) from someone in [#jellyfin-documentation](https://matrix.to/#/#jellyfin-documentation:matrix.org).</del> it's a one-liner.

**Issues**

### DIFF
--- a/docs/general/server/media/_video-multiversion.md
+++ b/docs/general/server/media/_video-multiversion.md
@@ -2,7 +2,7 @@
 
 ## Multiple Versions
 
-Multiple versions of the same video can be stored together in the library using a file suffix in Movie and Music Video library types. The part before any suffixes **MUST** be exactly the same, including any years and/or metadata provider IDs.
+Jellyfin supports storing multiple versions of the same video within a single movie folder by using filename suffixes. Each file **must** begin exactly with the parent folder name - including any year and/or metadata provider IDs - before adding a version label. This prefix must match character-for-character; otherwise, the files will be treated as separate movies.
 
 ```txt
 Movie (2021) [imdbid-tt12801262]


docs/general/server/media/_video-multiversion.md
```diff
@@ -2,7 +2,7 @@
 
 ## Multiple Versions
 
-Multiple versions of the same video can be stored together in the library using a file suffix in Movie and Music Video library types. The part before any suffixes **MUST** be exactly the same, including any years and/or metadata provider IDs.
+Multiple versions of the same video can be stored together in the library using a separate folder and a file suffix in Movie and Music Video library types. The part before any suffixes **MUST** be exactly the same as the parent folder name, including any years and/or metadata provider IDs.
```
etki: Looking at the code (https://github.com/jellyfin/jellyfin/blob/master/Emby.Naming/Video/VideoListResolver.cs#L191), the parent folder has to be merely a prefix to the files, but i feel it would just be more confusing for users.

BotBlake: Reviewing both the existing documentation and the linked codebase, I think the wording could be updated in general to be more user-friendly. How do you like this suggestion?

```suggestion
Jellyfin supports storing multiple versions of the same video within a single movie folder by using filename suffixes. Each file **must** begin exactly with the parent folder name - including any year and/or metadata provider IDs - before adding a version label. This prefix must match character-for-character; otherwise, the files will be treated as separate movies.
```

BotBlake (changes requested):
Hi there.
Thank you for opening a Pull Request!

The current selection behaviour for Multiple Versions works like so:

```
Movies:
|- My Awesome Movie (2000):
    |- My Awesome Movie (2000)<SUFFIX>.mkv
    |- My Awesome Movie (2000)<SUFFIX>.mkv
```

The Jellyfin documentation already recommends individual folders for each Movie ([here](https://jellyfin.org/docs/general/server/media/movies)).
That is NOT a thing specific to `Multiple Versions`

To me, this does not get much clearer through the changes you have made.
I would suggest rephrasing the sentence entirely (if at all).

Kind regards

etki:
@BotBlake 

> The Jellyfin documentation already recommends individual folders for each Movie

It recommends, but not requires. For multi-version media, this is strictly mandatory, as it can be seen in the linked code. I came up with these changes just because Jellyfin recognized my library without problems, but some step was missing for multi-version media, and doc didn't explicitly tell me what am i doing wrong.

docs/general/server/media/_video-multiversion.md
etki: Something is not set up properly in y account to commit suggested changes from github ui, so i've amended the commit itself and repushed it

BotBlake:
Hi again!

Thanks for addressing the review so quickly!

It looks like you accidentally committed a swap file instead of the actual change 😅
Could you please fix that?

If you prefer, I can correct it for you and apply my suggestion directly.
Just note that this will slow down the review process a bit since another reviewer will need to get involved.

Kind regards

etki:
@BotBlake i'm so sorry, i completely forgot about this one. Things should be fixed now

PR #1669, opened 2025-11-23
Change download status for Arch Linux to Community

JPVenson:
**Changes**
Change download status for Arch Linux to Community as its not maintained by us and makes this circumstance much clearer then the "OsDownload" label

**Copyediting**

- [x] I have received [out-of-band peer copyediting](https://jellyfin.org/docs/general/contributing/documentation#peer-copyediting) from someone in [#jellyfin-documentation](https://matrix.to/#/#jellyfin-documentation:matrix.org).

**Issues**

### DIFF
--- a/src/data/downloads.tsx
+++ b/src/data/downloads.tsx
@@ -299,7 +299,7 @@ sudo apt install jellyfin`}
     id: 'arch',
     name: 'Arch Linux',
     osTypes: [OsType.Linux],
-    status: DownloadStatus.OsPackage,
+    status: DownloadStatus.Community,
     features: [Feature.CustomFFmpeg],
     platforms: [Platform.Arch],
     description: 'Install Jellyfin via Arch-Extra Repository.',


BotBlake (commented):
I am aware we dont maintain the release packages ourselves.
However Jellyfin IS released as an [OS package](https://archlinux.org/packages/extra/x86_64/jellyfin-server/), not only as a [Community build](https://aur.archlinux.org/packages/jellyfin-server-bin).

Also, we are not linking to any specific Repository, but just to the package searches.
I think linking to the actual builds would be better.

"Community" might get users to assume to install Jellyfin from the the less-maintained, less official `Arch-User-Repository`, instead of the official `extra` repository.

JPVenson:
That repackaging is not made by us but a __community__ member that is not affiliated with us of which we have no control over AFAIK. Therfor the community badge should be used to communicate this fact. There are already people in our offical support spaces that assumed its made by us and expect support for it which we cannot provide. As an later alternative maybe both flags can be applied.

BotBlake:
I agree that adding the `Community` Tag does make sense!
However,  to me, "Arch Community Package" means Arch User Repository.

The Idea of adding both tags works for me!

nielsvanvelzen:
The `DownloadStatus.OsPackage` is meant for distributions that have their own package of Jellyfin. As far as I'm aware this is what Arch does so it should not be changed. Community is meant for a package that is NOT managed by the distro itself.

BotBlake:
@nielsvanvelzen 
Note that at the moment we link to BOTH the OS Packages Jellyfin AND a Community maintained one in the AUR.
I do not see a reason to link to the AUR build, as long as there is a OS Package. But if we want to continue doing so, marking it as both "OS Package" AND "Community" would still make sense to me.

I suggest EITHER removing the reference to the Arch User Repository, OR adding the `Community` badge in addition to the already applied `Os Package` one.

BotBlake:
> I suggest EITHER removing the reference to the Arch User Repository, OR adding the Community badge in addition to the already applied Os Package one.

@nielsvanvelzen @JPVenson What do you say?

BotBlake:
Superseded by https://github.com/jellyfin/jellyfin.org/pull/1707

PR #1676, opened 2025-11-30
Update fail2ban config

jammie-jelly:
This PR fixes 2 issues with the fail2ban docs:

1. The logpath in fail2ban jellyfin jail. 

- Current logpath does not match any files (logs or otherwise) within the jellyfin log folder.

2. Add systemd service to reload fail2ban on new log files.

- Fail2ban does not detect newly created files (rotated logs) which is a long standing [issue](https://github.com/fail2ban/fail2ban/issues/2947#issuecomment-784229259).

I have tested both solutions on Ubuntu 24.04 without any issues.

EDIT 1: Solution chosen prior relied on watching log folder with systemd but this strategy proved rather ineffective since path unit triggered service start on every file modification. This became an issue when the systemd service hit the default unit start rate limit as soon as one begins transcoding media for example. 

I've updated solution to use a daily systemd timer that runs 10mins past midnight. Seems the logs are rotated around midnight. This is what I have on my instance, even though more data could be needed around the appropriate time window (10mins seems safe enough) for the logs rotation.

```
for f in /home/ubuntu/docker/jellyfin/jellyfin/log/log_*; do
  echo "$f $(stat -c '%w' "$f")"
done

/home/ubuntu/docker/jellyfin/jellyfin/log/log_20251129.log 2025-11-29 00:01:23.579552038 +0300
/home/ubuntu/docker/jellyfin/jellyfin/log/log_20251130.log 2025-11-30 00:01:48.854113211 +0300
/home/ubuntu/docker/jellyfin/jellyfin/log/log_20251201.log 2025-12-01 00:03:19.396328197 +0300

```

EDIT 2: Seems 10mins cutoff time for the timer was too premature.

```
stat -c '%w' /home/ubuntu/docker/jellyfin/jellyfin/log/log_20251202.log
2025-12-02 00:12:05.977262554 +0300
```


clhodapp:
The rotation-reload logic still feels kind of flimsy because it fundamentally depends on clock-based synchronization rather than logical synchronization. Could you maybe use `inotifywait`?

jammie-jelly:
Jellyfin's default log rotation configuration is based on a daily schedule as said earlier and as seen in the [logging config](
https://github.com/jellyfin/jellyfin/blob/d446fde0091b3b101b2735f21dd24e49f0fd5660/Jellyfin.Server/Resources/Configuration/logging.json#L25). 

There's nothing wrong with following this schedule for reloading the fail2ban config after the new log file has been created. `inotifywait` adds another package to install and is a long running watch process for a single event that happens once a day.

joshuaboniface (approved):
Bit of a hack but it works.

FWIW though, on Debuntu systems, Jellyfin logs also go into syslog and thus should theoretically be available in `/var/log/daemon.log`, which doesn't have this problem. Maybe that's a better solution?

jammie-jelly:
@joshuaboniface That solution doesn't work for containerized environments. 

It should be possible to tell Jellyfin not to rotate logs and use a single log file removing the need for this workaround but you would lose any benefits of log rotation.

Protactin96:
This works for me but there is still the issue with the logging level that I updated in my MR. I also looked a bit further and the logging.json is different depending on if you deploy via docker or if you deploy on Proxmox. The [proxmox helper script](https://raw.githubusercontent.com/community-scripts/ProxmoxVE/main/install/jellyfin-install.sh) that I used replaces the logging level and replaces Information with Error for some reason. I will also raise the issue with the provider of the helper script but it might be good to add this to the requirements.

jammie-jelly:
@Protactin96 don't see how a non-default logging level relates to this PR. The aforementioned proxmox install script is also not official to my knowledge.

Protactin96:
> @Protactin96 don't see how a non-default logging level relates to this PR. The aforementioned proxmox install script is also not official to my knowledge.

The proxmox script is not official. My merge request also changed the log file name and added the log level to the requirements list. The log level in my case was changed from the proxmox helper script but that doesn't change the fact that it is a requirement for fail2ban working. I would have liked to include that requirement point in this PR, then I can close my PR and we only have one PR touching this file.

jammie-jelly:
@Protactin96 Your MR has nothing to do with my submitted PR. Fixing the fail2ban docs (my PR) does not require consideration for unsupported 3rd party scripts.

Protactin96:
> @Protactin96 Your MR has nothing to do with my submitted PR. Fixing the fail2ban docs (my PR) does not require consideration for unsupported 3rd party scripts.

It wasn't about anything 3rd party related but my MR is merged now so it can be ignored here. This MR looks fine to me, systemd timer is probably the easiest way to get around this. I didn't find a solution for this built into fail2ban which is surprising to me as log rotation seems like a common use case.

joshuaboniface (approved):

PR #1682, opened 2025-12-03
Add backup link

theguymadmax:
**Changes**
Add backup-and-restore page

**Issues**
https://github.com/jellyfin/jellyfin-web/pull/7380#issuecomment-3603919648


scripts/data/jellyfin-web-urls.json
```diff
@@ -1,4 +1,5 @@
 [
+  "/docs/general/administration/backup-and-restore.html",
```
thornbill:
```suggestion
  "/docs/general/administration/backup-and-restore",
```

The `.html` extension is a legacy thing. It isn't needed on new links since they don't include it.

PR #1686, opened 2025-12-05
remove fintunes

BotBlake:
Proposing removal of fintunes music client from the Clients list.

This goes in hand with the Client removal guidelines in our [documentation](https://jellyfin.org/docs/general/clients/#removal-from-the-clients-list).

The usage of the `jellyfin` name violates our [Branding guidelines](https://jellyfin.org/docs/general/contributing/branding).
https://github.com/leinelissen/jellyfin-audio-player

Additionally, the log-in process is sub-optimal (opening jellyfin-web, then stealing the token). 

Other then that, the App and website look pretty neat.


BotBlake:
Nothing has changed regarding this topic.
There have also not been any changed to the Client in 4 months now.
Opening the PR as ready for review.

PR #1705, opened 2025-12-28
Updated flathub & github addresses for the official client

guiweber:
**Changes**
Since the name of the official client for desktop was changed to Jellyfin Desktop, the flathub and github addresses have changed as well. This commit updates the website's links to point to the new addresses.

### DIFF
--- a/src/data/clients.ts
+++ b/src/data/clients.ts
@@ -40,7 +40,7 @@ export type Client = {
 const officialClients: Array<Client> = [
   {
     id: 'jellyfin-media-player',
-    name: 'Jellyfin Media Player',
+    name: 'Jellyfin Desktop',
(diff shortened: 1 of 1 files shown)


nielsvanvelzen (changes requested):
We'll wait with updating the name on our site until it's available on all platforms.

BotBlake:
I added an additional PR to rename JMP in two other places in the docs. 👍 
https://github.com/jellyfin/jellyfin.org/pull/1706

BotBlake:
This version of Jellyfin Desktop has been moved out of the Jellyfin organization.
You may find more details in this related issue: https://github.com/andrewrabert/jellium-desktop/issues/609

Therefore we should not update the Documentation to point to this client after all.

Kind regards.

Issue #1595, opened 2025-10-05
Spell checker CI


felix920506:
Spell check review comments can get annoying very quickly for both reviewers and contributors, so it might be better to have some automation deal with that instead

Manually checking spelling is annoying because
- reviewers have to read walls of text
- reviewers have to know English well enough to know the spelling is wrong
- one more thing to watch out for in addition to the content itself

Being manually spell checked is annoying because
- reviewers might miss stuff leading to multiple rounds with new stuff popping up each round
- reviewers might not provide the correct spelling

PR #1724, opened 2026-01-14
Spelling fixes

BotBlake:
Spelling changes (initial run of `cspell`)

Sometimes VsCode also applied some formatting rules.
If that is unwanted then I guess I can fight with it to get that removed :sweat_smile:

PR #1726, opened 2026-01-15
add cspell to CI

BotBlake:
Fixes https://github.com/jellyfin/jellyfin.org/issues/1595

Initial attempt of adding cspell to our Workflows.
The `project-words.txt` has been generated by evaluating the "spelling mistakes" that cspell finds.
All ACTUAL spelling mistakes that it found will be fixed with https://github.com/jellyfin/jellyfin.org/pull/1724

Note that CSpell with the current configuration marks some british-english words as "spelling mistakes"
Specifically `colour` and `behaviour`. Those two have been added to the `project-words` to avoid having to touch Blog posts.
 
I also did find ONE spelling mistake in a blog post: 
https://github.com/jellyfin/jellyfin.org/blob/master/blog/2021/04-12-mirrorbits-cdn.md?plain=1#L329
For now I added `mirorrbits` to the `projects-words.txt`.

I am open of changing this to however wanted, I just followed cspell's "Getting Started" guide for this :laughing:

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -1,6 +1,7 @@
 {
   "$schema": "https://raw.githubusercontent.com/streetsidesoftware/cspell/main/cspell.schema.json",
   "version": "0.2",
+  "language": "en, en-US, en-GB",
   "files": ["**/*.md", "**/*.mdx", "**/*.tsx"],
   "dictionaryDefinitions": [
     {
--- a/project-words.txt
+++ b/project-words.txt
@@ -34,8 +34,6 @@ bantime
 barebones
 Battlemage
 BDMV
-behaviour
-behaviours
 behindthescenes
 blitter
 Blitter
@@ -67,8 +65,6 @@ clinfo
 codenames
 codesign
 collectionnumber
-colour
-colours
 comskip
 configdir
 confs


.github/workflows/spellcheck.yml
```diff
@@ -0,0 +1,16 @@
+name: 'Check spelling'
```
nielsvanvelzen: I'd recommend running cspell on this exact file name

nielsvanvelzen:
There is an eslint plugin for cspell, perhaps it would be easier to use that instead of adding yet another tool to run separately.

cspell.json
```diff
@@ -0,0 +1,14 @@
+{
+  "$schema": "https://raw.githubusercontent.com/streetsidesoftware/cspell/main/cspell.schema.json",
+  "version": "0.2",
+  "files": ["**/*.md"],
```
nielsvanvelzen: We have a lot of documentation/blogs using mdx files to allow usage of React components, which you're now skipping. I'd also argue we should check the regular pages and components containing texts which are mostly in tsx files

nielsvanvelzen (changes requested):

cspell.json
BotBlake: Done :+1:

cspell.json
```diff
@@ -0,0 +1,25 @@
...
+  "files": ["**/*.md", "**/*.mdx", "**/*.tsx"],
+  "dictionaryDefinitions": [
+    {
+      "name": "project-words",
+      "path": "./project-words.txt",
+      "addWords": true
+    }
+  ],
+  "dictionaries": ["project-words"],
+  "ignorePaths": [
```
thornbill, line 14: Looks like instead of adding individual words to the dictionary for Canadian (British) spellings, we should probably just configure the languages appropriately?

https://streetsidesoftware.com/vscode-spell-checker/docs/configuration/languages-and-dictionaries#cspelllanguage

```
"language": "en,en-US,en-GB"
```

BotBlake: I avoided that on purpose, as we currently only use TWO words from another "language". I think it makes more sense to keep everything in American english.

thornbill: Except that the project leadership is Canadian and will likely continue to use GB spelling variants.

BotBlake: I personally think we should set a standard and enforce it.

I will add the GB language to the config tomorrow though if that is what you want.

nielsvanvelzen (approved):

PR #1728, opened 2026-01-16
add old blogposts to ignores

BotBlake:
Some of the old blog-posts have linting and spelling errors appear.
To prevent `eslint` from running on those files (now and in the future), this aims to disable it.

Since `eslint` SHOULD run on new blog posts, this is quite the perfect timing to add this.

Just an Idea to solve issues like this:
https://github.com/jellyfin/jellyfin.org/actions/runs/21069274923/job/60594286766

Note that actual linting errors (with the current config) only appear in the `2020` blog-posts folder.
If the linting rules will ever change they MIGHT touch the old blog posts though.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -43,6 +43,16 @@ export default ts.config([
       }
     },
 
+    ignores: [
+      'blog/2019/**',
+      'blog/2020/**',
+      'blog/2021/**',
+      'blog/2022/**',
+      'blog/2023/**',
+      'blog/2024/**',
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
I would prefer adding ignores on a file-by-file basis once they start having issues. You can disable eslint rules using a comment on the top of the files.

BotBlake:
All current blog-posts are passing the current markdownlint/ eslint CI.
Therefore they dont have to be added to the ignores.

PR #1738, opened 2026-01-22
fix cspell failure

BotBlake:
Fix cspell issues on `master` branch.
Those appeared now, as the initial `addition of cspell`-PR only included fixes for normal markdown files, but `cspell` is running on `.md`, `.mdx` and `.tsx` files.

### DIFF
--- a/project-words.txt
+++ b/project-words.txt
@@ -245,11 +245,7 @@ mcarlton
 mediaserver
 mergerfs
 metaarchive
-metaarchives
-Metaarchives
 metaimage
-metaimages
-Metaimages
 metapackage
 metapackages
 Metapackages


project-words.txt
```diff
@@ -236,18 +244,18 @@ Mbps
...
 metaimage
 metaimages
+Metaimages
 metapackage
 metapackages
+Metapackages
 mintls
-mirrorbits
 mirorrbits
+mirrorbits
```
nielsvanvelzen, line 253: why add this one twice?

project-words.txt
```diff
@@ -236,18 +244,18 @@ Mbps
...
 mergerfs
-Metaarchives
-Metaimages
-Metapackages
 metaarchive
 metaarchives
+Metaarchives
 metaimage
 metaimages
+Metaimages
```
nielsvanvelzen: why add this one twice?

project-words.txt
```diff
@@ -9,8 +9,9 @@ alac
 albumartistsort
 allports
 alpn
-alpn
 analyzeduration
+Anidb
+Anilist
```
nielsvanvelzen: correct capitalization is AniList

project-words.txt
```diff
@@ -9,8 +9,9 @@ alac
 albumartistsort
 allports
 alpn
-alpn
 analyzeduration
+Anidb
```
nielsvanvelzen: correct capitalization is AniDB

project-words.txt
BotBlake: Its not the same word. 
One is the TYPO `mirorrbits` and the other is `mirrorbits` itself.

But since the typo was in an old blogpost anyways, this can be removed from the exception list (blogposts are now part of the ignore files)

project-words.txt
BotBlake: It WAS already added, I just pressed "sort alphabetically" because I apparantly f-ed that up last time :laughing: 

Initially it was added twice to allow capitalization.
But since this PR enabled `CompoundWords` this is no longer needed. Will remove

PR #1746, opened 2026-01-29
Update Chrome MKV support status

FooIbar:
MKV is supported on Chrome since 145:
https://chromiumdash.appspot.com/commit/d0813004e27eb2f679dd4a9e26a13baeac4be9f2


BotBlake:
@gnattu I remember we disabled ?Firefox? mkv support specifically because it did not properly work for our usecase?
Is that different for Chrome? Or can we list this as "supported" officially?
Thanks :+1:

PR #1777, opened 2026-02-26
Workflow seperation

BotBlake:
https://github.com/jellyfin/jellyfin.org/pull/1709 Tried to seperate the build workflow.
Since the main workflow just called it, it had no security context change and had access to all secrets as well.

This PR fully seperates `build` and `deploy` workflows. Deployment will only be triggered after `build` is finished.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,4 @@
-name: Build
+name: build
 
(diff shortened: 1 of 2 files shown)


andrewrabert (approved):
Good but could use some tweaks. Opening a separate PR for those

PR #1778, opened 2026-02-26
Move Cloudflare preview deploys to workflow_run for fork PR support

andrewrabert:
The pull_request_target → pull_request change in 6fad5b23 broke Cloudflare preview deploys for fork PRs because secrets are unavailable. Split the publish job into a separate workflow triggered by workflow_run, which runs in the default branch context with secret access after the untrusted build completes.

Uses gh pr list --head to find the PR number by branch, with the owner:branch format for forks. This works reliably for both fork and same-repo PRs (unlike workflow_run.pull_requests[] which is empty for forks).


thornbill:
This could likely be copied straight from web

PR #1780, opened 2026-02-27
fix: add actions:read permission for cross-run artifact downloads

andrewrabert:
The deploy and publish jobs use download-artifact with run-id to fetch artifacts from the triggering build workflow. This requires actions:read permission, which was missing from both jobs.


andrewrabert:
Hot fix for https://github.com/jellyfin/jellyfin.org/pull/1777

PR #1783, opened 2026-02-28
Add Dee Jellie client information

nflore:
**Changes**
This PR adds Dee Jellie, a third-party proprietary Jellyfin client, to the Clients documentation page.


BotBlake (changes requested):
We are in the process of refining our strategy regarding the client list. Until this is ready we will not be adding any new clients.

PR #1784, opened 2026-03-01
Adding Shark Marmalade to clients.ts

bendardenne:
**Changes**

Adding Shark Marmalade to the clients page. It's been available on the play store for several months, is stable, and has around 150 users by  now. Someone complained to me that it was not easy to find, so perhaps this helps.


BotBlake (changes requested):
We are in the process of refining our strategy regarding the client list. Until this is ready we will not be adding any new clients.

PR #1789, opened 2026-03-04
Remove note about installing `jellyfin-server` and `jellyfin-web`

Valloric:
The note being removed says that instead of installing `jellyfin`, the user can install `jellyfin-web` and `jellyfin-server` packages instead. 

**This is wrong and leads to a broken install** because only the `jellyfin` metapackage depends on `jellyfin-ffmpeg7` package without which the server refuses to start.

The part of the note that says:

> The jellyfin-server package will automatically select the right jellyfin-ffmpeg package for you as well.

is also incorrect; the `jellyfin-server` package did _not_ pull in `jellyfin-ffmpeg7`; it's only a "Recommends" dependency, not a "Depends" dependency.

In short, just tell the user to install `jellyfin` and don't over-complicate matters. :)

**Changes**

**Copyediting**

**Issues**

### DIFF
--- a/docs/general/installation/advanced/manual.md
+++ b/docs/general/installation/advanced/manual.md
@@ -315,18 +315,6 @@ If you would prefer to install everything manually, the full steps are as follow
    sudo apt install jellyfin
    ```
 
-   :::note
-
-   If you want to be explicit, instead of the metapackage, you can install the sub-packages individually:
-
-   ```sh
-   sudo apt install jellyfin-server jellyfin-web
-   ```
-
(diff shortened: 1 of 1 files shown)


felix920506 (commented):
This is the section for advanced installation methods so we should still let users know of this option. If the Jellyfin-ffmpeg package needs to be installed manually instead of being installed automatically, the documentation should just be changed to reflect that reality. 

Jellyfin also does not have a hard dependency on jellyfin-ffmpeg. It just needs some form of ffmpeg for the less advanced features.

felix920506:
Is this still being worked on?

Valloric:
> Is this still being worked on?

No, sorry.

PR #1828, opened 2026-04-24
Add Developer Docs for WebSocket Message Types, Device Profiles, Posted Playback Info, and Audio Streams

anultravioletaurora:
**Changes**

This PR adds additional developer documentation around WebSocket message types that support subscriptions, as well as Device Profiles + Media Info + Audio. 

Device Profiles, Media Info, and Audio developer docs include generic example usages as well as usage with the Typescript SDK.

_`gemma4:e4b` was used for wordsmithing and organization of my thoughts, formatting of the markdown files, and creating example code snippets_ 

**Copyediting**

**Issues**


content/developers/docs/api/audio.md
```diff
@@ -0,0 +1,50 @@
...
+function buildAudioStreamUrl(itemId: string, playSessionId: string, apiBase: string) {
+	const params = new URLSearchParams({
+		playSessionId,
+		static: 'true',
+		startTimeTicks: '0',
+	});
+	return `${apiBase}/Audio/${itemId}/stream?${params}`;
+}
+
+// Usage:
```
BotBlake: No need to add comments in the codeblock when you can put them in markdown below the codeblock

content/developers/docs/api/audio.md
```diff
@@ -0,0 +1,50 @@
+---
+title: Audio
+sidebar_position: 9
+---
+
+# Audio
+
+When playing audio, the client must decide whether to use a transcoding stream or a direct play stream. This is determined by the [playback info response](./media-info.md) from the server.
```
BotBlake: I dont KNOW this but: Does audio also have a "Direct stream"?

On video we have `Direct Play`, `Direct Stream` and `Transcoding`.
`Direct Play Stream` sounds like a mix of both and may cause confusion(?)

content/developers/docs/api/audio.md
```diff
@@ -0,0 +1,50 @@
...
+
+### 1. Use the Transcoding URL (if provided)
+
+If the playback info response includes a `transcodingUrl` (or `TranscodingUrl` in the media source), use this URL to fetch the audio stream. This means the server will transcode the media to a format compatible with the device profile.
```
BotBlake: Link to the device profiles page here.

content/developers/docs/api/audio.md
```diff
@@ -0,0 +1,50 @@
...
+
+If no transcoding URL is provided, the client can perform direct play using the [Audio API stream operation](https://api.jellyfin.org/#tag/Audio/operation/GetAudioStream). The direct play URL is constructed as follows:
+
+- **Endpoint:** `/Audio/{itemId}/stream`
+- **Query Parameters:**
+	- `playSessionId`: The sessionId returned by the posted playback info operation (required for authentication).
+	- `static=true`: Ensures that no encoding is performed (static file delivery).
+	- `startTimeTicks=0`: (optional) Start at the beginning
+
+### Example (TypeScript SDK-style)
```
BotBlake: This is an example for Direct Play URL construction.
Therefore the header indent is incorrect: 

```suggestion
#### Example (TypeScript SDK-style)
```

But I personally dislike having the "TS SDK-style" be part of the heading.

content/developers/docs/api/audio.md
```diff
@@ -0,0 +1,50 @@
...
+// Usage:
+// If playbackInfo.transcodingUrl exists, use that URL directly.
+// Otherwise, use buildAudioStreamUrl(itemId, playbackInfo.sessionId, api.basePath)
+```
+
+### Notes
+
+- Always use the `sessionId` (or `playSessionId`) returned by the posted playback info operation for authentication.
+- The `static=true` parameter ensures the server does not transcode or re-encode the file for direct play.
+- If transcoding is required, always prefer the transcoding URL provided by the server.
```
BotBlake: prefer the Transcoding URL over what?
The above guide says "if the device profile needs transcoding, the server will provide a Transcoding URL".
So given the device profile is correct, then OBVIOUSLY the transcoding url needs to be used 🗿

content/developers/docs/api/audio.md
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// Usage:
+// If playbackInfo.transcodingUrl exists, use that URL directly.
+// Otherwise, use buildAudioStreamUrl(itemId, playbackInfo.sessionId, api.basePath)
+```
+
+### Notes
+
+- Always use the `sessionId` (or `playSessionId`) returned by the posted playback info operation for authentication.
```
BotBlake, line 28: Why? What would be the alternative here?
Also this is a note regarding the transcoding URL again. Since you have a hole section about that, why not make it a sentence in there?

content/developers/docs/api/audio.md
```diff
@@ -0,0 +1,50 @@
...
+- The `static=true` parameter ensures the server does not transcode or re-encode the file for direct play.
```
BotBlake: I just dont understand this.
If the server says "you dont have to transcode" and you build your stream url...
You have to tell the server not to transcode?

+ re-encode sounds like `Direct Stream` to me 👀 
So this also CAN happen for audio?

content/developers/docs/api/device-profiles.md
```diff
@@ -0,0 +1,74 @@
...
+}
+```
+
+## How Device Profiles Are Used
+
+1. **Client Sends Profile**: When a client connects, it provides its device profile to the server (or the server uses a default profile for known devices).
+2. **Server Evaluates Media**: When a user requests playback, the server checks the device profile to see if the media can be sent directly or needs to be transcoded.
+3. **Optimized Delivery**: The server streams the media in the best format supported by the device, using direct play whenever possible.
+
+## Fields Explained
```
BotBlake: You literally repeated the "What is a Device profile" section...

content/developers/docs/api/device-profiles.md
```diff
@@ -0,0 +1,74 @@
...
+
+- `MaxStreamingBitrate`: The maximum bitrate the device can handle for streaming.
+- `DirectPlayProfiles`: List of containers/codecs that can be played without transcoding.
+- `TranscodingProfiles`: List of formats/codecs the server can use when transcoding for this device.
+- `MusicStreamingTranscodingBitrate` (optional): Maximum bitrate for music transcoding.
+- `ContainerProfiles`, `CodecProfiles`, etc.: Advanced options for fine-tuning compatibility.
+
+## Example (TypeScript SDK)
```
BotBlake: LITERALLY a different heading for the example between both guides.
Either you call it "TS SDK" OR you call it "TS SDK-style".
OR we just call it "Example", since no other examples are provided anyways...

content/developers/docs/api/device-profiles.md
```diff
@@ -0,0 +1,74 @@
...
+		{ Container: 'mp3', Type: DlnaProfileType.Audio },
+		{ Container: 'mp4', Type: DlnaProfileType.Video },
+	],
+	TranscodingProfiles: [
+		{ Container: 'ts', AudioCodec: 'aac', VideoCodec: 'h264', Type: DlnaProfileType.Video },
+	],
+};
+```
+
+## See Also
```
BotBlake, line 44: Imo unneeded to have an entire section for ONE url.
But thats just a me thing.

content/developers/docs/api/media-info.md
```diff
@@ -0,0 +1,51 @@
+---
+title: Media Info
+sidebar_position: 8
+---
+## Fetching Playback Info with Device Profiles
+
+Once a [device profile](./device-profiles.md) is created, it can be used to request playback information for a media item from the Jellyfin server. This is done using the [Post Playback Info](https://api.jellyfin.org/#tag/MediaInfo/operation/GetPostedPlaybackInfo) operation.
+
+## How It Works
```
BotBlake, line 9: Repetition of the "How device profiles are used" section on the device profiles guide

content/developers/docs/api/media-info.md
```diff
@@ -0,0 +1,51 @@
...
+
+1. **Send Device Profile**: The client sends its device profile along with the media item ID to the server.
+2. **Server Decision**: The server evaluates the device profile and the media item to determine if direct play is possible or if transcoding is required.
+3. **Playback Info Response**: The server responds with playback details, including URLs for direct play or transcoding, and a session ID.
+
+## Key Fields in the Response
```
BotBlake, line 15: repetition of the Audio guide.
You already say that it returns a transcoding URL and a sessionId.
This is now an entire guide that says that AGAIN.

content/developers/docs/api/media-info.md
```diff
@@ -0,0 +1,51 @@
...
+
+// Usage:
+// const playbackInfo = await fetchMediaInfo(profile, 'itemId');
+// const url = playbackInfo.transcodingUrl || playbackInfo.mediaSources[0]?.path;
+// const sessionId = playbackInfo.sessionId;
+```
+
+## Notes
+
+- Always provide the correct sessionId when fetching the media stream.
```
BotBlake, line 46: DUH.

content/developers/docs/api/media-info.md
```diff
@@ -0,0 +1,51 @@
...
+- The server will return the optimal playback method based on the device profile and media item.
```
BotBlake, line 47: This is only the 10th time mentioning this

BotBlake (changes requested):
Keep in mind this is not supposed to replace the TS SDK Documentation but rather explain the whole frame. Right now it seems VERY Typescript SDK heavy.

Additionally note that not all API actions require their own Page in the developer documentation. You may also create ONE page named playback and mention the needed API actions there.

content/developers/docs/api/audio.md
anultravioletaurora: There is no alternative AFAIK

content/developers/docs/api/audio.md
anultravioletaurora: Basically if you get a transcoding URL, you should use that instead of targeting the AudioAPI with a custom call

content/developers/docs/api/audio.md
BotBlake: Why mention it specifically then?
If there IS no other option?

content/developers/docs/api/audio.md
BotBlake: Yes but you already said that before.
If a transcoding URL is provided, use that.

You are now saying that URL should be preferred. 🗿

Preferred over what?
Preferred over building the DirectPlay URL?
Yea ofc. Because it requires transcoding 🗿

content/developers/docs/api/audio.md
anultravioletaurora: I honestly am not sure, Mr. van Velzen might know tho

content/developers/docs/api/audio.md
anultravioletaurora: I mainly wanted to call it out to settle ambiguity regarding what the playSessionId is used for.

When I had first started Jellify, I had no idea the ramifications behind no supplying the correct playSessionId

content/developers/docs/api/media-info.md
anultravioletaurora: What feels repetitive here?

content/developers/docs/api/device-profiles.md
anultravioletaurora: Removed

content/developers/docs/api/device-profiles.md
anultravioletaurora: Removed the duplicated section

content/developers/docs/api/audio.md
anultravioletaurora: I believe it's only Direct Play

content/developers/docs/api/audio.md
anultravioletaurora: Yeah exactly.

Updated to clarify this.

content/developers/docs/api/audio.md
anultravioletaurora: Removed the TS SDK example

content/developers/docs/api/audio.md
anultravioletaurora: Done.

content/developers/docs/api/media-info.md
BotBlake: I’ll just explain this using one example, but the other points I mentioned are very similar.

In this section you wrote:
> **Send Device Profile**: The client sends its device profile along with the media item ID to the server.

And in another section:
> **Client Sends Profile**: When a client connects, it provides its device profile to the server (or the server uses a default profile for known devices).

This doesn’t just *feel* repetitive. Both statements communicate essentially the same concept with slightly different wording. I think we should try to avoid that kind of duplication throughout the docs.

content/developers/docs/api/media-info.md
BotBlake: We should avoid spreading general-purpose notes throughout the developer documentation.

I think it’s reasonable to assume that developers already understand the basics of interacting with the API. If we feel that assumption is not valid, then it would make more sense to create a dedicated “How to interact with the API” page that centrally explains concepts like validating inputs and providing correct values.

Otherwise, we could end up adding the same reminders to every page and every endpoint description, which would quickly become repetitive and add unnecessary noise to the docs.

In this specific case, I don’t think we need an additional reminder telling developers to provide correct values to the SDK/API.

content/developers/docs/api/audio.md
BotBlake: See my other comment regarding general-purpose Notes:

> We should avoid spreading general-purpose notes throughout the developer documentation.
>
> I think it’s reasonable to assume that developers already understand the basics of interacting with the API. If we feel that assumption is not valid, then it would make more sense to create a dedicated “How to interact with the API” page that centrally explains concepts like validating inputs and providing correct values.
> 
> Otherwise, we could end up adding the same reminders to every page and every endpoint description, which would quickly become repetitive and add unnecessary noise to the docs.
> 
> In this specific case, I don’t think we need an additional reminder telling developers to provide correct values to the SDK/API.

BotBlake (changes requested):

PR #1839, opened 2026-05-05
illustrations' higher contrast

erikbucik:
**Changes**
Illustrations' contrast improvements as per @BotBlake's request.


BotBlake (changes requested):
Unrelated changes.
- added xml header
- added doctype header
- somehow made everything smaller
- huge diff, should only be the fill color

This can not be merged in its current state.

PR #1841, opened 2026-05-06
Add rss link to blog page

### LINKED ISSUE #1664: Blog: Add a visible RSS icon that points to the feed URL
There seems to be a working RSS feed at `https://jellyfin.org/posts/rss.xml` but it's not obvious to the visitor. Please consider adding in the blog or main page a visible RSS icon that points to the feed URL, it helps users stay up to date with what's happening at Jellyfin.

BotBlake:
**Changes**
Mention the RSS Feed on the Blog website.
Since this page is entirely generated by docusaurus, it required swizzling some of docusaurus components.
I opted for a Typescript Eject of the Docusaurus component `BlogSidebar/Content`.

Since previews are still broken, here is a Screenshot:
grafik

I do know this is a lot of effort for such a simple change, so see this as a "suggestion".

**Issues**

https://github.com/jellyfin/jellyfin.org/issues/1664

**Notes**
I have absolutely no Idea where to put the `swizzle` output files. I just left them in the location `npm swizzle` created them at. I know I'm an Idiot :heart_hands:


BotBlake:
Merged master because it includes a docusaurus update.

BotBlake:
Merged master to test preview CI

nielsvanvelzen:
The index.xml is the URL for our old feed and shouldn't be used, although still kept up-to-date. The proper atom/rss feed is already in the meta tags.

BotBlake:
I wasn't sure which feed to use since we specifically have a [script](https://github.com/jellyfin/jellyfin.org/blob/master/scripts/copy-rss-feed.js) to copy the feed from `https://jellyfin.org/posts/rss.xml` to `https://jellyfin.org/index.xml`.

I will update to use the `rss.xml`.

src/theme/BlogSidebar/Content/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+            <BlogSidebarYearGroup key={year} year={year} yearGroupHeadingClassName={yearGroupHeadingClassName}>
+              <ListComponent items={yearItems} />
+            </BlogSidebarYearGroup>
+          ))}
+        </>
+      ) : (
+        <ListComponent items={items} />
+      )}
+      <div
+        style={{
```
nielsvanvelzen: No inline styling

src/theme/BlogSidebar/Content/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+          marginTop: '1rem',
+          paddingTop: '1rem',
+          borderTop: '1px solid var(--ifm-toc-border-color)'
+        }}
+      >
+        <a
+          href='https://jellyfin.org/posts/rss.xml'
```
nielsvanvelzen: Use relative path and scope it to the current blog plugin instance, otherwise this won't work on the developer blog

src/theme/BlogSidebar/Content/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+          target='_blank'
+          rel='noopener noreferrer'
+          style={{
```
nielsvanvelzen: No inline styling

nielsvanvelzen (changes requested):
Aside from the code review, for the design the icon shouldn't re-use the link color and both should be smaller (right now the link is even bigger then the links for actual posts)

BotBlake:
Agreed with the code changes.
Also agreeing on the color changes.

However the size should stay.
In my opinion its a heading.
image

nielsvanvelzen:
It's not a headings but a secondary link. It should visually be obvious it is not a primary link.

src/theme/BlogSidebar/Content/index.tsx
BotBlake: Note that this already came from the Docusaurus component swizzle.

src/theme/BlogSidebar/Content/index.tsx
```diff
@@ -24,8 +25,19 @@ function BlogSidebarYearGroup({
...
 
+function getBlogBasePath(items: Props['items']): string {
+  const firstItem = items[0];
+  if (!firstItem || !('permalink' in firstItem)) {
+    return '/posts';
+  }
+  const permalink = firstItem.permalink;
+  return permalink.substring(0, permalink.lastIndexOf('/'));
+}
+
```
nielsvanvelzen, line 36: This is a hack. Use a valid approach to get the current blog path based on context

BotBlake:
I lack the skill needed to do this change.

PR #1842, opened 2026-05-06
align book examples with wording

### LINKED ISSUE #1830: Books documentation conflict
The example within the Books documentation does not properly reflect the requirements outlined by the text, specifically that “Each book should be in their own folder.” The example shows Book1.flac outside of its own folder, sitting alongside the folder for Book2. 

If the one-book-per-folder is a rule rather than a suggestion, the documentation should show Book1.flac inside of a Book1 folder. 

Books
├── Audiobooks
│   ├── Author
│   │   ├── Book1.flac
│   │   └── Book2
│   │       └── Book2.mp3

https://jellyfin.org/docs/general/server/media/books/

BotBlake:
Technically this is supported, but it is not recommend.
It also differs from the wording just above.

Fixes https://github.com/jellyfin/jellyfin.org/issues/1830

### DIFF
--- a/docs/general/server/media/books.md
+++ b/docs/general/server/media/books.md
@@ -13,7 +13,8 @@ Books should be organized by type (Audiobooks, Books, Comics), then optionally b
 Books
 ├── Audiobooks
 │   ├── Author
-│   │   ├── Book1.flac
+│   │   ├── Book1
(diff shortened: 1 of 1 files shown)


anultravioletaurora (approved):
lgtm

PR #1855, opened 2026-05-12
enable lock file Maintenance

### LINKED ISSUE #1851: Update browser data regularly
When running a docusaurus build (through GitHub Actions), the following warning appears:

Image

Both the `baseline-browser-mapping` and the `browserlist` seem to be outdated.

To ensure a smooth experience across devices and browsers this data should be updated regularly.

BotBlake:
It seems to be recommended to enable `LockFileMaintenance` to keep the browserlist-db up-to-date.
https://github.com/renovatebot/renovate/issues/8615

Alternatively we could add it directly to the project dependencies, but I think this will do the trick for us.

Fixes https://github.com/jellyfin/jellyfin.org/issues/1851

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,18 +1,13 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": [
-    "github>jellyfin/.github//renovate-presets/nodejs",
-    ":semanticCommitsDisabled"
-  ],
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
We should add a direct dependency on caniuse-lite instead to fix the linked issue, like we already did to the jellyfin-web repository. Otherwise Renovate policy is configured in the .github repository with only repository-specific configuration made on the repository itself.

Closing as invalid.

PR #1872, opened 2026-05-24
Fix typo

d4rklynk:
Fix typo in http challenge


BotBlake (approved):
Its true
https://doc.traefik.io/traefik-hub/api-gateway/reference/install/tls/cert-resolvers/ref-certs-resolvers-acme

Issue #1878, opened 2026-05-27
Website redesign proposal - Downloads page


erikbucik:
I find our current downloads page confusing for ~~first time~~ visitors, and hard to download the appropriate file(s).

### I propose the whole page layout to change:

<sub>1. The main "Download" button should lead to "jellyfin.org/downloads".
We'll discuss "/server" later.</sub>
2. This page then shows you the official client for your platform (the browser knows the device & os you're on, doesn't it?) and its download options => easy to download/install.
3. Below that we can display some recommended 3rd-party clients.
4. And after that a link to all listed clients (auto-filtered for your platform).

<img width="495" height="700" alt="Image" src="https://github.com/user-attachments/assets/966f17f4-01dc-40ac-8bb8-4230b7f640b6" />

This solves the confusion for new users, because it's clear and easy to get **what they're most likely looking for** - an app to consume media with.

<sub>5. NEVER should any of the download buttons link to the repo!</sub>

timreichen:
I like that proposal in general.

> 2. This page then shows you the official client for your platform (the browser knows the device & os you're on, doesn't it?) and its download options => easy to download/install.

Yes, we can show the browser platform client  by analyzing the user agent as we do for server downloads already.
What if there are multiple official clients for a platform like ios (`jellyfin mobile`, `swiftfin`)?

> 4. And after that a link to all listed clients (auto-filtered for your platform).

I guess that would be the current filter page, so that could be worked at independently of the download landing page?

> <sub>5. NEVER should any of the download buttons link to the repo!</sub>

Agreed, we need a way to get the download urls for specific platforms. Is there an api to access these?

BotBlake:
Let me also add my feedback to this:
>  1. The main "Download" button should lead to "jellyfin.org/downloads".

Heavily disagreeing here! People who open "jellyfin.org" potentially do potentially not even know what Jellyfin is.
We are advertising mostly the SERVER on our landing page, hence the big Download button should allow you to download the SERVER as well.

or which button are you talking about right now?

<img width="515" height="277" alt="Image" src="https://github.com/user-attachments/assets/a7b79d66-2fde-42b4-944f-20c4828279f4" />

> 2. This page then shows you the official client for your platform (the browser knows the device & os you're on, doesn't it?) and its download options => easy to download/install.

Yesnt. Since this is a website we will realistically always exclusively recommend "Jellyfin-Web" here.
[...]

PR #1888, opened 2026-06-01
Move client filter into filter card

timreichen:
**Changes**
This PR moves the  `all`/`recommended` filter into the filter box.
This is in order to improve the readability of the page and in preparation for the redesign discussed https://github.com/jellyfin/jellyfin.org/issues/1878
This also goes hand in hand with https://github.com/jellyfin/jellyfin.org/pull/1886 where filters will always be visible.

**Copyediting**

**Issues**

### DIFF
--- a/src/pages/downloads/clients/index.tsx
+++ b/src/pages/downloads/clients/index.tsx
@@ -106,23 +106,6 @@ export default function ClientsPage({ recommended = true }: { recommended?: bool
               </div>
             </div>
 
-            <div className={clsx('col', 'margin-bottom--md', styles['header-pills-middle'])}>
-              <div className='pills'>
-                <Link
-                  to={`/downloads/clients${location.search}`}
-                  className={clsx('pills__item', { 'pills__item--active': filter.recommended })}
-                >
-                  Recommended
-                </Link>
-                <Link
-                  to={`/downloads/clients/all${location.search}`}
-                  className={clsx('pills__item', { 'pills__item--active': !filter.recommended })}
-                >
-                  All
-                </Link>
-              </div>
-            </div>
-
             <div className={clsx('col', 'margin-bottom--md', styles['header-pills-end'])}>
               <button
                 className={clsx(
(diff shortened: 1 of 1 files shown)


BotBlake (changes requested):
I'm not convinced by this change for a couple of reasons:

1. We currently only show "Recommended" downloads by default. With this change, it's no longer obvious that a filter is active unless the user opens the filter menu.
2. Switching between "All" and "Recommended" closes the filter menu automatically. None of the other filter options behave this way, which makes the interaction feel inconsistent.

Overall, I don't see a usability benefit here and would prefer to keep the current behavior.

More importantly, we already have an open issue discussing potential changes to this area. My understanding was that we were still evaluating and discussing possible approaches. Opening multiple implementation PRs before we've agreed on the desired UX creates additional review and maintenance overhead, especially if we end up deciding on a different direction.

I'd prefer that we first reach agreement in the issue and then move forward with implementation based on that consensus.

Kind regards

timreichen:
> I'm not convinced by this change for a couple of reasons:
> 
> 1. We currently only show "Recommended" downloads by default. With this change, it's no longer obvious that a filter is active unless the user opens the filter menu.
> 2. Switching between "All" and "Recommended" closes the filter menu automatically. None of the other filter options behave this way, which makes the interaction feel inconsistent.
> 

I agree, this Pr requires https://github.com/jellyfin/jellyfin.org/pull/1886 to always show filter options to be useful. Converted into a draft.

> Overall, I don't see a usability benefit here and would prefer to keep the current behavior.

The benefit is that all filter options are in one place and not scattered over the page. This reduces cognitive load.

> More importantly, we already have an open issue discussing potential changes to this area. My understanding was that we were still evaluating and discussing possible approaches. Opening multiple implementation PRs before we've agreed on the desired UX creates additional review and maintenance overhead, especially if we end up deciding on a different direction.

Maybe I misunderstood https://github.com/jellyfin/jellyfin.org/issues/1878#issuecomment-4587124375, I interpreted it that the filter page would be independent from the redesign "download landing page".

> I'd prefer that we first reach agreement in the issue and then move forward with implementation based on that consensus.
> 
> Kind regards

Sorry for the confusion.

PR #1890, opened 2026-06-02
Auto detect os for client downloads

timreichen:
**Changes**
This PR adds os detection to the client download page and presets the platform filter.
In order to do so the platform filter is modified to a single value instead of an array which also makes the filter more intuitive to use imo.
This goes hand in hand with https://github.com/jellyfin/jellyfin.org/pull/1886 and is in preparation for the redesign discussed [here](https://github.com/jellyfin/jellyfin.org/issues/1878)

**Copyediting**

**Issues**

### DIFF
--- a/src/pages/downloads/clients/index.tsx
+++ b/src/pages/downloads/clients/index.tsx
@@ -34,6 +34,43 @@ function toggleValue<Type>(array: Type[], value: Type): Type[] {
   }
 }
 
+function getPlatform(userAgent: string): Platform | null {
+  const parser = new UAParser(userAgent);
+  const os = parser.getOS();
+  switch (os.name) {
+    case 'macOS':
+      // TODO : Re-enable platform-specific clients once Platform.Desktop is no longer used for all desktop platforms.
+      // return Platform.MacOS;
+      return Platform.Desktop;
+    case 'Windows':
+      // TODO : Re-enable platform-specific clients once Platform.Desktop is no longer used for all desktop platforms.
+      // return Platform.MacOS;
+      return Platform.Desktop;
+    case 'Linux':
+    case 'Ubuntu':
+    case 'Debian':
+    case 'Arch':
+    case 'CentOS':
+    case 'Fedora':
+    case 'Gentoo':
+    case 'Sailfish':
+      // TODO : Re-enable platform-specific clients once Platform.Desktop is no longer used for all desktop platforms.
+      // return Platform.Linux;
+      return Platform.Desktop;
+    case 'Android':
+      return Platform.Android;
+    case 'watchOS':
+    case 'iOS':
+      return Platform.IOS;
+    case 'WebOS':
+      return Platform.WebOS;
+    case 'Xbox':
+      return Platform.Xbox;
+    default:
+      return null;
+  }
+}
+
 export default function ClientsPage(options: { filter?: ClientFilter }) {
   const history = useHistory();
   const location = useLocation();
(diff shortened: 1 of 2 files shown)


src/pages/downloads/clients/index.tsx
```diff
@@ -32,16 +34,82 @@ function toggleValue<Type>(array: Type[], value: Type): Type[] {
...
+        break;
+      case 'CentOS':
+        platform = Platform.CentOS;
+        break;
+      case 'Fedora':
+        platform = Platform.Fedora;
+        break;
+      case 'Gentoo':
+        platform = Platform.Gentoo;
+        break;
```
felix920506: why all these different linux cases?

timreichen: According to [UAParser docs](https://docs.uaparser.dev/info/os/name.html) these are all on the os detection list and we have them listed as separate platforms in `Platform`.
I guess we could just set the platform filter to `Linux` but would need to add an entry in each client so it would show up.

BotBlake: `Linux` should always be in the User header.
These are some example User Agend Headers from debian:

grafik

Or Arch:
grafik

timreichen: > `Linux` should always be in the User header. These are some example User Agend Headers from debian:
> 
> grafik
> Or Arch: grafik

Ok, I'll update the PR so all these linux OSs are set to `Platform.Desktop`.

BotBlake (changes requested):
What? 
Image

This implementation is very far from optimal and did not even address any of the concerns mentioned in the past review comments.


Bonus... It does not even work for Linux distros that are not specifically listed here.
Have you even tested your changes?

timreichen:
> What? 
> Image
> 
> This implementation is very far from optimal and did not even address any of the concerns mentioned in the past review comments.
> 
> 
> Bonus... It does not even work for Linux distros that are not specifically listed here.
> Have you even tested your changes?
> 
> 

I tried it on MacOS.

Which changes would you suggest?

BotBlake:
> I tried it on MacOS.

I'd suggestyou try it for all Platforms, otherwise only one third (even less concidering mobile platforms) of your changes are tested. You dont actually need to be on the platform in order to test this.
You can use a Browser Extension like [this one](https://webextension.org/listing/useragent-switcher.html) to modify your user header.

> Which changes would you suggest?

All mayor Linux distros have the word "Linux" or "linux" in the User-Agent header. Is it not possible to just set the platform to "Linux" whenever the string "linux" appears somewhere in the user-Agent header?

timreichen:
> > Which changes would you suggest?
> 
> All mayor Linux distros have the word "Linux" or "linux" in the User-Agent header. Is it not possible to just set the platform to "Linux" whenever the string "linux" appears somewhere in the user-Agent header?

I think we should avoid manual UA parsing because it is probably not as simple as `ua.includes("Linux")`. There might be many edge cases to consider depending on device type/browser etc. and is potentially a nightmare to maintain. So therefore I think we should actually list all relevant OSs in the switch (is the list incomplete or wrong? If so: What is missing or wrong?) and default to `null` otherwise as it does now to not preselect anything.

BotBlake:
Of course the list is incomplete. There are tons of Linux distros. We can't maintain a list of them here. Jellyfin supports everything Debian or RPM based 🗿.

timreichen:
> Of course the list is incomplete. There are tons of Linux distros. We can't maintain a list of them here. Jellyfin supports everything Debian or RPM based 🗿.

I am not at all familiar with all the different linux repos...

I guess we could do

```ts
switch (os.name) {
  case 'macOS':
    // TODO: Re-enable platform-specific clients once Platform.Desktop is no longer used for all desktop platforms.
    // return Platform.MacOS;
    return Platform.Desktop;
  case 'Windows':
    // TODO: Re-enable platform-specific clients once Platform.Desktop is no longer used for all desktop platforms.
    // return Platform.MacOS;
    return Platform.Desktop;
  case 'Android':
    return Platform.Android;
  case 'watchOS':
  case 'iOS':
    return Platform.IOS;
  case 'WebOS':
    return Platform.WebOS;
  case 'Xbox':
    return Platform.Xbox;
  default:
    if (/linux/i.test(userAgent)) {
      // TODO: Re-enable platform-specific clients once Platform.Desktop is no longer used for all desktop platforms.
      // return Platform.Linux;
      return Platform.Desktop;
    }
    return null;
}
```
as you suggested.

I think this would mostly work, but potentially could break on edge cases, though might be easier to maintain in such cases than a complete list of all linux repos.

WDYT?

BotBlake:
I don't think adding `/linux/i` in the default case really fixes the underlying issue.

We're parsing the UA with UAParser, switching on its structured OS result, and then if the OS isn't one we've explicitly handled, falling back to searching the raw UA string. That means the correctness of the fallback depends on UAParser having already recognized every non-desktop platform whose UA happens to contain Linux.

I also don't think enumerating every Linux distro is a good solution. That just moves the maintenance problem into this switch.

Could we instead model what this function actually needs to determine? We aren't really interested in the exact OS here; we're trying to choose the most appropriate download category. UAParser exposes device information as well as OS information, so I'd prefer deriving "desktop" from the parsed result where possible rather than reparsing the raw UA ourselves.

So how do you like the idea of using [GetDevice](https://docs.uaparser.dev/api/main/get-device.html) "type" and setting that to "mobile" / "desktop" / "TV"? It is not the same as Operating System detection, BUT it results in the same: The Client list having some filters pre-applied to show less stuff at once.

If there isn't enough information to classify an unknown UA confidently, returning null and showing the unfiltered downloads may be preferable to confidently selecting the wrong category.

PR #1893, opened 2026-06-03
Rework Custom CSS guide

BotBlake:
The CSS Customization documentation was heavily outdated. Most of the included CSS snippets had not been updated since Jellyfin 10.5 and no longer work with current Jellyfin releases.

Changes in this PR:

1. Remove the CSS snippet examples. Maintaining a large collection of examples across Jellyfin Web releases is difficult, and many of the existing snippets were already broken.
2. Remove the CSS guide section. Most of the linked guides originated from Reddit posts that are 5+ years old and are unlikely to work correctly with current Jellyfin versions.
3. ~~Remove community themes that have not been updated in years and can no longer be assumed to be compatible with current releases.~~
4. ~~Add [ElegantFin](https://github.com/lscambo13/ElegantFin) as a community theme to replace some of the removed entries, as it is actively maintained and widely used.~~
5. Rephrase parts of the section to reflect the above changes and clarify how custom CSS works.
6. Add a note about importing externally hosted CSS resources and the potential impact on client accessibility and offline functionality.
7. Remove unused images AND replace old "css field" image with a new one.
8. Add image for Custom-CSS User-settings. (+ move to seperate folder) 

Instead of 3. and 4. we agreed on entirely delisting the externally maintained section.
This significantly simplifies the future maintanance work for us, whilst it means we will have to give up on the promotion of some established community projects.


docs/general/clients/css-customization.md
```diff
@@ -5,16 +5,28 @@ title: CSS Customization
...
+Custom CSS is only applied to clients that load Jellyfin Web from your server. As a result, it does not affect all Jellyfin clients. Some web-based clients bundle their own version of Jellyfin Web and therefore do not load custom CSS from the server.
 
-If you have little or no experience with CSS, various resources and tutorials can be found online. Using the tweaks and examples below makes it quite easy to get started with making your own changes to your Jellyfin instance.
+To add custom CSS, navigate to `Dashboard > Branding` and enter your stylesheet in the "Custom CSS" field. Custom CSS is loaded after Jellyfin's default styles, allowing you to override existing rules and customize elements such as colors, layouts and sizing.
```
nielsvanvelzen: fwiw it is also possible to set user specific CSS in the users display settings. Might be worth mentioning here.

docs/general/clients/css-customization.md
```diff
@@ -61,744 +73,15 @@ CSS can be "chained" together to modify different sections together at the same
...
-- [Customizable Plug n' Play CSS for Jellyfin](https://www.reddit.com/r/jellyfin/comments/g9gmjj/customizable_plug_n_play_css_for_jellyfin)
-- [Easy Jellyfin custom CSS](https://www.reddit.com/r/jellyfin/comments/crxqk5/easy_jellyfin_custom_css)
-- [Custom CSS - updated for 10.5.0](https://www.reddit.com/r/jellyfin/comments/fgmu6k/custom_css_updated_for_1050)
-- [Sharing even more custom CSS (and some fixes to previous stuff)](https://www.reddit.com/r/jellyfin/comments/bvnt65/sharing_even_more_custom_css_and_some_fixes_to)
-- [Posting my Jellyfin Custom CSS](https://www.reddit.com/r/jellyfin/comments/p13yqg/posting_my_jellyfin_custom_css/)
+The community maintains a wide range of themes and guides for CSS customization. Some useful resources can be found below.
+Before applying a theme or customization, verify that it is actively maintained and compatible with your version of Jellyfin. Themes that have not been updated for recent releases may not work correctly or could cause unexpected UI issues. These resources are maintained by the community and are not supported by the Jellyfin Project.
 
 ### Community Themes
```
nielsvanvelzen: Lets remove this section so we don't need to maintain it. Otherwise it will be another one of those lists like clients/plugins where we have to fight which ones to add.

BotBlake: The css plugins listed NOW have been maintained for years and even support unstable/ nightly builds for master.

I think those ones are fine to keep in.
And for ruling I think the Client listing rules should just apply to ALL of the external applications we list on our website. So also for plugins + css customizations.

thornbill: +1 for removing. We literally just had an issue come in yesterday where elegantfin broke the Android app for someone.

BotBlake: My view is that this can be handled with a prominent warning, for example:

> Warning: Custom CSS can cause unintended behavior in Jellyfin Web. If you experience UI issues, please remove any custom CSS before reporting the problem.

If we're going to have a Custom CSS guide at all, I think it should contain at least some practical examples and guidance. Otherwise, the page effectively becomes:
> "Custom CSS exists, but we don't recommend using it."

At that point, I'd rather remove the guide entirely than keep a nearly empty page with no actionable content.

That said, if the consensus is that the community links should also be removed, I'm happy to make that change ofc.

BotBlake:
@nielsvanvelzen @thornbill 
I have entirely removed the `external ressources` as agreed on.

This should now be ready to go.

joshuaboniface (approved):

nielsvanvelzen (approved):

PR #1935, opened 2026-07-13
Update recommended TLD in internal networking

schmensch:
Switch from .local to .internal, as .local is bad practice for internal DNS (conflicts with mDNS). .internal is a reserved TLD more appropriate for such use cases.

See e. g. https://unix.stackexchange.com/questions/92441/whats-the-difference-between-local-home-and-lan for more information.

Alternatively, .home.arpa is another reserved TLD for internal use.

**Changes**

**Copyediting**

**Issues**

### DIFF
--- a/docs/general/post-install/networking/index.md
+++ b/docs/general/post-install/networking/index.md
@@ -43,7 +43,7 @@ This section focusses on how to make Jellyfin Available within Networks.
 Here you will find descriptions on how to make Jellyfin accessible both only locally and through the Internet.
 
 In general, Jellyfin will be available locally on the specified port over the host-ip - e.g. `http://10.0.0.2:8096`.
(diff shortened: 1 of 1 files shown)


BotBlake (approved):
Hi there!
Thanks for contributing to the Docs.

If jellyfin where to advertise itself via mDNS, having a `.local` tld would make sense, as it's specifically reserved for that.

However to my knowledge Jellyfin does NOT do that. Hence this value is a pure placeholder. (local DNS Server required to make this work)

It does make sense to have the placeholder prevent accidental Internet access. Therefore I think this change does make sense.

Waiting a bit before merging so that the Team can elaborate on the mDNS usage.

Kind regards

nielsvanvelzen:
We do have wishes to use mDNS but so far nobody has implemented it yet.

BotBlake:
Sweet, merging this for now then.
We need to update the Documentation anyways again when mDNS ever gets used.

PR #1955, opened 2026-07-27
Document SQLite database tuning and cache_size for large libraries

MaxRink:
**Changes**

Adds a Database section to the admin Configuration page. It documents the `database.xml` `CustomProviderOptions` that have existed since 10.11 (`cacheSize`, `lockingmode`, `journalsizelimit`, `tempstoremode`, `syncmode`, `pooling`, `command-timeout`, and `#PRAGMA:` passthrough), and gives cache_size guidance for large libraries.

Motivation: on large libraries the default 2 MiB SQLite page cache is too small to hold the working set for the broad metadata queries (`/Artists`, `/Studios`, `/Genres`), so it thrashes and re-reads pages from disk on every request. Raising `cacheSize` fixes it without any code change. On a 189k item library, `cacheSize=-262144` (256 MiB) took `/Studios` from tens of seconds to ~2-3s. This documents the existing knobs rather than changing defaults, matching the discussion in jellyfin/jellyfin#17405.

**Code assistance**

Benchmarking, the cache_size sweep, and drafting were done with LLM assistance. Numbers were measured against a real 189k item library; the wording and guidance were reviewed and verified by me.

**Issues**

Refs jellyfin/jellyfin#17405

### DIFF
--- a/docs/general/administration/configuration.md
+++ b/docs/general/administration/configuration.md
@@ -114,7 +114,7 @@ This section lists all the configuration options available and explains their fu
 
 ## Database
 
-Jellyfin stores its library metadata in a SQLite database (`jellyfin.db`) inside the [data directory](#data-directory). Since 10.11 the SQLite provider can be tuned through a `database.xml` file in the [configuration directory](#configuration-directory). This file is optional. If it is missing, the built in defaults are used.
+By default, Jellyfin stores its library metadata in a SQLite database (`<Data Directory>/jellyfin.db`) in the [data directory](#data-directory). Set the `path` custom database option to use a different database file. Since 10.11 the SQLite provider can be tuned through a `database.xml` file in the [configuration directory](#configuration-directory). This file is optional. If it is missing, the built in defaults are used.
 
 Most installs never need to touch this. The main reason to tune it is a large library (roughly 100k+ items) where broad queries such as `/Artists`, `/Studios` and `/Genres` feel slow, which is usually the SQLite page cache being too small to hold the working set (see [jellyfin/jellyfin#17405](https://github.com/jellyfin/jellyfin/issues/17405)).
 
@@ -127,10 +127,10 @@ Most installs never need to touch this. The main reason to tune it is a large li
   <LockingBehavior>NoLock</LockingBehavior>
   <CustomProviderOptions>
     <Options>
-      <!-- 1024 * 1024 * 1024 / 4096 = 1G -->
+      <!-- 256 MiB: negative cache_size values specify KiB -->
       <CustomDatabaseOption>
         <Key>cacheSize</Key>
-        <Value>262144</Value>
+        <Value>-262144</Value>
       </CustomDatabaseOption>
     </Options>
   </CustomProviderOptions>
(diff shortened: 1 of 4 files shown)


MaxRink:
Heads up on placement: there is already some overlap in the docs. `troubleshooting.md` documents editing `database.xml` for the `LockingBehavior` option (Database Locked Errors), and `storage.md` has a ZFS record-size note for the SQLite db. So `database.xml` and db tuning are currently spread across a couple of pages.

I put the new section in `configuration.md` since that felt like the natural home for config-file docs, but I dont have strong feelings about it. Happy to move it, or fold it into one of the existing spots and just cross-link, whatever you'd prefer. Where would you want this to live?

docs/general/administration/configuration.md
```diff
@@ -112,6 +112,53 @@ This section lists all the configuration options available and explains their fu
...
+<?xml version="1.0" encoding="utf-8"?>
+<DatabaseConfigurationOptions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+  <DatabaseType>Jellyfin-SQLite</DatabaseType>
+  <CustomProviderOptions>
+    <Options>
+      <CustomDatabaseOption Key="cacheSize" Value="-262144" />
+    </Options>
+  </CustomProviderOptions>
+</DatabaseConfigurationOptions>
+```
```
jhparker88, line 148: As noted by https://github.com/jellyfin/jellyfin/pull/15986#issuecomment-3764485801, the actual syntax is:

```
<?xml version="1.0" encoding="utf-8"?>
<DatabaseConfigurationOptions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
  <DatabaseType>Jellyfin-SQLite</DatabaseType>
  <LockingBehavior>NoLock</LockingBehavior>
  <CustomProviderOptions>
    <Options>
      <!-- 1024 * 1024 * 1024 / 4096 = 1G -->
      <CustomDatabaseOption>
        <Key>cacheSize</Key>
        <Value>262144</Value>
      </CustomDatabaseOption>
    </Options>
  </CustomProviderOptions>
</DatabaseConfigurationOptions>
```

MaxRink:
You're right, Fixed in b96f7ff.

One difference from your snippet: I kept the value negative (`-262144`). SQLite reads a positive `cache_size` as a page count and a negative one as KiB, so the negative form is page-size independent and gives a predictable 256 MiB regardless of the DB's page size. A positive `262144` would be 262144 pages, which is ~1 GiB at 4K pages, probably more than intended.

docs/general/administration/configuration.md
```diff
@@ -112,6 +112,66 @@ This section lists all the configuration options available and explains their fu
 | `FFmpeg:analyzeduration` | `"200M"`                               | The value to set for the FFmpeg `analyzeduration` format option. See the FFmpg [documentation](https://ffmpeg.org/ffmpeg-formats.html#Format-Options) for more details. |
 | `PublishedServerUrl`     | Server Url based on primary IP address | The Server URL to publish in udp Auto Discovery response.                                                                                                               |
 
+## Database
```
JPVenson, line 115: this needs a very big disclaimer that changes made here are only for advanced users and may result in database corruption if wrongly applied.

Also this configuration does NOT constitute our requirements of a stable API, so changes might not be properly migrated when things change.

docs/general/administration/configuration.md
```diff
@@ -112,6 +112,66 @@ This section lists all the configuration options available and explains their fu
...
+  <Key>#PRAGMA:mmap_size</Key>
+  <Value>268435456</Value>
+</CustomDatabaseOption>
+```
+
+### Cache size and large libraries
+
+`cacheSize` is the setting that matters most for large libraries. When it is unset, Jellyfin does not emit `PRAGMA cache_size`, so SQLite's default suggested cache size of `-2000` (2,048,000 bytes, approximately 2 MiB) applies. On a large database the metadata queries touch more pages than that in a single request, so the cache thrashes and pages are re-read from disk on every call.
+
+Raising `cache_size` so the hot working set stays in memory removes the thrashing. On a 189k item library, setting `cacheSize` to `-262144` (256 MiB) dropped `/Studios` from tens of seconds to roughly 2 to 3 seconds. The gain flattens out once the cache is large enough to hold the working set, so there is little point going much beyond a couple hundred MiB.
```
JPVenson: This needs to be phrased a lot more vague. Performance on those systems is a hit-or-miss thing and can be effected by quite a bit more then just this single parameter.

Also we do not really need to explain the inner workings of that parameter here, this is the jellyfin documentation, not the sqlite one.

JPVenson (changes requested):

docs/general/administration/storage.md
```diff
@@ -57,3 +57,5 @@ For ZFS datasets containing large media files (i.e., not the dataset containing
 Note that changing the record size on an existing ZFS dataset will not change the existing data within it, meaning performance will not be any different for anything but newly-written changes into the dataset. As such, it is recommended to rewrite your data into the dataset to take advantage of the change in record size; otherwise, the configuration change will not yield the expected change in performance.
 
 As ZFS snapshots can use a lot of storage over time without a sensible `destroy` schedule, there may be a temptation to keep your data on a mechanical drive instead of an SSD. Do not use ZFS-formatted mechanical drives to store your Jellyfin Server data (everything except your media files), as this will result in poor performance. An SSD is strongly recommended.
+
+For SQLite tuning options such as the `cacheSize` setting that helps large libraries, see the [Database section](/docs/general/administration/configuration#database) of the Configuration page.
```
JPVenson:
```suggestion
For SQLite tuning options such as the `cacheSize` setting that may help large libraries, see the [Database section](/docs/general/administration/configuration#database) of the Configuration page.
```

JPVenson (changes requested):

JPVenson (approved):

Shadowghost (approved):

IDisposable (approved):

jhparker88:
@MaxRink Any reason this PR got closed?

MaxRink:
No, just me fat fingering when cleaning up my old PRs

BotBlake (approved):

PR #1833, opened 2026-05-03
Update rules for inclusion in client list

joshuaboniface:
**Changes**
Officially promulgate our updated rules for inclusion in the official client list, as discussed by the team over the last few months.

The rules provide much greater clarity on what is permitted to be listed in the official clients list, while also including a blog post detailing the rationale behind each rule.

**Copyediting**

To avoid "nitpicky" reviews, please ensure all of the following have been done for any non-trivial changes.

- [x] I have run this PR [through a spellchecker](https://jellyfin.org/docs/general/contributing/documentation#please-self-review) (e.g. `aspell`).
- [x] I have re-read my PR at least twice and fixed any obvious mistakes I see.
- [x] I have received [out-of-band peer copyediting](https://jellyfin.org/docs/general/contributing/documentation#peer-copyediting) from someone in [#jellyfin-documentation](https://matrix.to/#/#jellyfin-documentation:matrix.org).

While you're waiting for someone to look at your pull request, How about looking at another one? You do not have to do this, but it will help ensure your PR is reviewed quickly in turn.

- [x] I have provided [a *substantive* review of another documentation PR](https://jellyfin.org/docs/general/contributing/documentation#peer-reviews).

**Issues**

N/A

PR #1960, opened 2026-07-28
Add Neptune to client list

need4swede:
```markdown
**Changes**

Adds Neptune as a third-party proprietary tvOS client, with links to the App Store and Neptune website.

**Copyediting**

This is a small data-only change.

**Issues**

None.
```

### DIFF
--- a/src/data/clients.ts
+++ b/src/data/clients.ts
@@ -680,6 +680,29 @@ const thirdPartyClients: Array<Client> = [
     recommended: true
   },
   {
+    id: 'neptune',
+    name: 'Neptune',
+    description: 'A native Jellyfin client for Apple TV.',
+    clientType: ClientType.ThirdParty,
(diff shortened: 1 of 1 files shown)


BotBlake (changes requested):
We are currently revising our [client listing guidelines](https://github.com/jellyfin/jellyfin.org/pull/1833). Until that process is complete, changes to the client list submitted by third parties are temporarily on hold.

Because the proposed guidelines are still subject to change, I cannot formally accept or reject this PR based on them. That said, Neptune in its current form does not appear to align with the intended direction of the client list.

Going forward, we do not intend to promote closed-source clients on our website, as this does not align with how Jellyfin and its ecosystem operate. This should not be interpreted as discouragement of closed-source clients in general.

need4swede:
Understood, thank you

PR #1962, opened 2026-07-29
Update k8s hardware acceleration docs to reflect current helm chart

ozzywalsh:
**Changes**
* Update incorrect helm values in k8s hardware acceleration docs 
The chart uses `volumes` & `volumeMounts`, not `extraVolumes` & `extraVolumeMounts` https://github.com/jellyfin/jellyfin-helm/blob/b8e2455d5dcaf1285e999b2058fb8c21218e2c4e/charts/jellyfin/values.yaml#L207


Shadowghost (approved):

BotBlake:
Well if Ghost says it's correct, I will certainly not question it.

PR #1970, opened 2026-08-05
Make hardware selection a category

felix920506:
**Changes**

Splits the hardware selection into a simple version and more details in another page to avoid overwhelming new users with a wall of text

**Copyediting**

**Issues**

### DIFF
--- a/docs/general/administration/hardware-selection/index.md
+++ b/docs/general/administration/hardware-selection/index.md
@@ -37,6 +37,9 @@ The following hardware is explicitly recommended against for a Jellyfin server, 
 
 - Intel "Atom" CPUs (Intel J/M/N/Y series low power CPUs) up to 11th gen
 - Prebuilt NAS Appliances
+  <details>
+    Prebuilt NAS appliances often come with weird software environments and very low end hardware. This will lead to both difficulty in setting Jellyfin up and/or poor performance.
+  </details>
 - Most Single Board Computers (SBC): If you really want to run Jellyfin on an SBC, you may wish to consider models based on the following platforms: Rockchip RK3588 / RK3588S, Intel Core, Intel 12th gen N series.
 - AMD Graphics: AMD Graphics have poor driver support on Windows and poor encoder quality.
   <details>
--- a/docs/general/administration/hardware-selection/index.md
+++ b/docs/general/administration/hardware-selection/index.md
@@ -13,7 +13,7 @@ This is a list of system requirements for the current release of Jellyfin. Syste
 
 ## Recommended System Requirements
 
-Below is a list of recommended specs to run Jellyfin. Whilst Jellyfin can work on relatively low-end hardware, the following specs are recommended for a good experience.
+Below is a list of recommended specs to run Jellyfin. A good Jellyfin experience can be had with minimal tweaking on platforms that meet these specs. Whilst Jellyfin can work on relatively low-end hardware, the following specs are recommended for a good experience.
 
 - CPU: Intel Core i5-2300, AMD Ryzen 3 1200, Apple M1 or above.
 - RAM: 8GB or above, 16GB or above for Windows 11.
--- a/docs/general/administration/hardware-selection/index.md
+++ b/docs/general/administration/hardware-selection/index.md
@@ -21,7 +21,7 @@ Below is a list of recommended specs to run Jellyfin. Whilst Jellyfin can work o
 - Internet: Ethernet, Broadband Internet
 - Storage 100GB Free Space, SSD.
 
-### Low Power or Budget Servers
+## Low Power or Budget Servers
 
 For users with expensive electricity, or running battery-powered servers:
 
--- a/docs/general/administration/hardware-selection/index.md
+++ b/docs/general/administration/hardware-selection/index.md
@@ -1,8 +1,3 @@
----
-uid: admin-hardware-selection
-title: Hardware Selection
----
-
 # Selecting Appropriate Hardware
 
 The following is intended to help you choose appropriate hardware for a Jellyfin server and take full advantage of its features (e.g. hardware acceleration).
@@ -85,7 +80,7 @@ These hardware platforms might lead to a poor Jellyfin experience. Please be car
 - Most Single Board Computers (SBC): Most SBCs (including the Raspberry Pi, **especially the Raspberry Pi 5**) are too slow to provide an acceptable Jellyfin experience as they often lack proper support for hardware acceleration. If you really want to run Jellyfin on an SBC, you may wish to consider models based on the following platforms: Rockchip RK3588 / RK3588S, Intel Core, Intel 12th gen N series
 - AMD Graphics: AMD Graphics have poor encoder quality and poor driver support. **This applies even on Linux**.
 - Low-end GPUs: Certain low-end GPUs (e.g. GT1030, RX6400) are not capable of hardware encoding. These models cannot be used for hardware acceleration for a Jellyfin Server.
-- Very old X86 CPUs: Starting with Jellyfin 10.11, X86 CPUs that support the SSE4.1 instruction set is a requirement.  For Intel, Penryn (Q4 2007) and newer support SSE4.1.  For AMD, Bulldozer (Q4 2011) and newer support SSE4.1.
+- Very old X86 CPUs: Starting with Jellyfin 10.11, X86 CPUs that support the SSE4.1 instruction set is a requirement. For Intel, Penryn (Q4 2007) and newer support SSE4.1. For AMD, Bulldozer (Q4 2011) and newer support SSE4.1.
 
 ## Detailed Guide
 
--- a/docs/general/administration/hardware-selection/index.md
+++ b/docs/general/administration/hardware-selection/index.md
@@ -1,69 +1,27 @@
-# Selecting Appropriate Hardware
+# Hardware Selection
 
 The following is intended to help you choose appropriate hardware for a Jellyfin server and take full advantage of its features (e.g. hardware acceleration).
 
-## Simple Guide
+## Basic System Requirements
+
+This is a list of system requirements for the current release of Jellyfin. Systems that do not meet these requirements will not function at all for Jellyfin.
+
+- CPU (x86): SSE4.1 Support (Intel 45nm Core 2 Duo, AMD FX or above).
+- CPU (ARM): 64-bit ARM CPU.
+- OS: Windows 10 or above, macOS 11 or above, Linux. 64-bit OS is required.
+- Storage: 2GB Free Space.
+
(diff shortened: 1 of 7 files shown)


docs/general/administration/hardware-selection/index.md
```diff
@@ -0,0 +1,40 @@
+# Hardware Selection
+
+The following is intended to help you choose appropriate hardware for a Jellyfin server and take full advantage of its features (e.g. hardware acceleration).
+
+## Basic System Requirements
+
+This is a list of system requirements for the current release of Jellyfin. Systems that do not meet these requirements will not function at all for Jellyfin.
```
BotBlake, line 7: > for the current release of Jellyfin

> Systems that do not meet these requirements will not function at all for Jellyfin.

Those sentences seem somewhat contradictory to another. Will Jellyfin not work at all, or just _the current Release_ ?

docs/general/administration/hardware-selection/index.md
```diff
@@ -0,0 +1,40 @@
...
+
+- CPU (x86): SSE4.1 Support (Intel 45nm Core 2 Duo, AMD FX or above).
+- CPU (ARM): 64-bit ARM CPU.
+- OS: Windows 10 or above, macOS 11 or above, Linux. 64-bit OS is required.
```
BotBlake, line 11: Will Jellyfin actuall not work on Windows 7?
I have not tested it but without looking into it, I see no reason why it would not.

docs/general/administration/hardware-selection/index.md
```diff
@@ -0,0 +1,40 @@
...
+- Storage: 2GB Free Space.
+
+## Recommended System Requirements
+
+Below is a list of recommended specs to run Jellyfin. Whilst Jellyfin can work on relatively low-end hardware, the following specs are recommended for a good experience.
```
BotBlake, line 16: Imo this categorie is _TOO SHORT_
When you do not intend to use transcoding, you will have a good experience even without a GPU at all.

docs/general/administration/hardware-selection/index.md
```diff
@@ -0,0 +1,40 @@
...
+
+- CPU: Intel Core i5-2300, AMD Ryzen 3 1200, Apple M1 or above.
+- RAM: 8GB or above, 16GB or above for Windows 11.
+- Graphics: Intel UHD7xx, Iris Xe, Arc or above; Nvidia GTX1660 or above; Apple M1 or above.
+- Internet: Ethernet, Broadband Internet
+- Storage 100GB Free Space, SSD.
+
+### Low Power or Budget Servers
```
BotBlake, line 24: Imo this is something for the "advanced" category, if at all.
We are recommending specific PRODUCTS here. In my oppinion this is not something we should do.

Also the category headline of "Low Power OR Budget Server" overpromises a lot, when there are only 3 products here at all. Not even explaining which ones are the "Low Power" and which the "Low Budget" ones.

docs/general/administration/hardware-selection/index.md
```diff
@@ -0,0 +1,40 @@
...
+- Rockchip RK3588 / RK3588S SBC (**Advanced Users Only**)
+
+## What not to buy
+
+The following hardware is explicitly recommended against for a Jellyfin server, even though they might technically meet the minimum requirements.
+
+- Intel "Atom" CPUs (Intel J/M/N/Y series low power CPUs) up to 11th gen
+- Prebuilt NAS Appliances
+- Most Single Board Computers (SBC): If you really want to run Jellyfin on an SBC, you may wish to consider models based on the following platforms: Rockchip RK3588 / RK3588S, Intel Core, Intel 12th gen N series.
+- AMD Graphics: AMD Graphics have poor encoder quality and poor driver support. **This applies even on Linux**.
```
BotBlake, line 39: > poor driver support

Is that true? From what I know AMD drivers for linux are pretty decent.
Their encoder quality may not be as good, but claiming the drivers are bad is something else.

docs/general/administration/hardware-selection/index.md
```diff
@@ -0,0 +1,40 @@
...
+- Intel 12th gen or above N series platforms
+- Apple M Series Mac mini
+- Rockchip RK3588 / RK3588S SBC (**Advanced Users Only**)
+
+## What not to buy
+
+The following hardware is explicitly recommended against for a Jellyfin server, even though they might technically meet the minimum requirements.
+
+- Intel "Atom" CPUs (Intel J/M/N/Y series low power CPUs) up to 11th gen
+- Prebuilt NAS Appliances
```
BotBlake, line 37: Imo a short reason for this is needed. This is TOO BROAD considering the sheer amount of prebuild NAS's

docs/general/administration/hardware-selection/index.md
```diff
@@ -0,0 +1,40 @@
...
+- Most Single Board Computers (SBC): If you really want to run Jellyfin on an SBC, you may wish to consider models based on the following platforms: Rockchip RK3588 / RK3588S, Intel Core, Intel 12th gen N series.
```
BotBlake, line 38: imo its worth to mention why here.
Otherwise people will start to assume: "Yea probably because of performance"
And then decide to buy a "RaspberryPi 5 16GB" because it "has a lot more performance then a PI 3 so it will be fine".

BotBlake (changes requested):
As discussed on Matrix we will focus on the simplified guide first and THEN figure out how to re-arrange the `advanced` snippets to fit into it.

I personally dislike a seperate page for those, as it will mostly duplicate  things from the basic page.

docs/general/administration/hardware-selection/index.md
felix920506: Just the current release

docs/general/administration/hardware-selection/index.md
felix920506: If you know you do not intend to use transcoding you are an advance user.

BotBlake: A lot of people I know started with Jellyfin on a Raspberry Pi. Without transcoding. Would you consider that "advanced"?

docs/general/administration/hardware-selection/index.md
felix920506: Low power and low budget used to have some divergence, but ever since Intel effectively deprecated HD6xx supported in their latest media drivers, they are basically the same. All 3 of the platforms listed here are very affordable, consume very little power and provide great performance for Jellyfin usage.

docs/general/administration/hardware-selection/index.md
felix920506: iirc according to nyan the AMD media drivers on Linux are a different story.

docs/general/administration/hardware-selection/index.md
felix920506: I consider that advanced since that needs some insignificant amount of extra tweaking to get a good experience

docs/general/administration/hardware-selection/index.md
felix920506: dotnet 9 (jellyfin 10.11) officially only supports Windows 10 and above.

docs/general/administration/hardware-selection/index.md
felix920506: we can't possibly keep a list of old things and when they broke, that's just impractical.

BotBlake: I never said we should do that, just that:
> Those sentences seem somewhat contradictory to another.

"not for the current release of X" and "not at all for X" are very different things.

docs/general/administration/hardware-selection/index.md
BotBlake: If you move that info to a collapsible, then I am fine with it.
So it doesn't take up more space but the (imo valuable) information is not lost.

docs/general/administration/hardware-selection/index.md
BotBlake: This category is simply too small.
There are no explanations, just 3 Products are recommended.
Documentation should explain, not blindly tell People to do/buy something.

docs/general/administration/hardware-selection/index.md
BotBlake: @nyanmisaka Can you elaborate please?

nyanmisaka: Driver quality was relatively poor in the past (prior to 2024). libva/VA-API is essentially an Intel-led API. AMD's role was primarily to align its implementation with Intel's behavior while adapting it to its own hardware. libva itself has inherent limitations, which could lead to incorrect behavior on AMD hardware.

The situation improved largely due to the arrival of new developers who gradually cleaned up and refactored the driver while actively backporting fixes. jellyfin-ffmpeg also incorporates workarounds for limitations associated with libva and AMD hardware.

As of 2026, I do not believe there are any significant issues with driver quality. Of course, refactoring between major Mesa releases can still cause regressions, much like the reason we do not immediately switch to the latest major ffmpeg release. Users running the Jellyfin Docker container or Deb package remain unaffected, as the jellyfin-ffmpeg Deb package bundles a verified driver version.

Regarding encoding quality, refer to https://jellyfin.org/docs/general/administration/hardware-selection#graphics-cards-gpus

My recommendation is that AMD cards are suitable for those who primarily use a Linux desktop and want a machine that offers good gaming compatibility while also serving as a Jellyfin server during idle time. Cards predating the RDNA3/RX7000 series are no longer recommended for a new purchase due to the lack of an AV1 encoder and mediocre encoding speeds.

BotBlake: Thanks for the input!
So what I'm getting is that "Amd has poor drivers everywhere" defenetly is too short and needs further details.

@felix920506 Can you incorporate nyan's recommendation into the guide please?

BotBlake (changes requested):
I’m still against introducing a separate `advanced` guide, for the same reasons I mentioned before my first review.

One issue is that this creates a hierarchy with only a single subcategory:

> 1. Category
>    1.1 Subcategory
> 2. Category

This is generally considered poor structure in hierarchical outlines: if a level is subdivided, there should be at least two items at the subordinate level. Otherwise, there is no actual subdivision taking place. See [Wikipedia](https://en.wikipedia.org/wiki/Outline_(list))

In our case, introducing a category containing only `advanced` adds another level to the navigation without actually grouping multiple related entries. I would keep that additional information at the existing level until there is at least one other guide that belongs under the same category.

This is especially relevant given that there is still significant overlap between the `advanced` guide and the basics guide. I think the advanced content should be integrated into the basics guide instead (for example, using collapsible sections) rather than repeating the same information in two places. Otherwise, we increase both the amount of documentation users have to navigate and the risk of the two guides becoming inconsistent over time.

PR #1977, opened 2026-08-09
Add Navarre (third-party Android TV / Android / Windows client)

NavarreDR:
Adds Navarre, a third-party client, to the clients list.

- Third-party, proprietary (clientType: ThirdParty, licenseType: Proprietary), consistent with the existing precedent for proprietary third-party clients such as Infuse.
- Living-room-first Android TV app with a setup wizard that discovers the server, plus Android phone and Windows desktop apps.
- Free for in-home use; remote access is a one-time paid unlock. No subscription.
- Store links: Google Play (TV and phone); the Windows installer is on the website.

Happy to adjust the entry, description, or placement to whatever the maintainers prefer.

### DIFF
--- a/src/data/clients.ts
+++ b/src/data/clients.ts
@@ -1024,6 +1024,34 @@ const thirdPartyClients: Array<Client> = [
         url: 'https://jellify.app'
       }
     ]
+  },
+  {
+    id: 'navarre',
+    name: 'Navarre',
+    description: 'A third-party client for Android TV, Android, and Windows, built around a turnkey living-room experience.',
+    clientType: ClientType.ThirdParty,
+    deviceTypes: [DeviceType.TV, DeviceType.Mobile, DeviceType.Desktop],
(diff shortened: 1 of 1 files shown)


thornbill:
- This PR violates our llm policy
- We are revamping the inclusion criteria
- And (the icing on the cake) you expect us to allow you to advertise your client on _our_ website when your marketing is to quote your website "Jellyfin is a pain"...

NavarreDR:
Understood, and accepting the close - not here to argue it.

You're right on the LLM policy: I leaned on AI tooling to prepare this PR and should have read the contribution docs first. That one's on me.

Point taken on the site line, too. Navarre only exists because the Jellyfin server is excellent - the problem I'm solving is the living-room setup experience, not the project - and I'm rewording the site so it says exactly that.

If the revamped criteria end up with room for proprietary third-party clients, I'll resubmit then. Either way, thanks for Jellyfin - my own household runs on it.

PR #1988, opened 2026-08-21
initial stub guide

BotBlake:
Documents Jellyfin's media placeholder behavior using .disc files.

Added a dedicated Media Placeholders page covering naming and optional placeholder types.
Client behavior cannot really be documented, as it strongly depends on the client.

(Note that I have no Idea how Jellyfin-Web shows stubs)

### DIFF
--- a/docs/general/server/media/placeholders.md
+++ b/docs/general/server/media/placeholders.md
@@ -71,3 +71,8 @@ Shows
 ## Client Behavior
 
 Jellyfin adds placeholders to the library like other media items, allowing them to have metadata and artwork. The item is marked as a placeholder instead of playable media.
+
+## Metadata
+
+`.disc`-Files will be treated as a real media file. You can attach additional files like images or `.nfo` files just as usual.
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):
Seems fine to me, maybe we should add a comment if e.g. NFOs and image files are supported?

BotBlake:
Can do that.
As I understand it, stub flies are treated exactly like actual media files.
Just without having a video file :laughing: 

So NFO and everything should be supported exactly the same.

thornbill:
> I have no Idea how Jellyfin-Web shows stubs

They look exactly the same as any other media right up until you hit play and it tells you it can't and you need to insert the disc. 😂

docs/general/server/media/placeholders.md
```diff
@@ -0,0 +1,78 @@
+---
+uid: server-media-placeholders
+title: Media Placeholders
```
thornbill: Calling it "Media Placeholders" when it's already under a "Media" section might be a little redundant 🤷‍♂️

BotBlake: Good point :+1: 
Fixed!

thornbill (approved):

PR #2025, opened 2026-09-17
add rules for documenting server prs

BotBlake:
Since opening a PR on jellyfin-server does not automatically mean it will find its way (at least in its current state) into a release.

To prevent stale PRs for never-merged features, or just duplicate work for PRs where the implementation is still changing, this PR introduces new rules about when to open PRs on the documentation repository.

This is mostly a proposal from myself, so opinions are welcome.


nielsvanvelzen (changes requested):
what is a PR

docs/general/contributing/documentation.md
```diff
@@ -9,6 +9,12 @@ title: Contributing to Documentation
 
 Documentation is fast-moving and ever-changing. Please think carefully about what you're changing and why; what will it contribute going forward? Extensive rewrites and new pages should be considered very carefully and coordinated in our [documentation Matrix room](/contact).
 
+## Documenting Server PRs
+
+If you have opened a Pull Request (PR) on the [jellyfin-server](https://github.com/jellyfin/jellyfin) repository, you are very welcome to open a PR to document the resulting changes yourself.
+
+To avoid duplicate work and reduce the number of stale documentation PRs, please wait until your PR on `jellyfin/jellyfin` has been merged before opening the corresponding Pull Request in the documentation repository.
```
theguymadmax, line 16:
```suggestion
To avoid duplicate work and reduce the number of stale documentation PRs, please wait until your PR on `jellyfin/jellyfin` has received at least one approval or a team member has requested updated documentation before opening the corresponding Pull Request in the documentation repository.
```
I honestly don't think this is a problem. If you want to add this as a recommendation, then fine, but I'm not in favor of waiting for the PR to be "merged" language. I think a better compromise would be to recommend that contributors wait until they get at least one approval on their PR. 

Also, what do you mean by "avoid duplicate work"? How does opening a PR for new documentation contribute to duplicate work?

theguymadmax (changes requested):

docs/general/contributing/documentation.md
BotBlake: We get PRs for documentation on features that haven’t even been implemented yet. Until the corresponding PR is merged into the server repository, a lot of the implementation can still change.

That means the documentation PR will also need to change. This has caused me to review the same PR multiple times because the author had to update it again after changes were made to the implementation.

It also creates extra work for the PR author. With every review and subsequent change to the server PR, they have to make sure the documentation PR still aligns with the implementation.

Why start documenting something before it has even made its way into an unstable release? At that point, the documentation PR will just sit around waiting.

Not to mention that our documentation moves fast. A PR that sits around for half a year while waiting for a feature to make it into a stable server release will almost certainly accumulate a lot of merge conflicts that someone will eventually have to resolve. So far, that job has fallen to me multiple times.

All of this could be avoided if we simply advised people not to open documentation PRs for features that have not yet been implemented in an unstable release.

And from previous experience, an “approval” from someone on the server repository does not mean that the implementation will stay as it is (or that it will ever actually make its way into the upstream code at all).

PR #2026, opened 2026-09-17
run previews for all PRs regardless of branch name

BotBlake:
Currently, previews will only run, if the source branch is not named `master`.
This PR changed it to run on all `pull_request` events, so all PRs will get preview deployments.


.github/workflows/deploy.yml
```diff
@@ -84,7 +84,7 @@ jobs:
 
   preview:
     needs: [metadata]
-    if: ${{ github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.head_branch != 'master' }}
+    if: ${{ github.event.workflow_run.conclusion == 'success' &&  github.event.workflow_run.event == 'pull_request' }}
```
nielsvanvelzen:
```suggestion
    if: ${{ github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'pull_request' }}
```

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):

